PR #466, opened 2026-03-29
SEC-12: Session-token storage hardening (ADR + near-term mitigations)

### LINKED ISSUE #156: SEC-12: Session-token storage hardening plan (localStorage risk reduction and migration path)
## Context
Current frontend session token handling uses browser storage. This is pragmatic, but raises XSS blast-radius risk and should be explicitly hardened or migrated.

Tracked by:
- #151

## Depends on
- #80
- #82
- #83

## Scope
- Produce ADR comparing current storage approach vs HttpOnly-cookie/session approaches for Taskdeck constraints.
- Define near-term mitigations (CSP hardening, token lifetime/rotation posture, strict sanitization checks).
- Define phased migration path if cookie/session model is selected.

## Acceptance Criteria
- Security tradeoffs and final recommendation are documented and approved.
- Immediate mitigations are implemented/tested for current posture.
- If migration selected, phased implementation issues are created and linked.

## Verification
- Security regression checks for auth flows pass.
- Manual validation confirms no auth UX regressions.

## Suggested branch
- `security/session-token-storage-hardening`

Chris0Jeky:
## Summary

Closes #156

- **ADR** (`docs/analysis/session-token-storage-adr.md`): Compares localStorage vs HttpOnly-cookie vs hybrid approaches for JWT storage. Recommends retaining localStorage with near-term mitigations for current local-first deployment model, with phased cookie migration path documented for future hosted scenarios.
- **Token storage abstraction** (`tokenStorage.ts`): Centralized module replacing scattered direct `localStorage` access across 4 files. Provides JWT structure validation, session data sanitization with field length limits, and consistent cleanup semantics.
- **CSP hardening**: Removed `'unsafe-inline'` from `script-src` in the API's default Content-Security-Policy. The API serves JSON, not HTML with inline scripts, so this directive was unnecessary and weakened XSS protection.
- **Session restore hardening**: Malformed tokens are rejected before storage; orphaned tokens (valid token but corrupted session metadata) are cleaned up on restore.
- **Security docs updated**: OWASP baseline cross-references the ADR and reflects the CSP change.

## Test plan

- [x] All 784 frontend vitest tests pass (including 30 new tokenStorage tests)
- [x] Frontend typecheck + build passes
- [x] All 1048 backend tests pass (including SecurityHeadersApiTests verifying CSP presence)
- [x] Demo mode unaffected (does not use token storage)
- [ ] Manual: verify login/logout flow works end-to-end
- [ ] Manual: verify SignalR board realtime reconnection works


Chris0Jeky:
## Self-Review

Reviewed the full PR diff. Findings:

**Looks good:**
- All 4 production files that previously accessed `localStorage` directly for token/session keys are now routed through `tokenStorage` -- no stragglers in production code
- CSP change is minimal and correct: only `'unsafe-inline'` removed from `script-src`, not from `style-src` (which Swagger still needs)
- The `validateSessionData` and `isValidJwtStructure` functions handle edge cases (null, non-string, oversized values, wrong part count)
- Orphaned token cleanup in `restoreSession` prevents inconsistent state when session metadata is corrupt but token is valid
- Test coverage is comprehensive: 30 new tests for tokenStorage, plus the existing sessionStore and realtime tests updated to work with structural validation
- ADR is thorough with clear comparison matrix and phased migration path

**Items noted but acceptable:**
- The existing `sessionStore.spec.ts` tests still use `localStorage.setItem` directly in test setup (to simulate pre-existing persisted state). This is correct -- tests should set up raw storage state to verify the abstraction reads it properly.
- `style-src` still has `'unsafe-inline'` -- documented in ADR as intentional (needed for CSS patterns and Swagger UI).
- Token lifetime reduction (mentioned in ADR Option A as a possible mitigation) is not implemented in this PR. This is a backend JWT config change that should be a separate scoped issue if desired.

**No issues blocking merge.**

Chris0Jeky:
**Follow-up fix**: Strengthened `isValidJwtStructure` to require the payload segment to be base64url-decodable JSON (not just three segments). A value like `aaa.bbb.ccc` is now rejected. Tests updated to cover the non-JSON and corrupted-payload cases.

PR #468, opened 2026-03-29
SEC-18: Abuse state model, operator tooling, and evaluation groundwork

### LINKED ISSUE #238: SEC-18: Managed-key abuse detection and automated containment policy
## Context
Tracked by: #235

Even with quotas, malicious users can still attempt abusive patterns (prompt flooding, scripted account churn, policy-evasion payloads, coordinated spikes).
We need explicit detection + containment before exposing managed-key mode broadly.

## Scope
- Define abuse signals and thresholds for managed-key traffic, including:
  - anomalous request/token velocity
  - repeated blocked/refused content patterns
  - repeated limit-hit evasion attempts
  - suspicious multi-account/IP concentration patterns
- Build a deterministic abuse state model:
  - observe
  - suspicious
  - restricted
  - blocked
- Define automated containment actions per state:
  - stricter throttles
  - temporary user lock
  - provider-call disable for actor/scope
  - mandatory manual review path
- Ensure all automated enforcement is auditable and reversible with operator tooling.

## Depends on
- #236
- #212
- #110

## Acceptance Criteria
- [ ] Abuse detection signals/thresholds are documented and implemented for managed-key paths.
- [ ] Automated containment actions trigger deterministically and emit audit events.
- [ ] False-positive handling flow is documented (override/unblock + evidence trail).
- [ ] Tests cover suspicious-pattern detection, containment transitions, and rollback paths.

## Verification
- `dotnet test backend/tests/Taskdeck.Application.Tests/Taskdeck.Application.Tests.csproj -c Release`
- `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release`

Chris0Jeky:
## Summary
- Adds deterministic 4-state abuse model (Observe → Suspicious → Restricted → Blocked) as domain entities
- Adds AbuseDetectionService with threshold-based escalation and on-demand evaluation
- Adds operator API endpoints for status, audit trail, override/unblock, and manual evaluation

**Scope note**: This PR delivers the state model, operator tooling, and evaluation infrastructure. Wiring automated signal recording into the live managed-key execution path (so Restricted/Blocked states actually gate traffic) is a planned follow-up slice.

Closes #238

## Test plan
- [x] `dotnet test backend/Taskdeck.sln -c Release -m:1` — all 52 new tests pass
- [x] State transitions, containment actions, and operator override paths tested

### DIFF
--- a/backend/tests/Taskdeck.Application.Tests/Services/AbuseDetectionServiceTests.cs
+++ b/backend/tests/Taskdeck.Application.Tests/Services/AbuseDetectionServiceTests.cs
@@ -1,0 +1,500 @@
+using FluentAssertions;
+using Moq;
+using Taskdeck.Application.Interfaces;
+using Taskdeck.Application.Services;
+using Taskdeck.Domain.Enums;
+using Xunit;
+
+namespace Taskdeck.Application.Tests.Services;
+
+public class AbuseDetectionServiceTests
+{
+    private readonly AbuseDetectionSettings _settings;
+    private readonly AbuseDetectionState _state;
+    private readonly Mock<ILlmUsageRecordRepository> _usageRecordsMock;
+
+    public AbuseDetectionServiceTests()
+    {
+        _settings = new AbuseDetectionSettings
+        {
+            Enabled = true,
+            SuspiciousSignalThreshold = 3,
+            RestrictedSignalThreshold = 6,
+            BlockedSignalThreshold = 10,
+            VelocityRequestsPerHourThreshold = 120,
+            VelocityTokensPerHourThreshold = 200_000
+        };
+        _state = new AbuseDetectionState();
+        _usageRecordsMock = new Mock<ILlmUsageRecordRepository>();
+    }
+
+    private AbuseDetectionService CreateService() => new(_settings, _state, _usageRecordsMock.Object);
+
+    [Fact]
+    public async Task RecordSignal_ShouldReturnNoChange_WhenDisabled()
+    {
+        _settings.Enabled = false;
+        var service = CreateService();
+        var userId = Guid.NewGuid();
+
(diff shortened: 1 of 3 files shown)


Chris0Jeky:
## Self-review

### What I checked
- All 16 new/modified files in the diff
- Architecture layer boundaries (Domain has no infra deps, Application depends only on Domain)
- Clean Architecture compliance: enums and entities in Domain, service/DTOs in Application, controller/DI in Api
- Test coverage across domain entities and application service

### Findings

**No blockers found.**

**Observations (non-blocking):**

1. **Controller authorization scope**: The `AbuseContainmentController` endpoints are guarded by `[Authorize]` but do not yet enforce operator/admin role checks for sensitive operations like `POST /api/abuse/actors/override`. This is consistent with the existing pattern on `LlmQuotaController` (which has a TODO comment about admin privileges). When the role system ships, these endpoints should be restricted to admin/operator roles.

2. **In-memory state not persisted**: `AbuseDetectionState` holds actor state and events in memory only, consistent with how `LlmKillSwitchSettings` works. If the process restarts, all abuse state is lost. This is acceptable for the current phase; persistence can be added as a follow-up if needed.

3. **Unused imports**: `AbuseContainmentController.cs` imports `Taskdeck.Application.Interfaces` and `Taskdeck.Domain.Enums` which are technically used transitively but could be cleaned up if the controller evolves.

4. **Override signal type edge case**: When overriding from Observe to Observe (same state), the service records `ManualEscalation` rather than `ManualOverride` because the comparison is `newState < previousState`. This is a minor semantic oddity but has no functional impact since both states are Observe. The test was adjusted to test the meaningful de-escalation case instead.

### Test results
- 25 domain tests pass (AbuseActor: 13, AbuseEvent: 7, plus 5 existing tests affected)
- 27 application service tests pass
- Full solution: 0 failures across all test projects including Architecture.Tests

PR #472, opened 2026-03-29
OPS: Harden Windows Git resolution and index.lock reliability

### LINKED ISSUE #121: OPS: Harden Windows Git resolution and index.lock reliability for local/agent workflows
## Problem
In this Windows environment, `git` can resolve to Cygwin instead of Git for Windows, which causes intermittent command failures (for example, signal/pipe errors). We also observed commit failures where Git could not create `.git/index.lock` due to permission-denied conditions.

## Why this matters
- Causes non-deterministic failures for both humans and coding agents.
- Increases risk of interrupted commits and stale lock handling mistakes.
- Creates friction for repository automation and local development reliability.

## Observed symptoms
- `git` resolves to an unexpected executable (`Cygwin`) in some sessions.
- Commit failure message indicates inability to create `.git/index.lock` (permission denied).

## Proposed implementation
1. Enforce Git executable precedence so `git` resolves to `C:\Program Files\Git\cmd\git.exe`.
2. Add local shell guardrails that warn when:
   - current `git` resolution is not Git for Windows
   - `.git/index.lock` exists while no git process is active
3. Keep concise fallback guidance in repo docs (`AGENTS.md`).
4. Validate and document final checks (`Get-Command git -All`, `where.exe git`, `git --version`).

## Acceptance criteria
- [ ] In a fresh PowerShell session, `git` resolves to Git for Windows first.
- [ ] No recurring `.git/index.lock` permission errors under normal commit flow.
- [ ] Startup warning surfaces actionable guidance when mismatch/lock conditions are detected.
- [ ] `AGENTS.md` includes a short fallback note for future contributors/agents.
- [ ] Verification output captured in issue comment before closure.

## Notes
When stale lock cleanup is needed, delete `.git/index.lock` only after confirming no git process is running.

Chris0Jeky:
## Summary
- Add `scripts/check-git-env.sh` diagnostic script that validates git resolves to Git for Windows (not Cygwin) and detects stale `.git/index.lock` files with no active git process
- Update `CLAUDE.md` and `AGENTS.md` Windows sections to reference the new script and add PATH remediation guidance

Closes #121

## Validation
- Script tested locally: confirms non-Cygwin git resolution and clean index.lock state
- Outputs actionable guidance when issues are detected (Cygwin path, stale lock with/without active processes)

## Test plan
- [ ] Run `bash scripts/check-git-env.sh` on a Windows machine with Git for Windows — should pass
- [ ] Simulate Cygwin git by temporarily aliasing — script should warn
- [ ] Create a fake `.git/index.lock` with no git processes — script should detect and advise removal
- [ ] Verify CLAUDE.md and AGENTS.md changes are accurate and consistent

### DIFF
--- a/scripts/check-git-env.sh
+++ b/scripts/check-git-env.sh
@@ -27,11 +27,11 @@ else
     echo "[INFO]  git path:    $GIT_PATH"
     echo "[INFO]  git version: $GIT_VERSION"
 
-    # Detect Cygwin git — its path typically contains /cygdrive/ or /usr/bin/
-    # and its version string includes "cygwin".
+    # Detect Cygwin or MSYS2 (non-MinGW) git — path typically contains
+    # /cygdrive/ or /usr/bin/. Git for Windows (MinGW) resolves to /mingw64/bin/git.
     case "$GIT_PATH" in
         /cygdrive/*|/usr/bin/git)
-            echo "[WARN]  git appears to be Cygwin git ($GIT_PATH)."
+            echo "[WARN]  git appears to be Cygwin or MSYS2 (non-MinGW) git ($GIT_PATH)."
             echo "        This can cause signal errors and path translation issues."
             echo "        Fix: add 'C:\\Program Files\\Git\\cmd' to the FRONT of your PATH,"
             echo "        or set alias git='\"C:\\Program Files\\Git\\cmd\\git.exe\"' in your shell profile."
@@ -55,7 +55,9 @@ fi
 # ---------------------------------------------------------------------------
 # In worktrees, .git is a file pointing to the real git dir.
 # Use git rev-parse to find the actual git directory.
-GIT_DIR="$(git -C "$REPO_DIR" rev-parse --git-dir 2>/dev/null || echo "$REPO_DIR/.git")"
+# --absolute-git-dir (Git 2.13+) returns an absolute path, avoiding the bug
+# where a relative ".git" would resolve against CWD instead of REPO_DIR.
+GIT_DIR="$(git -C "$REPO_DIR" rev-parse --absolute-git-dir 2>/dev/null || echo "$REPO_DIR/.git")"
 LOCK_FILE="$GIT_DIR/index.lock"
 
 if [ -f "$LOCK_FILE" ]; then


Chris0Jeky:
## Fresh Adversarial Review

### Critical Issues

**1. `GIT_DIR` relative path bug breaks lock-file detection when using explicit repo argument**

`git -C "$REPO_DIR" rev-parse --git-dir` returns a **relative** path (`.git`) even when `$REPO_DIR` is absolute. The script then checks `-f "$GIT_DIR/index.lock"`, which resolves relative to the current working directory, not relative to `$REPO_DIR`.

Reproduction:
```bash
cd /tmp
bash /path/to/scripts/check-git-env.sh /path/to/repo
# Lock check looks at /tmp/.git/index.lock instead of /path/to/repo/.git/index.lock
```

Fix: Prepend `$REPO_DIR/` when the result is relative, or use `git -C "$REPO_DIR" rev-parse --absolute-git-dir` (available since Git 2.13).

### Minor Issues

**2. File mode should be 755, not 644**

The script is `100644` but is a shell script intended for direct execution. While `bash scripts/check-git-env.sh` works regardless, the conventional mode for shell scripts is `100755`. This matters if someone tries `./scripts/check-git-env.sh`.

**3. Comment says "Cygwin" but `/usr/bin/git` also catches MSYS2's own git package**

The case pattern `/usr/bin/git` will also match MSYS2's non-MinGW git (installed via `pacman -S git` in MSYS2 shell, as opposed to the MinGW git at `/mingw64/bin/git`). The comment should note this is detecting "Cygwin or MSYS2 (non-MinGW)" git, since the practical guidance is the same but the diagnostic message would otherwise be confusing to someone who knows they don't have Cygwin installed.

### Observations

- **CI is fully green.** All 13 required checks pass.
- **Self-review was thorough.** The worktree fix (commit 8ca8896d) was a good catch, though it introduced the relative-path bug described above.
- **Script is read-only/diagnostic.** No destructive actions, which is good.
- **`set -euo pipefail` with accumulated `EXIT_CODE`**: The error handling pattern works because all potentially-failing commands are wrapped with `|| true`. This is correct but slightly unusual -- future editors should be careful not to add bare commands that could fail.
- **`pgrep -a git` false positives**: Already noted in self-review. The conservative behavior (reporting lock as potentially legitimate) is the right call.
- **`tasklist //FI` double-slash**: Correct for Git Bash's MSYS path translation. Would break on native bash (e.g., Linux), but `tasklist` wouldn't exist there anyway so the `pgrep` fallback handles it.
- **Acceptance criteria**: All five criteria from issue #121 are addressed. The first criterion ("In a fresh PowerShell session, git resolves to Git for Windows first") is partially met -- the script _detects_ the problem but doesn't _enforce_ resolution. The issue text says "enforce" but the implementation chose "detect and advise," which is arguably more appropriate for a diagnostic script. This seems intentional.

### Verdict

**One real bug to fix** (the relative `GIT_DIR` path), **one minor convention fix** (file mode), and **one comment accuracy tweak**. Otherwise solid work -- the script is well-structured, handles edge cases thoughtfully, and the doc updates are consistent. Fix the critical issue and this is ready to merge.

PR #525, opened 2026-03-29
fix: render markdown in chat assistant messages (#518)

### LINKED ISSUE #518: BUG: Chat messages render raw markdown as plaintext — no markdown rendering
## Summary
Assistant messages in the Automation Chat view are not rendered as HTML markdown. Raw markdown tokens (`###`, `--`, `**bold**`, list markers) appear as literal text characters in the chat bubble.

## Reproduction
1. Open Chat, start a session
2. Ask any question that produces a response with headings, bold text, or lists
3. Observe the response — all markdown syntax is visible as plaintext

## Expected
Markdown rendered to HTML: bold text appears bold, `###` headings are styled headings, list items are properly indented.

## Actual
Raw `### Heading`, `**bold**`, `--` etc. visible as literal characters.

## Fix
Add a markdown-to-HTML renderer (e.g., `marked` or `markdown-it`) to the chat message renderer component. Apply `v-html` with sanitized output (DOMPurify or equivalent).

_Discovered in Sonnet manual testing session 2026-03-29. See `docs/analysis/2026-03-29_manual_testing_consolidated_findings.md` (BUG-M4)._

Chris0Jeky:
## Summary
- Adds `marked` + `DOMPurify` dependencies for markdown-to-HTML rendering
- Assistant and System messages in Automation Chat now render formatted markdown (headings, lists, code blocks, bold/italic, etc.)
- User messages remain as plaintext `{{ }}` interpolation — they never touch `v-html`

## Fix
- `renderMarkdown(content)` calls `marked.parse()` (sync string in v17) then `DOMPurify.sanitize()` before binding with `v-html`
- `isAssistantOrSystemMessage()` uses the existing `normalizeChatRole()` helper to detect non-user roles — handles both numeric and string role values from the backend
- Scoped CSS added for `.td-message-content--markdown` to style paragraphs, headings, lists, code blocks, blockquotes, and links in both light/dark themes via CSS custom properties

## Security Note
- DOMPurify sanitization is applied **before** `v-html` on every render — no unsanitized content can reach the DOM
- User input path (`message.role === 'User'`) uses `{{ message.content }}` (Vue's auto-escaped text interpolation) — completely separate from `v-html`
- No other `v-html` introduction in this diff

## Closes
Closes #518

## Risk
Low — `v-html` only used for assistant/system messages after DOMPurify sanitization; user message path is unchanged.

### DIFF
--- a/frontend/taskdeck-web/src/utils/chat.ts
+++ b/frontend/taskdeck-web/src/utils/chat.ts
@@ -5,11 +5,11 @@ const chatSessionStatusByIndex = ['Active', 'Archived'] as const
 
 export function normalizeChatRole(value: ChatRoleValue): typeof chatRoleByIndex[number] {
   if (typeof value === 'number') {
-    return chatRoleByIndex[value] ?? 'Assistant'
+    return chatRoleByIndex[value] ?? 'User'
   }
 
   const found = chatRoleByIndex.find(v => v.toLowerCase() === value.toLowerCase())
-  return found ?? 'Assistant'
+  return found ?? 'User'
 }
 
 export function normalizeChatSessionStatus(value: ChatSessionStatusValue): typeof chatSessionStatusByIndex[number] {
--- a/frontend/taskdeck-web/src/views/AutomationChatView.vue
+++ b/frontend/taskdeck-web/src/views/AutomationChatView.vue
@@ -24,7 +24,7 @@ function renderMarkdown(content: string): string {
   if (!content) {
     return ''
   }
-  return DOMPurify.sanitize(marked.parse(content) as string)
+  return DOMPurify.sanitize(marked.parse(content, { async: false }))
 }
 
 const router = useRouter()


Chris0Jeky:
## Adversarial Security Review (Independent Pass)

### Verdict: REQUEST CHANGES

Two issues require attention before merge: one MAJOR (architectural risk worth fixing now) and two MINORs. No outright XSS blocker found, but the MAJOR is close enough to one that it should be corrected.

---

### Findings

#### MAJOR — `normalizeChatRole` falls back to `'Assistant'` for any unknown/invalid role

**File:** `frontend/taskdeck-web/src/utils/chat.ts`, `isAssistantOrSystemMessage()`

`normalizeChatRole` has this fallback:
```ts
return chatRoleByIndex[value] ?? 'Assistant'   // numeric out-of-range
return found ?? 'Assistant'                    // unknown string
```

So a message with `role: 99`, `role: "hacker"`, or any future role string not in the known list will normalize to `'Assistant'` and **route into `v-html` + `renderMarkdown()`**. If the backend ever stores or relays a malformed role value (e.g., from a compromised LLM response that injects into message metadata), a user-authored message body would hit `v-html`.

In practice DOMPurify is still the last line of defence, so this is not a live XSS today — but it's a logic inversion: unknown roles should fail-**closed** to the safe `{{ }}` path, not fail-open to `v-html`. The fix is a one-liner:

```ts
// utils/chat.ts  — change the fallback in normalizeChatRole from 'Assistant' to 'User'
return chatRoleByIndex[value] ?? 'User'
return found ?? 'User'
```

Or, alternatively, gate `isAssistantOrSystemMessage` with an explicit allowlist check without relying on the shared helper's fallback:
```ts
function isAssistantOrSystemMessage(message: ChatMessage): boolean {
  const role = normalizeChatRole(message.role)
  return role === 'Assistant' || role === 'System'
}
```
This is already correctly written — the problem is purely in the upstream helper's default. Change the fallback default to `'User'` (the safe path), since any message of unknown origin should be treated as user-generated.

#### MINOR — `marked.parse()` type is `string | Promise<string>`; the `as string` cast hides a correctness risk

**File:** `AutomationChatView.vue`, line 27:
```ts
return DOMPurify.sanitize(marked.parse(content) as string)
```

The installed `marked@17.0.5` type declaration at line 638 of `marked.d.ts` reads:
```ts
(src: string, options?: MarkedOptions | null): string | Promise<string>;
```

Without explicit `{ async: false }` in the options object, TypeScript cannot prove the return is synchronous — which is why the `as string` cast was necessary to silence the compiler. In the default configuration (no `walkTokens` extension), `marked.parse()` _is_ synchronous at runtime, but this is a fragile cast that:
1. Silences a real type warning without fixing the cause.
2. Would silently break if a `marked.use()` extension with an async `walkTokens` is ever added — the `v-html` binding would render `[object Promise]` instead of sanitized HTML.

**Recommended fix** — pin the sync overload explicitly:
```ts
import { marked, type MarkedOptions } from 'marked'

const SYNC_OPTIONS: MarkedOptions & { async: false } = { async: false }

function renderMarkdown(content: string): string {
  if (!content) return ''
  return DOMPurify.sanitize(marked.parse(content, SYNC_OPTIONS))
}
```
This removes the `as string` cast, the return type becomes `string`, and TypeScript enforces the contract going forward.

#### MINOR — Rendered `<a>` links open in the same tab; no `rel="noopener noreferrer"` on DOMPurify output

The CSS adds `color` and `underline` for links but the DOMPurify default configuration allows `<a href="...">` without restricting `target`. More importantly, DOMPurify by default **strips `target`** entirely, which means links open in the same tab — navigating the user away from the app. It does not add `rel="noopener noreferrer"` either.

This is not an XSS vector (DOMPurify correctly blocks `javascript:` hrefs), but it is a UX/security-hygiene issue: assistant-generated links should open in a new tab with `noopener noreferrer`. Fix with a DOMPurify hook:
```ts
DOMPurify.addHook('afterSanitizeAttributes', (node) => {
  if (node.tagName === 'A') {
    node.setAttribute('target', '_blank')
    node.setAttribute('rel', 'noopener noreferrer')
  }
})
```
Or configure this once at module scope rather than per-call.

---

### Security checks

| Check | Result |
|---|---|
| DOMPurify applied to HTML output (not markdown input) | **pass** — `DOMPurify.sanitize(marked.parse(content) as string)` correctly sanitizes the rendered HTML, not the raw markdown |
| User message content never reaches `v-html` | **conditional pass** — direct `role: 'User'` path uses `{{ }}` correctly, but `normalizeChatRole` fallback default routes unknown roles to `v-html` (see MAJOR above) |
| `marked.parse()` returns sync string (not Promise) at runtime | **pass at runtime** — marked@17.0.5 without async extensions is synchronous; however the `as string` cast suppresses a valid type warning (see MINOR) |
| No CSP violation introduced | **pass** — the backend CSP is `script-src 'self'` with no `unsafe-eval`; `v-html` injecting static HTML does not require script execution and DOMPurify strips event handlers and `<script>` tags |

### Correctness checks

| Check | Result |
|---|---|
| Role detection correctly gates `v-html` | **fail** — `normalizeChatRole` fallback is `'Assistant'` not `'User'`; unknown roles silently enter the `v-html` path |
| null/undefined content handled | **pass** — `if (!content) return ''` guards both null and empty string |
| Markdown CSS present for dark/light themes | **pass** — all selectors use `var(--td-*)` design tokens; both light and dark themes covered via the same custom property system |

### Build

| Check | Result |
|---|---|
| TypeScript clean | **pass** — `vue-tsc --noEmit` exits 0 with no errors (though the `as string` cast is what makes this pass; see MINOR above) |

---

### Summary

The core XSS chain (user input → `v-html`) is correctly blocked: user messages use `{{ }}`, and DOMPurify runs after `marked.parse()`. The MAJOR issue is that the role-gating logic has a fail-open fallback that could in theory route a message with an unexpected role through `v-html`. Fix: change `normalizeChatRole`'s default fallback from `'Assistant'` to `'User'`. The two MINORs (sync cast, link target) are worth addressing but are not blockers for initial merge if the MAJOR is fixed.

PR #568, opened 2026-03-29
fix: throttle board list polling (#510)

### LINKED ISSUE #510: BUG: Board list polling at ~3 requests/second causing visual thrashing
## Summary
The board list is being fetched approximately 3 times per second continuously, causing visual thrashing and wasting network/CPU resources.

## Reproduction
1. Navigate to any board
2. Open browser devtools → Network tab, filter by `/api/boards`
3. Observe repeated requests at ~3/second intervals

## Expected
Board list fetched once on mount, then updated via SignalR events or on explicit user action. No continuous polling loop.

## Actual
Continuous polling loop. Causes flickering and unnecessary re-renders across the board surface.

## Root Cause Hypothesis
A `setInterval` or reactive watcher in `boardStore` triggers repeated `fetchBoards` calls without a proper debounce or interval guard. Likely introduced with the realtime collaboration layer.

## Fix
1. Remove or dramatically increase the polling interval (≥30s with jitter minimum)
2. Rely on SignalR push events for board list updates
3. Add a network-request count assertion to the E2E smoke test to prevent regression

## Priority
**P1 — performance regression**

_Discovered in manual testing session 2026-03-29. See `docs/analysis/2026-03-29_manual_testing_consolidated_findings.md` (BUG-C3)._

Chris0Jeky:
## Summary
- Debounces `handleBoardMutation` in `useBoardRealtime.ts` with a 300 ms window so rapid-burst SignalR events coalesce into a single `fetchBoard` call
- Increases the SignalR fallback poll interval from 15 s to 30 s
- Adds a 5 s throttle guard on `fetchBoards` (board list) in `boardCrudStore.ts` to deduplicate mount-time requests from multiple views loading in parallel
- Closes #510

## Root cause
SignalR fires a `boardMutation` event for every individual board state change. When multiple mutations arrive in quick succession (e.g. bulk import, automation run, or rapid card moves), the previous `handleBoardMutation` handler issued a `fetchBoard` call for each event. `fetchBoard` itself dispatches 3 HTTP requests (`GET /api/boards/{id}`, `GET /api/boards/{id}/cards`, `GET /api/boards/{id}/labels`), which — multiplied by the burst rate — produced the observed `~3 req/s` flood visible in the Network tab filtered by `/api/boards`.

A secondary source was multiple views (BoardsListView, ActivityView, ReviewView, etc.) each calling `fetchBoards` on `onMounted` within the same navigation cycle, causing duplicate board-list requests.

## Fix
1. **`useBoardRealtime.ts`** — `handleBoardMutation` now schedules the `fetchBoard` call behind a 300 ms debounce. Each new event cancels the pending timer, so only one refresh fires at the end of a quiet window. The `mutationDebounceTimer` is cleared in `stop()` to prevent dangling callbacks on unmount.
2. **`boardCrudStore.ts`** — `fetchBoards` records `lastFetchBoardsAt` after a successful response and skips any repeat call within 5 s unless it carries a search or `includeArchived` filter (those are intentional targeted queries and always bypass the throttle).

## Test plan
- [x] `npx vitest --run` passes (1108 tests, 0 failures)
- [x] New test: `coalesces rapid burst mutation events into a single fetchBoard call` — 3 events in < 300 ms → exactly 1 `fetchBoard` invocation
- [x] Existing fallback-polling tests updated to reflect 30 s interval
- [x] Existing preserveSelection tests updated to advance fake clock past the throttle window between calls
- [ ] Network tab manual verification: navigate to a board, observe `/api/boards` requests settle to at most 1 per explicit action or SignalR quiet window

### DIFF
--- a/frontend/taskdeck-web/src/composables/useBoardRealtime.ts
+++ b/frontend/taskdeck-web/src/composables/useBoardRealtime.ts
@@ -68,16 +68,20 @@ export function createBoardRealtimeController(
     }, FALLBACK_POLL_INTERVAL_MS)
   }
 
+  const cancelMutationDebounce = () => {
+    if (mutationDebounceTimer !== null) {
+      clearTimeout(mutationDebounceTimer)
+      mutationDebounceTimer = null
+    }
+  }
+
   const handleBoardMutation = (event: BoardRealtimeEvent) => {
     if (!subscribedBoardId || event.boardId !== subscribedBoardId) {
       return
     }
 
     // Debounce: cancel any pending refresh scheduled by a prior burst event.
-    if (mutationDebounceTimer !== null) {
-      clearTimeout(mutationDebounceTimer)
-      mutationDebounceTimer = null
-    }
+    cancelMutationDebounce()
 
     mutationDebounceTimer = setTimeout(() => {
       mutationDebounceTimer = null
@@ -145,6 +149,10 @@ export function createBoardRealtimeController(
   }
 
   const joinBoard = async (boardId: string) => {
+    // Cancel any debounced mutation fetch from the previous board so it cannot
+    // fire against the newly-subscribed boardId after subscribedBoardId changes.
+    cancelMutationDebounce()
+
     const hubConnection = ensureConnection()
 
     if (hubConnection.state === HubConnectionState.Disconnected) {
@@ -187,10 +195,7 @@ export function createBoardRealtimeController(
 
   const stop = async () => {
     stopFallbackPolling()
-    if (mutationDebounceTimer !== null) {
-      clearTimeout(mutationDebounceTimer)
-      mutationDebounceTimer = null
-    }
+    cancelMutationDebounce()
     editingCardId = null
 
     if (!connection) {
--- a/frontend/taskdeck-web/src/tests/composables/useBoardRealtime.spec.ts
+++ b/frontend/taskdeck-web/src/tests/composables/useBoardRealtime.spec.ts
@@ -209,6 +209,30 @@ describe('createBoardRealtimeController', () => {
     expect(mockConnection.invoke).toHaveBeenCalledWith('JoinBoard', 'board-2')
   })
 
+  it('cancels a pending debounce timer when switching boards', async () => {
+    // Regression: a board-A mutation event with a debounce timer pending must
+    // not fire fetchBoard after subscribedBoardId has advanced to board-B.
+    vi.useFakeTimers()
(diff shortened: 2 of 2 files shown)


Chris0Jeky:
## Adversarial review findings

**One genuine bug found and fixed.**

### Bug: stale debounce timer fires against wrong board after `switchBoard()` (fixed)

`handleBoardMutation` schedules a 300 ms debounce timer that closes over `subscribedBoardId` at *fire time* (via `const boardId = subscribedBoardId` inside the timeout). When `switchBoard('board-2')` is called while a board-A mutation debounce is pending:

1. `subscribedBoardId` is updated to `'board-2'` by `joinBoard`
2. The stale timer fires 300 ms later
3. `subscribedBoardId` is now `'board-2'` — the `!subscribedBoardId` guard passes
4. `const boardId = subscribedBoardId` captures `'board-2'`
5. `fetchBoard('board-2')` is called — triggered by a board-A mutation event

Fix: extracted `cancelMutationDebounce()` and called it at the top of `joinBoard()` so any pending timer is always discarded before `subscribedBoardId` changes. Also applied in `stop()` for symmetry (replaces the inline timer-clear block).

Regression test added: `cancels a pending debounce timer when switching boards`.

### Checklist items that are fine

- **Stale closure on `fetchBoard` reference**: Not an issue — `options.fetchBoard` is captured once at construction and is stable.
- **Unmount safety**: `stop()` correctly clears the debounce timer and sets `mutationDebounceTimer = null` before any async work. The `.finally()` on in-flight fetches is safe to run post-stop since it only touches `refreshInFlight` on the abandoned controller.
- **300 ms debounce value**: Appropriate. The debounce coalesces burst events; a single mutation after 300 ms of silence triggers exactly one fetch. Tested.
- **30 s fallback poll**: Acceptable trade-off. SignalR is the primary update path; 30 s fallback is documented in the PR and matches the comment rationale.
- **Throttle bypass for filtered queries**: Expected and intentional; not abusable from UI (search/filter are explicit user actions).
- **`lastFetchBoardsAt` on logout**: `lastFetchBoardsAt` is module-level within the factory closure, which is re-created per `createPinia()` call. The PR author verified this resets correctly on each Pinia instance.
- **Fake timers in tests**: All debounce/throttle tests use `vi.useFakeTimers()`. The burst-coalescing test asserts `fetchBoard` is called exactly once (not zero), so it cannot false-pass.

All 1109 frontend unit tests pass after the fix.

PR #590, opened 2026-03-30
Add board keyboard card movement and move-to action menu

### LINKED ISSUE #248: UI-06: Board drag/drop premium behavior with keyboard move alternatives
## Context
Tracked by: #242

Premium board interactions require smooth drag behavior plus non-drag alternatives for accessibility and trust.

Sources:
- `LIBRARIES_AND_STACK_DECISIONS.md`
- `ACCESSIBILITY_KEYBOARD_SPEC.md`
- `PERFORMANCE_RESPONSIVENESS_FRONTEND.md`

## Scope
- Harden drag/drop UX for board cards (library adoption or hardened current path).
- Ensure drag handles and small action targets meet minimum target-size expectations.
- Provide deterministic keyboard/click alternatives for card movement operations.
- Keep drag interactions responsive and avoid layout-thrash regressions.

## Depends on
- #242
- #244

Coordinates with
- #45
- #160

## Acceptance Criteria
- [ ] Card movement is possible without drag-only interaction paths.
- [ ] Drag interaction responsiveness is stable under representative board loads.
- [ ] Accessibility and keyboard move paths are explicitly tested.

## Verification
- `cd frontend/taskdeck-web; npx vitest --run`
- `cd frontend/taskdeck-web; npx playwright test --reporter=line`
- `cd frontend/taskdeck-web; npm run build`

Chris0Jeky:
## Summary

- **Keyboard card movement**: Alt+Left/Right moves the selected card to the previous/next column. Alt+Up/Down reorders the card within its current column. Focus follows the card after moves.
- **Card action menu**: Cards now display a "Move to..." button alongside the drag handle (visible when the board has 2+ columns). Clicking opens a dropdown listing all columns, with the current column disabled. Provides a click-based, touch-friendly alternative to drag-and-drop.
- **Drag handle clarity**: Drag handle is now part of a structured action bar with flex layout, maintaining minimum 40px touch target size.
- **Focus management**: After keyboard-driven moves, the card element is scrolled into view and focused via `nextTick` so keyboard users never lose their place.
- **Existing behavior preserved**: All existing drag-and-drop, navigation shortcuts, and card interactions remain unchanged. Plain arrow navigation explicitly sets `alt: false` to prevent conflicts with new Alt+Arrow movement shortcuts.

Closes #248

## Test plan

- [x] Typecheck passes (`npm run typecheck`)
- [x] All 1374 unit tests pass (`npx vitest --run`)
- [x] Production build succeeds (`npm run build`)
- [ ] Manual: select a card with j/k, then Alt+Right moves it to next column
- [ ] Manual: select a card, then Alt+Left moves it to previous column
- [ ] Manual: Alt+Up/Down reorders card within column
- [ ] Manual: focus stays on the card after keyboard moves
- [ ] Manual: click the move button on a card, verify dropdown shows all columns
- [ ] Manual: select a column from the dropdown, card moves there
- [ ] Manual: existing drag-and-drop still works
- [ ] Manual: existing j/k/h/l navigation still works with no Alt conflicts
- [ ] Manual: keyboard help dialog (?) shows new Card Movement section

### DIFF
--- a/frontend/taskdeck-web/src/composables/useBoardKeyboardNav.ts
+++ b/frontend/taskdeck-web/src/composables/useBoardKeyboardNav.ts
@@ -159,10 +159,10 @@ export function useBoardKeyboardNav(
   }
 
   /**
-   * Move the currently selected card to the next column (right).
-   * The card is placed at the end of the target column.
+   * Shared logic for moving the selected card to an adjacent column.
+   * Private helper — not exported.
    */
-  async function moveCardToNextColumn() {
+  async function moveCardToAdjacentColumn(direction: 'next' | 'previous') {
     if (!boardId) return
     const columns = sortedColumns.value
     if (columns.length === 0) return
@@ -171,22 +171,26 @@ export function useBoardKeyboardNav(
     const currentColumn = columns[selectedColumnIndex.value]
     if (!currentColumn) return
 
-    const nextColIndex = selectedColumnIndex.value + 1
-    if (nextColIndex >= columns.length) return
+    const targetColIndex =
+      direction === 'next'
+        ? selectedColumnIndex.value + 1
+        : selectedColumnIndex.value - 1
 
-    const nextColumn = columns[nextColIndex]
-    if (!nextColumn) return
+    if (targetColIndex < 0 || targetColIndex >= columns.length) return
+
+    const targetColumn = columns[targetColIndex]
+    if (!targetColumn) return
 
     const cards = boardStore.cardsByColumn.get(currentColumn.id) || []
     const card = cards.find((c) => c.id === selectedCardId.value)
     if (!card) return
 
-    const targetCards = boardStore.cardsByColumn.get(nextColumn.id) || []
+    const targetCards = boardStore.cardsByColumn.get(targetColumn.id) || []
     const targetPosition = targetCards.length
 
     try {
-      await boardStore.moveCard(boardId(), card.id, nextColumn.id, targetPosition)
-      selectedColumnIndex.value = nextColIndex
+      await boardStore.moveCard(boardId(), card.id, targetColumn.id, targetPosition)
+      selectedColumnIndex.value = targetColIndex
       await focusSelectedCard()
     } catch {
       // moveCard already surfaces toast errors via the store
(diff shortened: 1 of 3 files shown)


Chris0Jeky:
## Adversarial Self-Review

### Accessibility

1. **Move menu keyboard navigation**: The move-to dropdown uses `role="menu"` and `role="menuitem"` correctly and items are native `<button>` elements (focusable by default). However, there is no arrow-key navigation *within* the menu itself (e.g., ArrowDown to cycle through items). Users can Tab through items, but ARIA menu pattern typically expects arrow-key navigation. **Severity: Low** -- Tab works, and this is a progressive enhancement. Could be improved in a follow-up.

2. **Escape to close move menu**: The move menu uses a click-away overlay for closing, but pressing Escape while the menu is open will trigger the BoardView `closeOpenUi` handler instead of closing the move menu first. The menu will close if the user clicks away, but Escape may close the wrong thing. **Severity: Medium** -- Worth fixing.

3. **Screen reader announcement after move**: After a keyboard move, the card is focused, but there is no ARIA live region announcing the move result (e.g., "Card moved to column X"). The toast from the store's moveCard provides visual feedback. **Severity: Low** -- toast exists, explicit announcement is a nice-to-have.

### Focus Management

4. **Focus after move-to menu action**: When using the click-based "Move to..." menu, after the card moves to a new column, the card's DOM node is destroyed and recreated in the new column. The focus is not explicitly restored. The keyboard path handles this via `focusSelectedCard()`, but the menu path does not. **Severity: Medium** -- Should add focus restoration after menu-driven moves.

### Event Handler Conflicts

5. **Alt+Arrow on macOS**: On macOS, Alt+Left/Right in some browsers triggers browser history navigation (back/forward). The `preventDefault` in `useKeyboardShortcuts` should prevent this, and the default `preventDefault !== false` is true. **Confirmed safe**: `preventDefault` is called before the action.

6. **Modifier key matching order**: Alt+Arrow shortcuts are placed before plain Arrow shortcuts in the array, and plain Arrow shortcuts set `alt: false` explicitly. The `useKeyboardShortcuts` handler breaks on first match. This is correct -- Alt+Arrow will match first and the plain Arrow won't fire. **Confirmed safe**.

### Existing Behavior

7. **Drag handle layout change**: The drag handle was changed from `w-[calc(100%+1rem)]` (full-width) to `flex-1` inside a flex container. The visual result should be similar but the handle is now slightly narrower when the move button is present (which takes ~40px). The remaining handle area is still well above the 44px minimum touch target height. **Severity: Low** -- acceptable tradeoff.

### Issues to Fix

- **Issue A (Medium)**: Escape key does not close the move menu. The move menu should intercept Escape before `closeOpenUi` fires.
- **Issue B (Medium)**: No focus restoration after menu-driven moves (only keyboard moves call `focusSelectedCard`).

### Verdict

Two medium-severity issues identified. Fixing both now.

PR #630, opened 2026-03-31
Fix board horizontal scrollbar visibility and column responsiveness

### LINKED ISSUE #620: UX-20: Board — fix horizontal scrollbar visibility and column responsiveness
## Context
Manual testing (2026-03-31) found the board's horizontal scrollbar appears below the visible viewport. Users must scroll down to reach it. Analysis: `docs/analysis/2026-03-31_manual_testing_ux_feedback.md` §8c.

## Problem
- `BoardCanvas.vue` line 88: `height: calc(100vh - 120px)` doesn't account for actual header height (toolbar + action rail + help callout + filter panel)
- Columns fixed at `width: 20rem` (320px) x 5 = 1600px + gaps exceeds viewport
- Native scrollbar sits below the container bottom edge
- 4px-wide scrollbar (`style.css`) is very thin and hard to grab

## Solution
1. Fix container height: measure actual header height dynamically, or increase offset to ~180px
2. Responsive columns: `min-width: 16rem; flex: 1; max-width: 22rem` so columns shrink when possible
3. Ensure `overflow-y: hidden` on canvas (card overflow is within columns, which already have `overflow-y: auto`)
4. Consider scroll arrows at board edges for discoverability
5. Increase scrollbar width from 4px to 8px for easier grabbing

## Acceptance Criteria
- [ ] Horizontal scrollbar always visible within the viewport (no vertical scroll needed to reach it)
- [ ] Board with 5 columns fits without scrolling on 1440px+ viewports
- [ ] Scrollbar is visually discoverable (not invisible-thin)
- [ ] Column card lists still scroll vertically within columns

Chris0Jeky:
## Summary
- **Board canvas height**: Increase `calc(100vh - 120px)` offset to `180px` so the horizontal scrollbar is visible within the viewport instead of being clipped below
- **Overflow-y: hidden** on canvas container — vertical card overflow is already handled by individual column lanes (`overflow-y: auto`)
- **Responsive columns**: Replace fixed `width: 20rem` with `min-width: 16rem; flex: 1; max-width: 22rem` so columns shrink on narrow viewports and grow when space allows
- **Scrollbar width**: Increase from 4px to 8px for easier grabbing on non-trackpad devices

Closes #620

## Test plan
- [ ] Open board view at various viewport widths (1024px, 1280px, 1920px) — columns should shrink/grow responsively
- [ ] Horizontal scrollbar should be visible within the viewport when columns overflow
- [ ] Scrollbar should be 8px wide and easy to grab
- [ ] Cards within columns should scroll vertically within the column, not the canvas
- [ ] Empty board state still renders correctly
- [ ] Dark mode scrollbar colors unchanged

### DIFF
--- a/frontend/taskdeck-web/src/components/board/ColumnLane.vue
+++ b/frontend/taskdeck-web/src/components/board/ColumnLane.vue
@@ -332,9 +332,8 @@ function handleCardDragOver(event: DragEvent) {
 <style scoped>
 /* ── Column Lane — token-based layout ── */
 .td-column-lane {
-  flex-shrink: 0;
   min-width: 16rem;
-  flex: 1;
+  flex: 1 0 16rem;
   max-width: 22rem;
   border-radius: var(--td-radius-lg);
   padding: var(--td-space-5);


Chris0Jeky:
## Adversarial Self-Review

### Issue found and fixed
- **flex shorthand override**: The initial commit had `flex-shrink: 0` followed by `flex: 1`, but `flex: 1` expands to `flex: 1 1 0%` which silently overrides `flex-shrink` back to 1. This would let columns shrink below their min-width on very narrow viewports before the horizontal scrollbar appeared. Fixed in commit 0dfb155d by using `flex: 1 0 16rem` (grow, no-shrink, basis=16rem) and removing the now-redundant `flex-shrink: 0`.

### Acceptance criteria check
1. Container height offset increased from 120px to 180px -- DONE
2. Responsive columns with min-width/flex/max-width -- DONE (16rem/1 0 16rem/22rem)
3. overflow-y: hidden on canvas -- DONE
4. Scrollbar width increased from 4px to 8px -- DONE
5. Horizontal scrollbar visible within viewport -- addressed by height fix

### Remaining considerations
- **Magic number 180px**: The 180px offset is still a hardcoded approximation of header + toolbar height. If the header layout changes this will drift again. The issue suggested "~180px or measure dynamically" -- we went with the static value for simplicity. A future improvement could use CSS `dvh` units or a JS-measured approach.
- **Firefox scrollbar**: The scrollbar styles use `::-webkit-scrollbar` which only affects Chromium/Safari. Firefox uses `scrollbar-width` and `scrollbar-color` CSS properties. This is a pre-existing gap, not introduced by this PR.
- **Global scrollbar change**: The 4px-to-8px scrollbar change is in `style.css` and affects ALL scrollbars app-wide, not just the board canvas. This is intentional per the issue but worth noting -- if any other view relied on the thin scrollbar aesthetic, it will look different.

### Security / correctness
- No logic changes, no data flow changes, CSS only -- no security concerns.

PR #667, opened 2026-04-01
Add board metrics dashboard (ANL-01)

### LINKED ISSUE #77: ANL-01: Board metrics dashboard (throughput, cycle time, WIP, blocked trends)
Source:
- `docs/WIP/FutureExpansionAndImprovementsChecklist.pdf` (2.5 Board Metrics Dashboard)
- `docs/InReview/HUMAN/01_PRODUCT_THESIS.md` (capture friction + trust outcomes)
- `docs/InReview/HUMAN/05_USER_RESEARCH_PLAYBOOK.md` (early validation metrics)
- `docs/InReview/outreach-crm/06_IMPLEMENTATION_PLAN_WAVES.md` (outreach scoreboard candidates)

Horizon:
- Future expansion tranche (post current Phase 4 backlog)

Depends on:
- #71
- #199

Scope:
- Define baseline analytics model for board throughput, cycle time, WIP, blocked duration, and completion trends.
- Add thesis-aligned capture metrics:
  - capture action -> persisted artifact latency
  - capture artifact -> proposal-ready latency
  - proposal review/apply funnel metrics (accept/reject/abandon)
- Add outreach scoreboard profile hooks for future expansion:
  - outreach actions sent
  - reply-rate proxy
  - call-booked proxy
  - referral-requested proxy
- Expose query endpoints with filtering (date range, board, label, actor).
- Build dashboard UI with reusable chart/table modules.

Acceptance Criteria:
- Dashboard renders core metrics with filter persistence and loading/error states.
- Metric definitions are documented and stable across backend/frontend.
- Capture/trust metrics are defined with clear semantics and data source contracts.
- Outreach scoreboard metrics are defined with stable data-source assumptions (deferred profile, no runtime claim before implementation).
- Query paths are tested for correctness and regression safety.
- At least one E2E path verifies filter-to-visualization behavior.

Suggested branch:
- `feature/board-metrics-dashboard`

Chris0Jeky:
## Summary

Implements GitHub Issue #77 — ANL-01: Board metrics dashboard with throughput, cycle time, WIP, and blocked card trends.

### Backend
- **BoardMetricsDtos.cs**: Query/response DTOs for throughput data points, cycle time entries, WIP snapshots, and blocked card summaries
- **IBoardMetricsService / BoardMetricsService**: Application-layer service that computes metrics from existing card, column, and audit data. Determines "done" column by rightmost position, calculates throughput by day, cycle time (creation to done), WIP per column, and blocked card duration
- **MetricsController**: `GET /api/metrics/boards/{boardId}?from=&to=&labelId=` — authenticated, authorization-checked endpoint with sensible defaults (last 30 days)
- **DI registration**: Wired with IAuthorizationService for read-access enforcement
- **12 unit tests**: Covering validation (empty IDs, invalid date ranges), permission denied, board not found, empty board metrics, WIP computation, blocked card counting, and static computation methods

### Frontend
- **metrics.ts types**: TypeScript interfaces matching backend DTOs
- **metricsApi.ts**: HTTP client with proper URL encoding and query parameter handling
- **metricsStore.ts**: Pinia store with loading/error/data states and demo mode support
- **MetricsView.vue**: Full dashboard with:
  - Board selector and date range picker (7/14/30/60/90 days)
  - Summary cards (throughput total, avg cycle time, WIP count, blocked count)
  - CSS bar chart for throughput trend
  - Horizontal bar chart for WIP by column (with WIP limit violation highlighting)
  - Tables for cycle time details and blocked cards
  - Loading spinner, error state with retry, empty states
- **Route**: `/workspace/metrics` with lazy loading
- **Navigation**: "Metrics" item in sidebar (workbench secondary, available in guided/agent modes)
- **4 API client tests**: URL construction, query params, encoding

## Test plan
- [x] Backend builds: `dotnet build backend/Taskdeck.sln -c Release` — 0 errors
- [x] Backend tests: `dotnet test backend/Taskdeck.sln -c Release -m:1` — all pass (including 12 new metrics tests)
- [x] Frontend typecheck: `npm run typecheck` — clean
- [x] Frontend build: `npm run build` — successful
- [x] Frontend tests: `npx vitest --run` — 1495 tests pass (4 new metricsApi tests)
- [ ] E2E: Manual verification of dashboard rendering with seeded data
- [ ] Verify filter persistence across board selection changes

Closes #77

### DIFF
--- a/backend/src/Taskdeck.Application/Services/BoardMetricsService.cs
+++ b/backend/src/Taskdeck.Application/Services/BoardMetricsService.cs
@@ -1,7 +1,9 @@
+using System.Text.RegularExpressions;
 using Taskdeck.Application.DTOs;
 using Taskdeck.Application.Interfaces;
 using Taskdeck.Domain.Common;
 using Taskdeck.Domain.Entities;
+using Taskdeck.Domain.Enums;
 using Taskdeck.Domain.Exceptions;
 
 namespace Taskdeck.Application.Services;
@@ -9,16 +11,20 @@ namespace Taskdeck.Application.Services;
 public class BoardMetricsService : IBoardMetricsService
 {
     private readonly IUnitOfWork _unitOfWork;
-    private readonly IAuthorizationService? _authorizationService;
+    private readonly IAuthorizationService _authorizationService;
 
-    public BoardMetricsService(IUnitOfWork unitOfWork)
-        : this(unitOfWork, authorizationService: null)
+    /// <summary>
+    /// Well-known column names that indicate "done" status, checked case-insensitively.
+    /// The first match (by highest position among matches) wins.
+    /// </summary>
+    private static readonly string[] DoneColumnNames =
     {
-    }
+        "done", "complete", "completed", "finished", "closed", "shipped", "released"
+    };
 
     public BoardMetricsService(
         IUnitOfWork unitOfWork,
-        IAuthorizationService? authorizationService)
+        IAuthorizationService authorizationService)
     {
         _unitOfWork = unitOfWork;
         _authorizationService = authorizationService;
@@ -35,19 +41,16 @@ public class BoardMetricsService : IBoardMetricsService
         if (actingUserId == Guid.Empty)
             return Result.Failure<BoardMetricsResponse>(ErrorCodes.ValidationError, "Acting user ID cannot be empty");
 
-        if (query.From >= query.To)
+        if (query.From > query.To)
             return Result.Failure<BoardMetricsResponse>(ErrorCodes.ValidationError, "From date must be before To date");
 
         // Enforce read permission
-        if (_authorizationService != null)
-        {
-            var canRead = await _authorizationService.CanReadBoardAsync(actingUserId, query.BoardId);
-            if (!canRead.IsSuccess)
-                return Result.Failure<BoardMetricsResponse>(canRead.ErrorCode, canRead.ErrorMessage);
+        var canRead = await _authorizationService.CanReadBoardAsync(actingUserId, query.BoardId);
+        if (!canRead.IsSuccess)
+            return Result.Failure<BoardMetricsResponse>(canRead.ErrorCode, canRead.ErrorMessage);
 
-            if (!canRead.Value)
-                return Result.Failure<BoardMetricsResponse>(ErrorCodes.Forbidden, "You do not have permission to view metrics for this board");
-        }
+        if (!canRead.Value)
+            return Result.Failure<BoardMetricsResponse>(ErrorCodes.Forbidden, "You do not have permission to view metrics for this board");
 
         // Verify board exists
         var board = await _unitOfWork.Boards.GetByIdAsync(query.BoardId, cancellationToken);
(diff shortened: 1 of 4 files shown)


Chris0Jeky:
## Adversarial Review -- PR #667 (ANL-01: Board Metrics)

### BLOCKING

**1. Cycle time uses `UpdatedAt` as a proxy for "moved to done" -- fundamentally unreliable**
`BoardMetricsService.ComputeThroughput` and `ComputeCycleTime` (lines 98-103, 127-131) filter cards with `c.UpdatedAt >= from && c.UpdatedAt <= to` to detect when a card reached the "done" column. However, `UpdatedAt` is mutated by *any* entity change (title edit, description change, block/unblock, label add) -- not just column moves. This means:
- A card already in "done" that gets its title edited will be re-counted in throughput.
- A card in "done" that had a label added after the query date range will be excluded from cycle time even though it was completed within the range.
- Cycle time `(c.UpdatedAt - c.CreatedAt)` will overestimate if the card was edited after being moved to done.

**Fix**: Use the audit log (`AuditEntry`) to find the actual timestamp when `ColumnId` changed to the done column, or add a dedicated `CompletedAt` field to `Card`. Without this, the numbers are misleading.

**2. "Done" column detection assumes rightmost column is always "done"**
`BoardMetricsService.cs` line 68: `var doneColumn = columns.OrderByDescending(c => c.Position).FirstOrDefault()`. A board with columns [Backlog(0), In Progress(1), Review(2), Archive(3)] would treat "Archive" as done. A single-column board would treat that column as "done", counting all cards as completed. There is no way for users to designate which column is "done".

**Fix**: Either let users mark a column as "done" (domain change), or use a heuristic that checks the column name (fragile but better), or document this assumption prominently and add a parameter for the done column ID.

### SHOULD-FIX

**3. Interface placed in `Services` namespace instead of `Interfaces`**
`IBoardMetricsService` is defined in `Taskdeck.Application.Services` namespace. The project convention is that interfaces live in `Taskdeck.Application.Interfaces` (see `IAuthorizationService.cs` in that namespace). The architecture tests may not catch this because the implementation also lives in `Services`, but it breaks the layer convention.

**4. Constructor allows `null` authorization service -- auth bypass in production possible**
`BoardMetricsService` has a public constructor `BoardMetricsService(IUnitOfWork unitOfWork)` that sets `_authorizationService = null`, and line 42 checks `if (_authorizationService != null)` before enforcing permissions. If someone accidentally registers the single-arg constructor in DI, *all* authorization is silently skipped. The null-auth constructor should be `internal` or removed, with the service always requiring `IAuthorizationService`.

**5. All cards loaded into memory regardless of date range**
`GetByBoardIdAsync` (line 59) loads *every* card for the board, then filters in-memory. For boards with 10,000+ cards, this is an O(n) memory load per request with no pagination. The throughput/cycle-time computations only need cards in the done column within the date range. Consider adding a repository method that filters by column and date range at the SQL level.

**6. `From >= To` check should probably be `From > To` to allow same-day queries**
Line 38: `query.From >= query.To` rejects queries where from == to. A user selecting "today" as both start and end date gets a validation error. This should likely be `query.From > query.To`.

**7. Frontend: metricsStore swallows demo mode silently**
In `metricsStore.ts` lines 17-23, demo mode sets `metrics.value = null` and returns. The user sees "No metrics data available" with no indication that this feature is unavailable in demo mode. Other stores (e.g., audit) handle demo mode by populating mock data. At minimum, show a message like "Metrics are not available in demo mode."

**8. Blocked duration estimate is unreliable**
`ComputeBlocked` (line 172) estimates blocked duration as `DateTimeOffset.UtcNow - c.UpdatedAt`. But `UpdatedAt` is touched by any change, not just `Block()`. If a blocked card has its title edited, the blocked duration resets to near-zero. Same root cause as issue #1.

### MINOR

**9. Throughput groups by `c.UpdatedAt.Date` -- timezone-unaware**
`UpdatedAt.Date` uses the server's local timezone to bucket. A card completed at 23:59 UTC and another at 00:01 UTC would be in the same or different buckets depending on server timezone. Consider using `c.UpdatedAt.UtcDateTime.Date` consistently.

**10. `maxThroughput` / `maxWipCount` computed uses `Math.max(...spread)` -- stack overflow risk**
In `MetricsView.vue` lines 75-76 and 80-81, `Math.max(...metrics.value.throughput.map(...))` will throw a `RangeError` if the array has more than ~65,000 elements (engine-dependent). Unlikely for daily granularity but worth noting.

**11. No `aria-live` on dashboard content transitions**
The dashboard uses `v-if`/`v-else-if` chains for loading/error/data states, but only the loading state has `aria-live="polite"`. Screen readers may miss transitions to error or data states.

**12. Frontend types use `string` for all IDs**
Backend uses `Guid` but frontend types use `string`. This is fine and consistent with the rest of the codebase, but `boardId` in `MetricsQuery` accepts any string -- there is no UUID format validation on the client side before sending to the API.

---

**Summary**: The two BLOCKING issues (#1 and #2) mean the metric values displayed to users will frequently be incorrect. The core computation logic needs to use a more reliable signal for "card completed" than `UpdatedAt`, and the done-column detection needs to be explicit rather than assumed. I recommend fixing these before merge or clearly documenting the limitations in the UI itself.

Issue #618, opened 2026-03-31
LLM-03: Spike — LLM tool-calling / function-calling architecture for chat


Chris0Jeky:
## Context
Manual testing (2026-03-31) and strategic analysis identified that the current regex-based instruction extraction is fundamentally limited. The LLM needs actual tools to query and operate on board state. Analysis: `docs/analysis/2026-03-31_manual_testing_ux_feedback.md` §7.

## Problem
Current chat architecture:
- Static regex-based intent classification (8 hardcoded patterns)
- LLM asked to output JSON instructions but has no ability to query state
- Can't handle "move all cards to next stage" — requires board enumeration
- Can't handle multi-step reasoning about board state

## Spike Scope
Design and prototype function-calling / tool-use architecture:

### Read Tools (no review gate needed)
- `list_boards()` → board names and IDs
- `get_board_columns(board_id)` → columns with positions
- `list_cards_in_column(board_id, column_id)` → cards with IDs, titles, labels
- `search_cards(query)` → fuzzy card search
- `get_card_details(card_id)` → full card info

### Write Tools (produce proposals, review-first)
- `propose_card_create(title, column, description?)`
- `propose_card_move(card_id, target_column)`
- `propose_card_archive(card_id)`
- `propose_bulk_move(column_from, column_to)` → multi-card proposal

### Key Constraint
Write tools MUST produce proposals, never direct mutations. This preserves GP-06 (Review-First Automation Safety).

## Deliverables
- [ ] Architecture document: tool inventory, safety boundaries, provider format differences
- [ ] Flow diagrams: user → LLM → tool call → response → proposal
- [ ] Prototype: one read tool (`list_cards_in_column`) working end-to-end
- [ ] Mock provider: deterministic tool simulation
- [ ] Risk assessment: token cost, latency impact, error handling

## Related
- #576 — Conversational refinement loop
- LLM-04 — MCP server (builds on this foundation)

PR #669, opened 2026-04-01
LLM-06: Phase 1 -- Read tools, orchestrator, and provider tool-calling extension

### LINKED ISSUE #649: LLM-06: Phase 1 — Read tools, orchestrator, and provider tool-calling extension
## Overview

Phase 1 of LLM tool-calling implementation. The LLM can dynamically query board state instead of relying on static context.

**Tracker**: #647 | **Spike**: `docs/spikes/SPIKE_618_COMPLETED.md` §10 Phase 1

## Deliverables

1. `TaskdeckToolSchema` record types in Application layer (provider-agnostic tool definition)
2. `LlmToolCompletionResult` and `ToolCallRequest`/`ToolCallResult` records
3. `ILlmProvider.CompleteWithToolsAsync()` on interface with `NotSupportedException` default
4. `OpenAiLlmProvider.CompleteWithToolsAsync()` — tool schemas → `tools[]`, `tool_calls` parsing
5. `GeminiLlmProvider.CompleteWithToolsAsync()` — tool schemas → `functionDeclarations`, `functionCall` parsing
6. `MockLlmProvider.CompleteWithToolsAsync()` — pattern-based dispatch table for deterministic simulation
7. `ToolCallingChatOrchestrator` wrapping ChatService — multi-turn loop (max 5 rounds, 60s total timeout)
8. Read tool executors: `list_board_columns`, `list_cards_in_column`, `get_card_details`, `search_cards`, `get_board_labels`
9. Updated system prompt for tool-calling mode (§6 of spike doc)
10. SignalR `ToolStatusEvent` integration for intermediate states
11. Unit tests for orchestrator, tool executors, mock dispatch, schema conversion per provider

## Verification

"What cards are in my Backlog?" produces a dynamic, accurate response using `list_cards_in_column` — verified with Mock provider, tested with live provider if keys available.

**Prototype first**: Get `list_cards_in_column` working end-to-end before implementing other read tools. It exercises the full flow: schema definition → provider conversion → multi-turn loop → mock simulation.

## Key Design Constraints

- Read tools are `ToolRiskLevel.Low` and execute directly (no review gate)
- Read tools are scoped to the current board session by default
- Mock provider uses pattern-matching dispatch table (see spike §7)
- Orchestrator must handle: tool call timeout (30s/round), total timeout (60s), provider errors (graceful degradation to single-turn)
- Existing `CompleteAsync` path must continue working for non-tool-calling flows

## Estimated Scope

1–2 weeks

## Blocked By

None — can start immediately

## Unblocks

- #TBD (Phase 2: Write tools)
- #576 (Conversational refinement — partial, read tools enable clarification)

Chris0Jeky:
## Summary

Implements the Phase 1 deliverables from the LLM tool-calling spike (#618), enabling the LLM to dynamically query board state through function-calling instead of relying on static context. Closes #649.

### What's new

- **Tool-calling record types** (`TaskdeckToolSchema`, `ToolCallRequest`, `ToolCallResult`, `LlmToolCompletionResult`) in the Application layer as provider-agnostic abstractions
- **`ILlmProvider.CompleteWithToolsAsync()`** with a default `NotSupportedException` implementation so existing providers continue working unchanged
- **OpenAI `CompleteWithToolsAsync`** -- converts schemas to `tools[]` format, parses `tool_calls` responses, handles tool result messages with `"tool"` role
- **Gemini `CompleteWithToolsAsync`** -- converts schemas to `functionDeclarations`, parses `functionCall` parts, generates synthetic call IDs when Gemini omits them
- **Mock `CompleteWithToolsAsync`** -- pattern-based dispatch table (`MockToolCallDispatcher`) with deterministic simulation via `MockToolResults`
- **5 read tool executors**: `list_board_columns`, `list_cards_in_column`, `get_card_details`, `search_cards`, `get_board_labels` -- all board-scoped with structured error responses
- **`ToolCallingChatOrchestrator`** -- multi-turn loop (max 5 rounds, 30s/round, 60s total timeout) with graceful degradation
- **SignalR `ToolStatusEvent`** integration for intermediate status messages ("Looking up cards in Backlog...")
- **Tool-calling system prompt** for tool-aware mode (replaces instruction-extraction prompt)
- **ChatService integration** -- board-scoped sessions delegate to orchestrator when available; falls back to single-turn path when no tools are called or provider doesn't support tool-calling

### Key design decisions

- Read tools are `ToolRiskLevel.Low` and execute directly (no review gate), per GP-06
- When the LLM responds without calling any tools, the existing proposal creation path is preserved (prevents regressions)
- Tool schemas use `additionalProperties: false` for OpenAI strict mode compatibility
- All new code stays within Clean Architecture boundaries (record types in Application, executors in Application, provider changes in Application, SignalR notifier in API)

## Test plan

- [x] 67 new unit tests passing: dispatch patterns, mock results, mock provider, OpenAI parsing, Gemini parsing, orchestrator loop, schema validation
- [x] All 2053 existing tests pass (357 Domain + 1260 Application + 424 API + 4 CLI + 8 Architecture)
- [x] `dotnet build backend/Taskdeck.sln -c Release` succeeds with 0 errors
- [x] Existing Chat API integration test `CreateSession_And_SendActionableMessage_ShouldReturnProposalReference` still passes
- [ ] Manual verification: "What cards are in my Backlog?" produces dynamic response via Mock provider

### DIFF
--- a/backend/src/Taskdeck.Application/Services/GeminiLlmProvider.cs
+++ b/backend/src/Taskdeck.Application/Services/GeminiLlmProvider.cs
@@ -357,9 +357,16 @@ public class GeminiLlmProvider : ILlmProvider
                     var callId = functionCall.TryGetProperty("id", out var idEl)
                         ? idEl.GetString() ?? $"gemini-{Guid.NewGuid():N}"[..16]
                         : $"gemini-{Guid.NewGuid():N}"[..16];
-                    var args = functionCall.TryGetProperty("args", out var argsEl)
-                        ? argsEl.Clone()
-                        : JsonDocument.Parse("{}").RootElement.Clone();
+                    JsonElement args;
+                    if (functionCall.TryGetProperty("args", out var argsEl))
+                    {
+                        args = argsEl.Clone();
+                    }
+                    else
+                    {
+                        using var emptyDoc = JsonDocument.Parse("{}");
+                        args = emptyDoc.RootElement.Clone();
+                    }
 
                     toolCalls.Add(new ToolCallRequest(callId, name, args));
                 }
--- a/backend/src/Taskdeck.Application/Services/OpenAiLlmProvider.cs
+++ b/backend/src/Taskdeck.Application/Services/OpenAiLlmProvider.cs
@@ -325,7 +325,7 @@ public class OpenAiLlmProvider : ILlmProvider
                 {
                     var callId = tc.TryGetProperty("id", out var idEl) ? idEl.GetString() ?? "" : "";
                     var funcName = "";
-                    var funcArgs = default(JsonElement);
+                    JsonElement funcArgs;
 
                     if (tc.TryGetProperty("function", out var func))
                     {
@@ -334,9 +334,20 @@ public class OpenAiLlmProvider : ILlmProvider
                         {
                             // OpenAI returns arguments as a JSON string that needs re-parsing
                             var argsStr = argsEl.GetString() ?? "{}";
+                            if (string.IsNullOrWhiteSpace(argsStr)) argsStr = "{}";
                             using var argsDoc = JsonDocument.Parse(argsStr);
                             funcArgs = argsDoc.RootElement.Clone();
                         }
+                        else
+                        {
+                            using var emptyDoc = JsonDocument.Parse("{}");
+                            funcArgs = emptyDoc.RootElement.Clone();
+                        }
+                    }
+                    else
+                    {
+                        using var emptyDoc = JsonDocument.Parse("{}");
+                        funcArgs = emptyDoc.RootElement.Clone();
                     }
 
                     calls.Add(new ToolCallRequest(callId, funcName, funcArgs));
--- a/backend/src/Taskdeck.Application/Services/ToolCallingChatOrchestrator.cs
+++ b/backend/src/Taskdeck.Application/Services/ToolCallingChatOrchestrator.cs
@@ -89,13 +89,15 @@ public sealed class ToolCallingChatOrchestrator
             catch (NotSupportedException)
             {
                 _logger.LogWarning("Provider does not support tool calling; falling back to single-turn.");
-                return BuildDegradedResult(provider, model);
+                return BuildDegradedResult(provider, model, totalTokensUsed, round, toolCallLog,
+                    "Provider does not support tool calling.");
             }
             catch (Exception ex)
             {
                 _logger.LogError("Tool-calling round {Round} failed. {ExceptionSummary}",
                     round, SensitiveDataRedactor.SummarizeException(ex));
-                return BuildDegradedResult(provider, model);
+                return BuildDegradedResult(provider, model, totalTokensUsed, round, toolCallLog,
+                    "Tool-calling round failed due to an internal error.");
             }
 
             totalTokensUsed += llmResult.TokensUsed;
@@ -261,17 +263,21 @@ public sealed class ToolCallingChatOrchestrator
             DegradedReason: "Maximum tool-calling rounds exceeded.");
     }
 
-    private static ToolCallingResult BuildDegradedResult(string provider, string model)
+    private static ToolCallingResult BuildDegradedResult(
+        string provider, string model,
+        int tokensUsed = 0, int round = 0,
+        List<ToolCallLogEntry>? log = null,
+        string? reason = null)
     {
         return new ToolCallingResult(
             Content: null,
-            TokensUsed: 0,
+            TokensUsed: tokensUsed,
             Provider: provider,
             Model: model,
-            Rounds: 0,
-            ToolCallLog: new List<ToolCallLogEntry>(),
+            Rounds: round,
+            ToolCallLog: log ?? new List<ToolCallLogEntry>(),
             IsDegraded: true,
-            DegradedReason: "Tool calling is not available; falling back to single-turn.");
+            DegradedReason: reason ?? "Tool calling is not available; falling back to single-turn.");
     }
 
     private static string TruncateForLog(string content)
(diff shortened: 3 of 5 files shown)


Chris0Jeky:
## Adversarial Review -- PR #669 (LLM-06 Tool-Calling Phase 1)

Reviewed all 26 changed files, the spike doc, existing Copilot/Gemini comments, and the self-review. Findings below, sorted by severity.

---

### BLOCKING (fixed in 453ed723)

**B1. OpenAI parser: `funcArgs = default(JsonElement)` produces Undefined ValueKind**
`OpenAiLlmProvider.cs` line 328

When OpenAI omits `function.arguments` (which can happen with malformed or edge-case responses), `funcArgs` stays as `default(JsonElement)` with `ValueKind == Undefined`. Downstream, the orchestrator's `BuildStatusMessage` calls `arguments.TryGetProperty(...)` which throws `InvalidOperationException` on Undefined elements. The mock dispatcher and tool executors also call `TryGetProperty` on Arguments.

**Fix**: Initialize `funcArgs` to a parsed `{}` empty object in all branches where `arguments` is missing or empty/whitespace.

**B2. `BuildDegradedResult` discards accumulated tokens and tool call log**
`ToolCallingChatOrchestrator.cs` lines 264-275

When a later round fails (NotSupportedException after round 1, or general Exception mid-conversation), `BuildDegradedResult` returns `TokensUsed: 0`, `Rounds: 0`, `ToolCallLog: new List<>()`. This silently discards:
- All tokens consumed in earlier rounds (breaks quota enforcement)
- The tool call log (breaks auditability -- a core review-first requirement)
- The round counter (masks how far the orchestration progressed)

**Fix**: `BuildDegradedResult` now accepts optional `tokensUsed`, `round`, `log`, and `reason` parameters. Both catch blocks pass accumulated state.

---

### SHOULD-FIX (fixed in 453ed723)

**S1. Gemini parser: undisposed `JsonDocument.Parse("{}")` leak**
`GeminiLlmProvider.cs` line 362

`JsonDocument.Parse("{}").RootElement.Clone()` in a ternary expression creates a `JsonDocument` that is never disposed. Under load this adds GC pressure. Changed to a proper `using` block.

**S2. Dead `JsonOptions` field in `ReadToolSchemas`**
`ReadToolSchemas.cs` line 10-13

`JsonOptions` is defined but never referenced. Removed.

---

### SHOULD-FIX (not yet fixed -- deferred or needs discussion)

**S3. OpenAI missing `strict: true` for structured output mode**
The spike doc (Section 3.3) explicitly shows `"strict": true` in the OpenAI tool schema wire format for strict mode compatibility. The implementation does not include it. Without `strict: true`, OpenAI may hallucinate extra fields or omit required ones. This is not a runtime crash but reduces output reliability for live providers.

**S4. Double LLM call for non-tool messages (noted in self-review, confirmed)**
When `_toolCallingOrchestrator != null && session.BoardId.HasValue`, every message first goes through `CompleteWithToolsAsync`. If the LLM responds without tool calls, the result is discarded and `CompleteAsync` is called again. For Mock provider this is free; for live providers this doubles latency and cost. The self-review correctly notes this for Phase 3.

**S5. `GetCardDetailsExecutor` loads all board cards to resolve short ID**
Line 26 does `GetByBoardIdAsync(boardId)` then scans in memory. For boards with many cards this is O(n) with full entity graph loading. A repository method for short-ID prefix lookup would be better, but this is acceptable for Phase 1 read-only tools.

**S6. `ListBoardColumnsExecutor` loads full card entities just for counting**
Same pattern -- `GetByBoardIdAsync` loads all cards with labels just to `GroupBy` and count. A lightweight count query would be more appropriate but is not blocking for Phase 1.

---

### MINOR

**M1. Gemini synthetic call ID truncation is fragile**
`$"gemini-{Guid.NewGuid():N}"[..16]` works but the substring length (16) cuts into the GUID portion. If the "gemini-" prefix changes length, this silently breaks. Consider using a fixed-length suffix or just the full string.

**M2. `ToolCallingSystemPrompt` is `internal static` but referenced from tests via the public API**
Not a problem now, but if tests need to assert on the prompt content directly, the visibility will need to change.

**M3. No test for cancellation token propagation through orchestrator**
There's a test for `NotSupportedException` and timeout, but no test that verifies `ct.ThrowIfCancellationRequested()` propagates correctly when the caller cancels mid-orchestration.

---

### Architecture & Safety Verification (all pass)

- **Clean Architecture**: All new types in Application layer. Tool executors depend on `IUnitOfWork` interfaces, not Infrastructure. SignalR notifier correctly in Api layer implementing Application-layer interface. Architecture tests (8/8) pass.
- **Thread safety**: `ToolExecutorRegistry` is scoped (new per request). `MockToolCallDispatcher` is static with compiled regexes (thread-safe). Orchestrator is scoped with no shared mutable state.
- **Existing flow preservation**: When orchestrator is null or non-board-scoped, the single-turn `CompleteAsync` path is untouched. When orchestrator runs but no tools are called, falls through to single-turn. Integration test `CreateSession_And_SendActionableMessage_ShouldReturnProposalReference` still passes.
- **5-round limit**: Enforced in orchestrator for-loop. Tested by `ExecuteAsync_MaxRoundsExceeded_ReturnsExhaustedResult`.
- **60s timeout**: Enforced via `Stopwatch` check at top of each round. Per-round 30s timeout via linked `CancellationTokenSource`.
- **Error degradation**: All provider errors, tool execution errors, and unknown tools return structured degraded results. Never crashes the chat flow.
- **JSON serialization**: snake_case via `JsonNamingPolicy.SnakeCaseLower` across all tool results. Schemas use `additionalProperties: false`.
- **Test coverage**: 67 new tests covering dispatch patterns, mock results, mock provider, OpenAI parsing, Gemini parsing, orchestrator loop, schema validation, timeout, error cases.

### Fixes applied

Commit `453ed723` fixes B1, B2, S1, S2. All 2053 tests pass (357 Domain + 1260 Application + 424 API + 4 CLI + 8 Architecture). Build clean with 0 errors.

PR #697, opened 2026-04-02
Show human-readable targets in proposal diff view

### LINKED ISSUE #682: REV-02: Proposal diff view should show human-readable operation targets instead of raw IDs
## Context
Live browser testing on 2026-04-02 confirmed a presentation mismatch inside `/workspace/review`.

## Problem
`View Diff` currently renders operation lines like:

- `0. create card:559961d9-f4eb-ad3c-3617-9615ed43b5ef`
- `1. create card:d2d8c7d2-f385-759b-ece0-ddd108b24153`

Elsewhere on the same review card, the UI already uses human-readable summaries such as `Affected cards` and readable proposal titles. The diff panel is the outlier and forces users to reason about raw IDs instead of intended board changes.

## Repro
1. Open `/workspace/review`.
2. Expand any capture-linked proposal with `View Diff`.
3. Compare the diff content with the rest of the proposal card.

## Expected
The diff view should describe operations in human-readable terms, for example by card title, column name, or action summary.

## Actual
The diff view exposes raw operation target IDs, which are lower-signal than the surrounding UI and make review harder.

## Proposed Fix
1. Translate operation targets into readable labels before rendering the diff.
2. Preserve raw identifiers only as secondary technical detail if needed.
3. Add frontend review coverage for readable diff output.

## Acceptance Criteria
- [ ] `View Diff` shows human-readable operation targets.
- [ ] Raw IDs are not the primary review surface for normal proposal review.
- [ ] Review diff output remains consistent with the card's readable summary and affected-entity sections.

Chris0Jeky:
## Summary
- Proposal diff view now shows human-readable operation descriptions instead of raw GUIDs
- Card IDs resolved to titles, column IDs to names via batch lookups per board
- Create operations show `Create card "Fix login bug" in column "To Do"` instead of `create card:559961d9-...`
- Move operations show `Move card "Setup CI" to column "In Progress"` instead of `move card:abc123-...`
- Falls back gracefully when boardId is null or entity lookups fail
- Frontend diff area now has an "Operation details" label, ARIA region, and word-wrapping

Closes #682

## Test plan
- [x] Backend: 3 new tests covering create-card, move-card, and null-boardId fallback
- [x] Frontend: new test verifying readable diff content and label rendering
- [x] All 1763 backend tests pass
- [x] All 1523 frontend tests pass
- [x] Frontend typecheck passes
- [ ] Manual: diff shows "Create card «title»" instead of raw GUID
- [ ] Manual: move operations show target column name
- [ ] Manual: falls back gracefully if card/column deleted

### DIFF
--- a/backend/src/Taskdeck.Application/Services/AutomationProposalService.cs
+++ b/backend/src/Taskdeck.Application/Services/AutomationProposalService.cs
@@ -296,11 +296,7 @@ public class AutomationProposalService : IAutomationProposalService
                 foreach (var card in cards)
                     cardTitles[card.Id] = card.Title;
             }
-            catch (OperationCanceledException)
-            {
-                throw; // Respect cancellation
-            }
-            catch
+            catch (Exception ex) when (ex is not OperationCanceledException)
             {
                 // Non-critical: if lookups fail, fall back to IDs
             }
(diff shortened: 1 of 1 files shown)


Chris0Jeky:
## Self-review checklist

### Correctness
- [x] N+1 queries: batch loads cards and columns per board in 2 queries, builds dictionaries
- [x] Null/deleted entities: `TryGetValue` returns false gracefully; falls back to generic labels
- [x] XSS: diff rendered via Vue `{{ }}` interpolation (auto-escaped), no `v-html`
- [x] Cancellation: `OperationCanceledException` re-thrown (fixed in follow-up commit)

### Accessibility
- [x] Diff area has `role="region"` and `aria-label`
- [x] "Operation details" label provides visual context

### Edge cases
- [x] Proposal without boardId: skips entity lookups, still extracts title from parameters JSON
- [x] Proposal with stored DiffPreview: returns stored value unchanged (existing behavior)
- [x] Unknown action types: fall through without column context suffix (only `move`/`create` get one)
- [x] Malformed parameters JSON: `ExtractGuidParameter` catches `JsonException` gracefully

### Minor notes
- `ExtractGuidParameter` re-parses JSON for each property lookup. Acceptable for an on-demand diff endpoint. Could be optimized with a single parse if performance profiling shows it matters.
- Loading all cards for a board could be slow for very large boards, but matches existing patterns in `OperationHandlerRegistry`.

Chris0Jeky:
## Adversarial Review -- PR #697

### Findings

| # | Severity | Issue | Status |
|---|----------|-------|--------|
| 1 | **CRITICAL** | **ID fallback dropped**: When card/column name resolution fails (deleted entity, lookup miss), the diff line shows only `"Move card"` with no identifier at all. Old code showed `targetType:targetId`. Reviewers lose the ability to identify which specific item will change. | **FIXED** |
| 2 | **HIGH** | **Column context never appended for composite action types**: `actionNorm = operation.ActionType.ToLowerInvariant()` yields `"card.move"` for composite types, but `actionNorm is "move"` checks for exact match. `HumanizeActionVerb` already handles this correctly, so using `verb == "Move"` is the right approach. | **FIXED** |
| 3 | **MEDIUM** | **Card title lookup applied to non-card targets**: For `reorder column` operations, the code tried resolving `operation.TargetId` against `cardTitles`. While unlikely to collide (random GUIDs), this is semantically wrong and could produce incorrect labels if IDs ever overlap. | **FIXED** |
| 4 | **MEDIUM** | **Column GUID hidden on lookup miss**: When `columnNames` dict misses the column ID (deleted column), no column info shown at all. Now falls back to showing the raw column GUID. | **FIXED** |
| 5 | **MEDIUM** | **Bare catch swallows OperationCanceledException**: The `catch(OCE) { throw; } catch { }` pattern is correct but verbose. Replaced with `catch (Exception ex) when (ex is not OperationCanceledException)` which is idiomatic and simpler. | **FIXED** |
| 6 | **LOW** | **Redundant GUID extraction**: `ExtractGuidParameter` manually parsed string GUIDs before calling `TryGetGuid`, which already handles string values. Simplified. | **FIXED** |

### Bot comment triage

All three bots (Gemini, Codex, Copilot) flagged real issues. Addressed:
- **Gemini**: Cancellation handling (finding 5), targetType guard (finding 3), verb-based column check (finding 2), redundant GUID parse (finding 6) -- all fixed.
- **Codex**: ID fallback regression (finding 1), cancellation propagation (finding 5) -- both fixed.
- **Copilot**: ID fallback regression (finding 1), cancellation (finding 5), performance concern about loading full aggregates -- the performance concern is noted but deferred (matches existing patterns, and this is an on-demand endpoint).

### Not fixed (acceptable)

- **Performance**: Loading all cards/columns for a board via `GetByBoardIdAsync` loads full aggregates. This matches existing patterns in `OperationHandlerRegistry`. For an on-demand diff endpoint this is acceptable. A lightweight `GetNamesByBoardIdAsync` could be added later if profiling shows issues on large boards.

### Verification

- Backend: 1,879 tests pass (Domain 114, Application 1,318, Api 433, Cli 4, Architecture 8)
- Frontend: 1,523 tests pass, typecheck clean
- All fixes in commit `113f9eaa`

PR #805, opened 2026-04-09
PLAT-02: Distributed caching strategy and cache-invalidation semantics

### LINKED ISSUE #85: PLAT-02: Distributed caching strategy and cache-invalidation semantics
Source:
- `docs/WIP/Scaling and Hardening Taskdeck (Vue 3 + ASP.NET Core) – Comprehensive Guide.pdf` (Caching guidance)

Horizon:
- Future expansion tranche (post current Phase 4 backlog)

Depends on:
- #53
- #79

Scope:
- Identify hot read paths and define cache-aside policy (TTL, key strategy, invalidation triggers).
- Implement initial cache integration with environment-safe fallback to no-cache mode.
- Add observability for hit/miss/error rates and stale data diagnostics.

Acceptance Criteria:
- At least one high-value read path is cache-backed with measurable latency improvement.
- Cache invalidation semantics are documented and test-covered.
- Cache failures degrade safely without breaking correctness.
- Metrics emitted for cache effectiveness and failure analysis.

Suggested branch:
- `feature/distributed-cache-semantics`

Chris0Jeky:
## Summary

- Add `ICacheService` abstraction in Application layer with cache-aside pattern for hot read paths
- Implement `InMemoryCacheService` (thread-safe ConcurrentDictionary with periodic sweep), `RedisCacheService` (StackExchange.Redis with lazy connection and safe degradation), and `NoOpCacheService`
- Apply caching to `BoardService.ListBoardsAsync` (60s TTL) and `GetBoardDetailAsync` (120s TTL) with invalidation on create/update/delete/archive
- All cache failures degrade safely without breaking correctness — no exceptions propagated to callers
- Cache hit/miss/error metrics emitted via structured logging
- Configurable via `appsettings.json` (`Cache:Provider`, `Cache:RedisConnectionString`, TTLs)
- ADR-0023 documents the cache-aside strategy, alternatives considered, and consequences
- 32 new tests: 12 InMemoryCacheService, 5 NoOpCacheService, 15 BoardServiceCache behavior

## Test plan

- [x] InMemoryCacheService: get/set, expiry, remove, prefix removal, overwrite, idempotent operations, metrics logging, key prefix isolation, concurrent access
- [x] NoOpCacheService: all operations are safe no-ops
- [x] BoardServiceCacheTests: cache hit returns cached value (no DB query), cache miss queries DB and populates cache, filtered queries bypass cache, not-found results not cached, invalidation on create/update/delete, fallback when no cache service
- [x] Full backend test suite passes (3400+ tests)

Closes #85


Chris0Jeky:
## Adversarial Self-Review Findings

### Fixed in follow-up commit (9473f355)

1. **Log noise (MEDIUM)**: `LogCacheMetric` was emitting at `Information` level for every cache hit/miss. In high-traffic scenarios this would flood logs. **Fixed**: changed to `Debug` level in both `InMemoryCacheService` and `RedisCacheService`.

2. **Incomplete Dispose (LOW)**: `InMemoryCacheService.Dispose()` disposed the sweep timer but did not clear the `ConcurrentDictionary`, leaving cached data in memory. **Fixed**: added `_cache.Clear()` to `Dispose()`.

3. **Unused field (LOW)**: `_logger` field was declared in `BoardService` but never referenced in any method. **Fixed**: removed the field and its constructor parameter.

### Acknowledged but acceptable

4. **Stale cache for shared-board users (MEDIUM)**: When User B updates a board that User A has access to, only User B's board list cache is invalidated (invalidation uses `board.OwnerId`). User A sees stale data until TTL expires (60s default). **Accepted**: 60s staleness is tolerable for the local-first use case. Cross-user invalidation would require tracking all users with board access, adding complexity disproportionate to the staleness window. The ADR documents this trade-off.

5. **Board detail cached without per-user scope (LOW)**: `GetBoardDetailAsync(Guid id)` caches using `board:{id}:detail` regardless of who requested it. This is correct because the auth check in `GetBoardDetailAsync(Guid id, Guid actingUserId)` happens **before** the cache read, so unauthorized users never reach the cache. The non-authorized overload is only used when `_authorizationService is null` (MCP stdio mode, tests).

6. **Redis SCAN in RemoveByPrefixAsync (LOW)**: `server.Keys()` uses SCAN internally which is production-safe but O(n) on the full keyspace. Current usage (board list key removal by exact key, not prefix) does not hit this code path. The method exists for future use and the SCAN approach is the recommended Redis pattern.

7. **`_cacheSettings!` null-forgiving operator (LOW)**: Used in three places where `canCache` is true, which guarantees `_cacheService is not null`. The `_cacheSettings` field defaults to `new CacheSettings()` in the constructor, so it is always non-null. The `!` is technically safe but is a minor code smell. Accepted.

### Thread safety verification

- `InMemoryCacheService`: All operations use `ConcurrentDictionary` which is thread-safe. The sweep timer callback iterates a snapshot (safe for concurrent modification). The `Count` property is eventually consistent per `ConcurrentDictionary` semantics.
- `RedisCacheService`: `Lazy<T>` ensures thread-safe connection initialization. All Redis operations are async and StackExchange.Redis `ConnectionMultiplexer` is designed for concurrent access.
- `BoardService` cache operations: Cache reads and writes are not atomic with database operations, which is by design in cache-aside. Worst case: a concurrent write invalidates the cache between the DB read and the cache write, resulting in one extra DB query on the next request.

### Security check

- Cache keys contain GUIDs (board IDs, user IDs) which are not predictable. No sensitive data (passwords, tokens) is cached. Board DTOs contain only public board metadata (name, description, dates, column counts).

PR #811, opened 2026-04-09
OBS-02: Error tracking and product analytics foundation

### LINKED ISSUE #549: OBS-02: Error tracking and product analytics foundation
## Summary

Set up error tracking and privacy-respecting product analytics for the hosted cloud instance.

## Acceptance Criteria

- [ ] Sentry (or similar) for frontend and backend error tracking
- [ ] Plausible or Umami (self-hosted) for web analytics (no cookies, no PII)
- [ ] Opt-in product telemetry for activation/retention metrics
- [ ] Telemetry dashboard showing: captures/day, proposals/day, activation rate, retention

## Implementation Notes

- Sentry has a free tier for open source projects
- Plausible can be self-hosted for free, or hosted for $9/mo
- Product telemetry must be opt-in with clear disclosure
- Builds on #341 (product telemetry taxonomy)

## Related
- #341 Product telemetry taxonomy

## Parent
#531

Chris0Jeky:
## Summary

Implements issue #549 — error tracking, web analytics, and product telemetry foundation for Taskdeck.

- **Backend Sentry integration**: Config-gated Sentry SDK (`Sentry.AspNetCore`) with hard privacy guardrails (`SendDefaultPii` always `false`, auth headers stripped from breadcrumbs). Disabled by default.
- **Backend telemetry event service**: `ITelemetryEventService` with event name validation against the taxonomy (`noun.verb` format from `docs/product/TELEMETRY_TAXONOMY.md`), opt-in guard, and batch size limits.
- **Backend telemetry API**: `GET /api/telemetry/config` (anonymous, returns client config) and `POST /api/telemetry/events` (authenticated, records batched events).
- **Frontend telemetry store**: Pinia store with opt-in consent (persisted in localStorage), event buffering with 30s periodic flush, and dual-gate activation (user consent AND server config).
- **Frontend analytics script injection**: Composable for Plausible/Umami self-hosted analytics — cookie-free, no PII, script injected/removed based on consent state.
- **Frontend consent UI**: Telemetry & Privacy section in Settings page with clear disclosure of what is/isn't collected.
- **Documentation**: `docs/ops/OBSERVABILITY_SETUP.md` covering all configuration options.

All integrations are **disabled by default** and require explicit configuration (server) + opt-in (user).

Closes #549

## Key design decisions

- **Dual-gate telemetry**: Both server-side config AND user consent must be enabled for any telemetry to flow
- **No PII enforcement**: `SendDefaultPii` forced to `false` in code (not just config), auth headers stripped from Sentry breadcrumbs
- **Event validation**: Backend validates event names match `noun.verb` taxonomy before recording
- **Config endpoint is anonymous**: DSNs are public identifiers, not secrets — no auth needed for the frontend to fetch config

## Test plan

- [x] Backend: 25 new TelemetryEventService unit tests (validation, opt-in guard, batch limits)
- [x] Backend: 13 new API integration tests (DI registration, config endpoint, events endpoint, default-disabled verification)
- [x] Backend: All 3800 existing tests pass (including architecture boundary tests)
- [x] Frontend: 22 new telemetry store tests (consent, isActive, emit, flush, config loading, sentryAvailable, analyticsConfig)
- [x] Frontend: 3 new telemetry API client tests
- [x] Frontend: All 1922 existing tests pass
- [x] Frontend: TypeScript typecheck passes


Chris0Jeky:
## Independent Adversarial Review (Round 2) -- Privacy & Security Focus

### CRITICAL Findings

#### C1: No BeforeSend handler -- PII leaks through exception messages [CRITICAL]
**File:** `backend/src/Taskdeck.Api/Extensions/SentryRegistration.cs`

`SendDefaultPii = false` prevents Sentry from auto-collecting usernames, IPs, and request bodies. However, it does **nothing** to scrub PII from exception messages themselves. Any thrown exception with a message like "User john@example.com not found" or "Invalid token for user admin" will be sent verbatim to Sentry.

There is no `SetBeforeSend` handler to sanitize `SentryEvent.Message` or exception stack data. This is the most common PII leak vector in Sentry deployments.

**Recommendation:** Add a `SetBeforeSend` callback that at minimum:
- Strips email-pattern strings from event messages and exception values
- Redacts any JWT-shaped tokens (eyJ...) from breadcrumb messages and exception text
- Consider using `options.ServerName = null` to prevent hostname leakage

---

#### C2: TelemetryEvent.Properties is an unsanitized Dictionary -- PII sink [CRITICAL]
**File:** `backend/src/Taskdeck.Application/Services/TelemetryEvent.cs` (line 42)
**File:** `backend/src/Taskdeck.Application/Services/TelemetryEventService.cs` -- ValidateEvent does NOT inspect Properties

The Properties dictionary accepts arbitrary key-value pairs from the frontend with zero validation. A frontend bug or a malicious caller could send:
```json
{"properties": {"card_title": "private medical info", "email": "user@corp.com"}}
```

The backend logs these at LogInformation level (line 41-45 of TelemetryEventService.cs) without inspecting or sanitizing them. The doc comment says "Keys and values must not contain PII" but this is an unenforced honor-system constraint.

**Recommendation:**
- Validate property keys against an allowlist of known safe keys
- Cap the number of properties (e.g., max 10)
- Cap individual value size (e.g., max 200 chars)
- At minimum, refuse to log Properties until validation exists

---

### HIGH Findings

#### H1: Analytics ScriptUrl is injected into DOM without URL validation -- XSS via config [HIGH]
**File:** `frontend/taskdeck-web/src/composables/useAnalyticsScript.ts` (line 25)

```typescript
script.src = config.scriptUrl
```

The scriptUrl comes from the backend config endpoint, which returns it directly from appsettings.json. If an operator misconfigures this (or if an attacker gains config write access), any arbitrary script URL -- including javascript: URIs or attacker-controlled domains -- gets injected into the DOM.

There is no validation that scriptUrl is:
- An HTTPS URL
- On an allowlisted domain
- Not a javascript:, data:, or blob: URI

**Recommendation:** Validate scriptUrl on the backend with a URL format check (must start with https://). On the frontend, add a guard before injection: `if (!config.scriptUrl.startsWith('https://')) return`.

---

#### H2: Auth test is broken -- creates unauthClient but uses _client [HIGH]
**File:** `backend/tests/Taskdeck.Api.Tests/TelemetryApiTests.cs` (lines 47-61)

```csharp
var unauthClient = new HttpClient { BaseAddress = _client.BaseAddress };
var response = await _client.PostAsJsonAsync(...)  // BUG: uses _client, not unauthClient
```

The test creates unauthClient (line 50) but never uses it. It then posts with _client (from the factory). This test passes only because _client happens to not have auth headers set at that point -- but the test does NOT prove what it claims. The unauthClient is also not configured with the test server's HttpMessageHandler, so even if it were used, it would try to make a real HTTP call, not hit the in-memory test server.

**Impact:** The test gives false confidence that auth is enforced. If someone changes the factory to auto-authenticate, this test would still pass while no longer testing the unauthenticated path.

**Recommendation:** Remove the unused unauthClient. The test currently works by accident because _client has no auth headers, but add a comment explaining this.

---

#### H3: No DNT (Do Not Track) / GPC (Global Privacy Control) respect [HIGH]
**File:** `frontend/taskdeck-web/src/store/telemetryStore.ts`, `frontend/taskdeck-web/src/composables/useAnalyticsScript.ts`

Neither the telemetry store nor the analytics script checks navigator.doNotTrack or navigator.globalPrivacyControl. While telemetry defaults to off and requires opt-in, the consent UI does not inform users about DNT/GPC, and the system will happily accept opt-in from a user whose browser signals DNT=1.

For a product that values privacy as a core principle, respecting these signals is both an ethical and potential legal requirement (GPC has legal force under CCPA).

**Recommendation:**
- Check navigator.globalPrivacyControl and navigator.doNotTrack in initialize()
- If GPC is set, do not auto-restore consent from localStorage
- Display a notice in the consent UI when DNT/GPC is detected

---

### MEDIUM Findings

#### M1: SentrySettings.SendDefaultPii property is misleading -- creates false sense of control [MEDIUM]
**Files:** `backend/src/Taskdeck.Application/Services/SentrySettings.cs` (line 33), `backend/src/Taskdeck.Api/appsettings.json` (line 46)

The config model exposes SendDefaultPii as a settable property and appsettings.json includes it. The docs say "Always forced to false in code" but having the property in config invites someone to set it to true expecting it to work. The code silently ignores it -- a classic pit of confusion.

**Recommendation:** Remove SendDefaultPii from SentrySettings and from appsettings.json. If it must exist for documentation purposes, add a startup warning if config sets it to true.

---

#### M2: No rate limiting on POST /api/telemetry/events [MEDIUM]
**File:** `backend/src/Taskdeck.Api/Controllers/TelemetryController.cs`

The events endpoint has auth but no rate limiting. An authenticated user could flood the server with telemetry batches (100 events each, unlimited frequency). This is a DoS vector that would fill application logs (since events are logged at Information level).

**Recommendation:** Apply the existing rate limiting infrastructure to the telemetry endpoint, or add a dedicated per-user throttle.

---

#### M3: Sentry breadcrumb scrubbing is case-incomplete and misses other sensitive headers [MEDIUM]
**File:** `backend/src/Taskdeck.Api/Extensions/SentryRegistration.cs` (lines 44-45)

```csharp
var sensitiveKeys = new[] { "Authorization", "authorization", "Cookie", "cookie" };
```

HTTP headers are case-insensitive (RFC 7230). This only handles two casings. Headers like AUTHORIZATION, COOKIE, Set-Cookie, or X-Api-Key would pass through. The comparison should be case-insensitive.

**Recommendation:** Use case-insensitive comparison for header key matching.

---

#### M4: Frontend telemetry emit() does not validate event names [MEDIUM]
**File:** `frontend/taskdeck-web/src/store/telemetryStore.ts` (lines 114-138)

The backend validates event names against noun.verb regex, but the frontend emit() accepts any string. Invalid events are buffered, sent to the server, and rejected -- wasting bandwidth and buffer space. Frontend validation would catch typos early and avoid unnecessary network traffic.

**Recommendation:** Add the same noun.verb regex validation in emit() before buffering.

---

#### M5: dispose() calls flush() as fire-and-forget -- events silently lost on page unload [MEDIUM]
**File:** `frontend/taskdeck-web/src/store/telemetryStore.ts` (lines 188-194)

`void flush()` is fire-and-forget async. On page unload, the browser will likely kill the request before it completes. Using navigator.sendBeacon would provide reliable delivery for the final flush.

Additionally, dispose() is never actually called -- it is exported but there is no beforeunload listener or app-level teardown that invokes it.

**Recommendation:** Wire up `window.addEventListener('beforeunload', dispose)` in initialize(), and use navigator.sendBeacon for the final flush.

---

### LOW Findings

#### L1: Consent disclosure says "Error codes (no error messages or stack traces)" but Sentry sends full stack traces [LOW]
**File:** `frontend/taskdeck-web/src/views/ProfileSettingsView.vue` (line 169)

The "What data is collected?" section tells users "Error codes (no error messages or stack traces)" -- but when Sentry is enabled, full exception stack traces ARE sent to Sentry. The consent toggle controls both product telemetry events AND Sentry, but the disclosure only describes the telemetry events.

**Recommendation:** Either separate Sentry consent from telemetry consent, or update the disclosure to mention that error tracking (when enabled by the server) includes stack traces.

---

#### L2: TelemetryEventService is Singleton with no thread safety concerns documented [LOW]
**File:** `backend/src/Taskdeck.Api/Extensions/SettingsRegistration.cs` (line 71)

The service is registered as Singleton and is stateless (just logs), so this is technically fine. But if a future iteration adds event persistence (as suggested in the doc comment), the singleton lifetime will create concurrency issues.

**Recommendation:** Consider Scoped lifetime, or add a comment warning future contributors about the singleton constraint.

---

#### L3: Config endpoint leaks environment and tracesSampleRate even when Sentry is disabled [LOW]
**File:** `backend/src/Taskdeck.Api/Controllers/TelemetryController.cs` (lines 48-51)

When Sentry is disabled, Dsn is correctly returned as empty, but Environment (e.g., "production") and TracesSampleRate are still returned. This leaks infrastructure information to unauthenticated callers.

**Recommendation:** Return empty/default values for all Sentry fields when disabled, not just the DSN.

---

### INFO

**I1: No ADR for telemetry architecture decisions.** This PR introduces a cross-cutting concern (telemetry) with several design decisions (dual-gate model, event taxonomy, consent persistence strategy). Per CLAUDE.md, decisions with cross-cutting impact should have ADRs.

**I2: Good: sessionId is per-app-load, not persistent.** The session ID is regenerated on each app load and stored only in memory (not localStorage), preventing cross-session tracking.

**I3: Good: dual-gate design is sound.** Requiring both server config AND user consent is the right architecture.

---

### Summary
- **2 CRITICAL**: Exception message PII leaks to Sentry; Properties dictionary accepts arbitrary PII
- **3 HIGH**: XSS via unvalidated ScriptUrl; broken auth test; no DNT/GPC respect
- **5 MEDIUM**: Misleading SendDefaultPii config; no rate limiting; incomplete header scrubbing; no frontend event validation; lost events on dispose
- **3 LOW**: Inaccurate consent disclosure; singleton lifetime; config info leak
- **3 INFO**: Missing ADR; good session ID design; good dual-gate design

The CRITICAL findings are PR-blocking. The Properties dictionary being an open PII sink and the lack of Sentry BeforeSend scrubbing must be addressed before merge.

PR #812, opened 2026-04-09
CLD-03 follow-up: DB-backed auth codes, PKCE, account linking

### LINKED ISSUE #676: CLD-03 follow-up: Distributed auth code store, PKCE, and account linking
## Context
PR #668 (GitHub OAuth).

## Issues
1. **In-memory auth code store** (\`ConcurrentDictionary\`) doesn't survive restarts and won't work in multi-instance deployments. Needs Redis or DB-backed storage.
2. **No PKCE** enabled. GitHub supports it; ASP.NET Core 8 has \`UsePkce = true\`. Defense-in-depth against authorization code interception.
3. **No account linking**. Existing password-based users cannot link their GitHub account. Would need a "Link GitHub" button in settings.

## Parent
#537

Chris0Jeky:
## Summary

Implements #676 — three concrete improvements to the GitHub OAuth integration:

- **In-memory auth code store replaced with SQLite/EF Core**: Auth codes now persist across restarts and work in multi-instance deployments. Uses `OAuthAuthCode` entity with atomic `TryConsumeAtomicAsync` for race-safe single-use semantics via raw SQL `UPDATE WHERE IsConsumed = 0`.

- **PKCE enabled**: Sets `UsePkce = true` on the GitHub OAuth handler. ASP.NET Core 8 automatically generates code_verifier/code_challenge — defense-in-depth against authorization code interception.

- **Account linking**: Authenticated users can link/unlink GitHub accounts from the Settings page. Uses a dedicated link flow: `GET /api/auth/github/login?mode=link` stores GitHub identity in a link code, then `POST /api/auth/github/link` exchanges the code while requiring JWT auth. Prevents linking already-linked accounts with proper conflict detection.

### Key files changed

**Backend:**
- `backend/src/Taskdeck.Domain/Entities/OAuthAuthCode.cs` — New entity with Purpose (login/link), TryConsume, CreateForLinking
- `backend/src/Taskdeck.Api/Controllers/AuthController.cs` — DB-backed exchange, link/unlink/linked-accounts endpoints
- `backend/src/Taskdeck.Api/Extensions/AuthenticationRegistration.cs` — PKCE enabled
- `backend/src/Taskdeck.Application/Services/AuthenticationService.cs` — CompleteAccountLinkAsync, UnlinkExternalLoginAsync

**Frontend:**
- `frontend/taskdeck-web/src/views/ProfileSettingsView.vue` — Linked Accounts section with Link/Unlink GitHub
- `frontend/taskdeck-web/src/api/authApi.ts` — linkGitHub, unlinkGitHub, getLinkedAccounts
- `frontend/taskdeck-web/src/types/auth.ts` — LinkedAccount type

**Tests:**
- 15 domain tests for OAuthAuthCode entity
- 9 application tests for account linking/unlinking
- Updated OAuthTokenLifecycleTests for DB-backed store (concurrent exchange, cleanup, replay)
- Updated AuthControllerEdgeCaseTests for atomic consume
- Updated ProfileSettingsView test for router/auth mocks

## Test plan

- [x] Backend: `dotnet test backend/Taskdeck.sln -c Release -m:1` — all 1181 tests pass
- [x] Frontend: `npm run typecheck && npx vitest --run` — all 1898 tests pass, no errors
- [ ] Manual: Verify GitHub OAuth login still works end-to-end
- [ ] Manual: Test account linking from Settings page
- [ ] Manual: Verify PKCE parameters in GitHub authorization URL

Closes #676

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/AuthController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/AuthController.cs
@@ -1,5 +1,5 @@
-using System.Collections.Concurrent;
 using System.Security.Cryptography;
+using System.Text.Json;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
@@ -11,6 +11,7 @@ using Taskdeck.Api.RateLimiting;
 using Taskdeck.Application.DTOs;
 using Taskdeck.Application.Interfaces;
 using Taskdeck.Application.Services;
+using Taskdeck.Domain.Entities;
 using Taskdeck.Domain.Exceptions;
 using AuthenticationService = Taskdeck.Application.Services.AuthenticationService;
 
@@ -18,6 +19,7 @@ namespace Taskdeck.Api.Controllers;
 
 public record ChangePasswordRequest(string CurrentPassword, string NewPassword);
 public record ExchangeCodeRequest(string Code);
+public record LinkExchangeRequest(string Code);
 
 /// <summary>
 /// Authentication endpoints — register, login, change password, and GitHub OAuth flow.
@@ -30,16 +32,14 @@ public class AuthController : AuthenticatedControllerBase
 {
     private readonly AuthenticationService _authService;
     private readonly GitHubOAuthSettings _gitHubOAuthSettings;
+    private readonly IUnitOfWork _unitOfWork;
 
-    // Short-lived, single-use authorization codes to avoid exposing JWT in URLs.
-    // Key: code, Value: (token, expiry). Codes expire after 60 seconds.
-    private static readonly ConcurrentDictionary<string, (AuthResultDto Result, DateTimeOffset Expiry)> _authCodes = new();
-
-    public AuthController(AuthenticationService authService, GitHubOAuthSettings gitHubOAuthSettings, IUserContext userContext)
+    public AuthController(AuthenticationService authService, GitHubOAuthSettings gitHubOAuthSettings, IUserContext userContext, IUnitOfWork unitOfWork)
         : base(userContext)
     {
         _authService = authService;
         _gitHubOAuthSettings = gitHubOAuthSettings;
+        _unitOfWork = unitOfWork;
     }
 
     /// <summary>
@@ -117,10 +117,11 @@ public class AuthController : AuthenticatedControllerBase
 
     /// <summary>
     /// Initiates GitHub OAuth login flow. Only available when GitHub OAuth is configured.
+    /// Pass mode=link to start an account-linking flow instead of login.
     /// </summary>
     [HttpGet("github/login")]
     [EnableRateLimiting(RateLimitingPolicyNames.AuthPerIp)]
-    public IActionResult GitHubLogin([FromQuery] string? returnUrl = null)
+    public IActionResult GitHubLogin([FromQuery] string? returnUrl = null, [FromQuery] string? mode = null)
     {
         if (!_gitHubOAuthSettings.IsConfigured)
             return NotFound(new ApiErrorResponse(ErrorCodes.NotFound, "GitHub OAuth is not configured"));
@@ -131,9 +132,15 @@ public class AuthController : AuthenticatedControllerBase
 
         var properties = new Microsoft.AspNetCore.Authentication.AuthenticationProperties
         {
-            RedirectUri = Url.Action(nameof(GitHubCallback), new { returnUrl }),
+            RedirectUri = Url.Action(nameof(GitHubCallback), new { returnUrl, mode }),
             Items = { { "LoginProvider", "GitHub" } }
         };
+
+        // Store mode in the auth properties so the callback can detect linking
+        if (mode == "link")
+        {
+            properties.Items["mode"] = "link";
+        }
 
         return Challenge(properties, "GitHub");
     }
@@ -143,7 +150,7 @@ public class AuthController : AuthenticatedControllerBase
     /// </summary>
     [HttpGet("github/callback")]
     [EnableRateLimiting(RateLimitingPolicyNames.AuthPerIp)]
-    public async Task<IActionResult> GitHubCallback([FromQuery] string? returnUrl = null)
+    public async Task<IActionResult> GitHubCallback([FromQuery] string? returnUrl = null, [FromQuery] string? mode = null)
     {
         if (!_gitHubOAuthSettings.IsConfigured)
             return NotFound(new ApiErrorResponse(ErrorCodes.NotFound, "GitHub OAuth is not configured"));
@@ -171,6 +178,38 @@ public class AuthController : AuthenticatedControllerBase
                 "GitHub did not return a user identifier"));
         }
 
+        // Sign out the temporary cookie used during the OAuth handshake
+        await HttpContext.SignOutAsync("GitHub");
+
+        // Account linking flow: store the GitHub identity as a link code
+        if (mode == "link")
+        {
+            var linkCode = GenerateAuthCode();
+            var providerData = JsonSerializer.Serialize(new
+            {
+                provider = "GitHub",
+                providerUserId,
+                displayName,
+                avatarUrl
+            });
+
+            var linkAuthCode = OAuthAuthCode.CreateForLinking(
+                code: linkCode,
+                providerData: providerData,
+                expiresAt: DateTimeOffset.UtcNow.AddSeconds(60));
+
+            await _unitOfWork.OAuthAuthCodes.AddAsync(linkAuthCode);
+            await _unitOfWork.SaveChangesAsync();
+
+            var linkReturnUrl = !string.IsNullOrWhiteSpace(returnUrl) && Url.IsLocalUrl(returnUrl)
+                ? returnUrl
+                : "/";
+
+            var linkSeparator = linkReturnUrl.Contains('?') ? "&" : "?";
+            return Redirect($"{linkReturnUrl}{linkSeparator}oauth_link_code={Uri.EscapeDataString(linkCode)}");
+        }
+
+        // Normal login flow
         // GitHub may not return an email if user's email is private
         if (string.IsNullOrWhiteSpace(email))
             email = $"{providerUserId}@users.noreply.github.com";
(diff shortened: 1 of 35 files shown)


Chris0Jeky:
## Adversarial Self-Review

### Security Findings

**[FIXED] MEDIUM: Link mode from query string instead of OAuth state**
The `mode=link` parameter was read from the callback query string, which could be tampered with. Fixed to prefer reading from `authenticateResult.Properties.Items` (cryptographically protected OAuth state).

**[FIXED] MEDIUM: Fire-and-forget cleanup could use disposed scope**
`_ = CleanupExpiredCodesAsync()` ran as fire-and-forget, but used the scoped `_unitOfWork` which could be disposed after the request completes. Fixed to `await` the cleanup within the request scope.

**[FIXED] LOW: Dead code (LinkExternalLoginAsync)**
`LinkExternalLoginAsync` was in the interface but never called from any controller. Removed.

**LOW: Auth code replay - MITIGATED**
Auth code replay is prevented by `TryConsumeAtomicAsync` which uses `UPDATE WHERE IsConsumed = 0` for atomic single-use semantics. Concurrent exchange tests verify only one of 5 concurrent requests succeeds.

**LOW: PKCE downgrade - NOT APPLICABLE**
PKCE is enforced server-side via `UsePkce = true`. The ASP.NET Core OAuth handler always includes code_challenge in the authorization request. GitHub validates it during token exchange. There is no client-side code that could skip PKCE.

**LOW: Account linking CSRF - MITIGATED**
The link endpoint `POST /api/auth/github/link` requires `[Authorize]` (valid JWT). The link code is single-use with 60-second TTL. An attacker would need both a valid JWT and a fresh link code to exploit this. The link flow also verifies the GitHub account isn't already linked to another user (409 Conflict).

**LOW: Timing attacks on code validation - ACCEPTABLE**
Code lookup uses `GetByCodeAsync` which is a standard database query on an indexed column. While not constant-time, the code itself is 32 bytes of cryptographic randomness (256 bits of entropy), making brute-force infeasible within the 60-second TTL.

### Migration Safety
The migration adds a new `OAuthAuthCodes` table with no impact on existing tables. The old in-memory `ConcurrentDictionary` is removed, so any active auth codes at the time of deployment will be lost (they had a 60-second TTL, so this is acceptable).

### Test Coverage
- 15 domain tests for OAuthAuthCode entity validation and consumption
- 9 application tests for account linking/unlinking edge cases
- Integration tests: concurrent exchange (5 parallel), replay prevention, expired codes, cleanup
- Frontend: 1898 tests all passing, ProfileSettingsView test updated for new dependencies

Chris0Jeky:
## Independent Adversarial Review (Round 2) -- Security-Focused

### CRITICAL Findings

#### 1. [CRITICAL] Account Linking CSRF: Attacker can link their GitHub account to victim's session

The account-linking flow has a CSRF vulnerability. The attack:

1. Attacker visits `GET /api/auth/github/login?mode=link`, authenticates with **their own** GitHub account.
2. The callback creates a link code and redirects to `?oauth_link_code=ATTACKER_CODE`.
3. Attacker captures the link code from the redirect URL (intercepts the redirect before completing).
4. Attacker tricks the victim (who is logged in) into visiting `/workspace/settings?oauth_link_code=ATTACKER_CODE` (e.g., via a crafted link, image tag, or XSS).
5. The victim's browser automatically exchanges the code via `POST /api/auth/github/link` with the victim's JWT.
6. The attacker's GitHub account is now linked to the victim's account.
7. Attacker can now log in as the victim via GitHub OAuth.

**Root cause**: The link code is not bound to any particular user session. It is a bearer token that can be exchanged by *anyone* who holds a valid JWT. The `mode=link` endpoint at `GET /api/auth/github/login` does not require authentication (no `[Authorize]` attribute), so anyone can initiate a link flow and generate a code.

**Fix**: The link code must be bound to the user who initiated the link flow. Either:
- Require `[Authorize]` on the `github/login?mode=link` endpoint and store the caller's user ID in the `OAuthAuthCode.UserId` field, then verify it matches at exchange time.
- Or use a CSRF token / state parameter tied to the authenticated session.

**File**: `AuthController.cs` lines 122-146 (GitHubLogin) and 322-367 (LinkGitHub)

#### 2. [CRITICAL] TryConsumeAtomicAsync does NOT check expiry -- expired codes can be consumed

The `TryConsumeAtomicAsync` SQL statement:
```sql
UPDATE OAuthAuthCodes SET IsConsumed = 1, ConsumedAt = {0}, UpdatedAt = {1} WHERE Code = {2} AND IsConsumed = 0
```

This query does **not** filter on `ExpiresAt`. While the controller checks `authCode.IsExpired` before calling `TryConsumeAtomicAsync`, there is a **TOCTOU (Time-of-Check-Time-of-Use) race condition**:

1. Request A reads the auth code at T=59s (not expired, TTL is 60s).
2. Request A checks `IsExpired` -- returns false.
3. Clock ticks past 60s.
4. Request A calls `TryConsumeAtomicAsync` -- succeeds because the SQL only checks `IsConsumed = 0`, not expiry.
5. The code was consumed **after it expired**.

In practice the window is narrow (under 1 second), but for auth-critical code the atomic operation should enforce ALL invariants:
```sql
WHERE Code = {2} AND IsConsumed = 0 AND ExpiresAt > {3}
```

**File**: `OAuthAuthCodeRepository.cs` line 26

#### 3. [CRITICAL] JWT stored in plaintext in database -- token theft via DB access

The `OAuthAuthCode.Token` field stores the **fully formed JWT token** in plaintext in the SQLite database. If an attacker gains read access to the database (SQL injection elsewhere, backup exposure, filesystem access), they can extract valid JWT tokens from unconsumed auth codes. The 60-second TTL limits the window, but consumed codes (with `IsConsumed = 1`) also retain the token indefinitely until cleanup runs.

This is worse than the previous in-memory store because:
- SQLite files persist on disk (survives restarts by design).
- The `DeleteExpiredAsync` cleanup only removes expired codes, not consumed codes, so consumed-but-not-expired codes with JWTs linger.
- No encryption at rest.

**Mitigation**: Either (a) clear the Token field upon consumption, or (b) do not store the JWT at all -- store only the userId and re-issue a fresh JWT at exchange time, or (c) hash/encrypt the token.

**File**: `OAuthAuthCode.cs` line 38, `AuthController.cs` lines 244-248, 308

---

### HIGH Findings

#### 4. [HIGH] DeleteExpiredAsync loads ALL auth codes into memory -- DoS vector

```csharp
var allCodes = await _context.Set<OAuthAuthCode>().ToListAsync(cancellationToken);
```

This loads every `OAuthAuthCode` row into memory. If an attacker can generate auth codes at scale (the callback endpoint is rate-limited per IP, but rate limits can be bypassed with distributed IPs or if the rate limiter is misconfigured), they could accumulate thousands of rows and cause OOM when cleanup runs.

Even without attack, this is a correctness bug -- the comment says "Auth codes are short-lived and few in number" but this is only true if cleanup actually runs regularly, and it only runs during successful OAuth callbacks (best-effort, swallowed exceptions).

**Fix**: Use raw SQL for cleanup like `TryConsumeAtomicAsync` does:
```sql
DELETE FROM OAuthAuthCodes WHERE ExpiresAt < {0}
```

**File**: `OAuthAuthCodeRepository.cs` lines 34-50

#### 5. [HIGH] Link code is not bound to authenticated user -- no session binding

Related to finding #1, but distinct: the `LinkGitHub` endpoint accepts a link code from *any* authenticated user, not just the user who initiated the link flow. The link code stores `ProviderData` but no target user identity. This means:

- User A initiates a link, gets code X.
- User B (also authenticated) could POST code X to `/api/auth/github/link` and link User A's GitHub to User B's account.

The code should store the intended user ID (from the JWT at initiation time) and verify it at consumption time.

**File**: `AuthController.cs` lines 192-217 (link code creation has no userId), lines 322-367 (consumption does not verify initiator)

#### 6. [HIGH] No cleanup of consumed auth codes -- unbounded table growth

`DeleteExpiredAsync` only deletes codes where `ExpiresAt < cutoff`. Consumed-but-not-yet-expired codes are retained. Over time, the table grows unboundedly since:
- Each OAuth callback creates a new row.
- Cleanup only removes expired rows.
- Consumed rows with future expiry are never cleaned up.
- There is no background job; cleanup only runs opportunistically during callbacks.

For a production system, this table will grow indefinitely if OAuth is used frequently.

**Fix**: Also delete consumed codes in cleanup (`WHERE IsConsumed = 1 OR ExpiresAt < cutoff`), or add a background cleanup job.

**File**: `OAuthAuthCodeRepository.cs` line 42, `AuthController.cs` line 432

---

### MEDIUM Findings

#### 7. [MEDIUM] ExchangeCode exposes timing side-channel for code existence

The exchange endpoint returns different error messages depending on whether the code exists, is expired, is consumed, or is for the wrong purpose:
- "Invalid or expired code" (not found)
- "This code is for account linking, not login" (wrong purpose)
- "Code has expired" (expired)
- "Invalid or expired code" (already consumed)

An attacker can use these distinct responses to enumerate valid codes and determine their state. All failure paths should return the same generic error message.

**File**: `AuthController.cs` lines 276-292

#### 8. [MEDIUM] mode query parameter on callback is attacker-controllable

The callback reads `mode` from both the OAuth state (tamper-proof) and the query string:
```csharp
var isLinkMode = mode == "link";
if (authenticateResult.Properties?.Items.TryGetValue("mode", out var stateMode) == true)
{
    isLinkMode = stateMode == "link";
}
```

While the state parameter takes precedence, the fallback to `mode == "link"` from the query string means an attacker could append `?mode=link` to a callback URL for a login flow, potentially causing the callback to create a link code instead of completing login. The query string should be ignored entirely; only the OAuth state should determine the mode.

**File**: `AuthController.cs` lines 186-190

#### 9. [MEDIUM] Frontend XSS risk via displayName in success message

```typescript
linkSuccess.value = `GitHub account linked successfully (${linked.displayName || linked.providerUserId})`
```

While Vue's `{{ }}` template binding auto-escapes HTML, the `displayName` is user-controlled (from GitHub profile). If this string is ever rendered via `v-html` or used in a context where it is not escaped, it could lead to XSS. Currently safe due to Vue's default escaping, but fragile.

**File**: `ProfileSettingsView.vue` line 103

#### 10. [MEDIUM] avatarUrl rendered as img src without validation

```html
<img v-if="gitHubAccount.avatarUrl" :src="gitHubAccount.avatarUrl" ... />
```

The `avatarUrl` comes from GitHub via the link code and is stored unvalidated. A malicious OAuth provider response could inject a `javascript:` URL or a tracking pixel URL. Should validate that `avatarUrl` starts with `https://` and belongs to a known GitHub domain.

**File**: `ProfileSettingsView.vue` lines 231-236

---

### LOW Findings

#### 11. [LOW] PKCE implementation relies entirely on framework -- no verification

The PKCE setup is just `options.UsePkce = true`. While ASP.NET Core 8 does handle PKCE automatically, there is no test verifying that the authorization URL actually contains `code_challenge` and `code_challenge_method` parameters. If a future framework upgrade or config change silently disables PKCE, there would be no test failure.

**File**: `AuthenticationRegistration.cs` line 104

#### 12. [LOW] ExecuteSqlRawAsync parameter passing format

While `ExecuteSqlRawAsync` with `{N}` placeholders does parameterize queries (preventing SQL injection), the method name `Raw` is misleading and future developers might confuse it with string interpolation. Consider using `ExecuteSqlInterpolatedAsync` which makes parameterization more explicit and reduces risk of future mistakes.

**File**: `OAuthAuthCodeRepository.cs` lines 25-29

#### 13. [LOW] Missing CancellationToken forwarding in TryConsumeAtomicAsync

The `cancellationToken` parameter is accepted but never passed to `ExecuteSqlRawAsync`. If the HTTP request is cancelled, the SQL update will still complete.

**File**: `OAuthAuthCodeRepository.cs` line 20-31

---

### INFO

#### 14. [INFO] No ADR for security-impacting architecture change

Moving from in-memory to DB-backed auth codes, adding account linking, and enabling PKCE are security-posture changes. Per CLAUDE.md: "Do not skip ADRs for decisions that affect architecture, security posture, or cross-cutting conventions." An ADR should document this decision.

#### 15. [INFO] Migration changes ProductVersion from 8.0.14 to 8.0.25

The model snapshot bump from `8.0.14` to `8.0.25` suggests EF Core tooling was upgraded. This is fine but should be noted for reviewers.

---

### Summary

| Severity | Count | Key Issues |
|----------|-------|------------|
| CRITICAL | 3 | Account linking CSRF, TOCTOU on expiry check, JWT plaintext in DB |
| HIGH | 3 | DoS via full-table load, no session binding on link codes, unbounded table growth |
| MEDIUM | 4 | Timing side-channel, mode parameter injection, XSS risk, unvalidated avatar URL |
| LOW | 3 | No PKCE verification test, raw SQL naming, missing cancellation token |
| INFO | 2 | Missing ADR, EF version bump |

The most urgent issues are #1 (CSRF account linking), #2 (TOCTOU on expiry), and #3 (JWT plaintext storage). Issues #1 and #5 together mean an attacker can link their GitHub account to any victim who visits a crafted URL, then log in as the victim.

Issue #509, opened 2026-03-29
BUG: Board auto-switches between boards every few seconds without user input


Chris0Jeky:
## Summary
After creating a second board, the active board view switches between boards every few seconds unprompted, making the board surface nearly unusable.

## Reproduction
1. Log in with an existing account (has onboarding board)
2. Create a second board (e.g., calendar from default templates)
3. Navigate to the calendar board
4. Wait ~3-5 seconds with no input

## Expected
Board view stays on the user-selected board until the user navigates away.

## Actual
View auto-switches: calendar → onboarding (a few seconds) → calendar (1 second) → onboarding (persists). Happens continuously.

## Root Cause Hypothesis
A polling/subscription side-effect is triggering a board-selection store mutation. When `fetchBoards` returns results in a different order, the "active board" selector likely resets to the first item rather than preserving the user's selection. Alternatively, a SignalR reconnect event is resetting board context.

## Fix
In `boardStore`: add a `preserveSelection` guard so that `setBoards`/`fetchBoards` does NOT overwrite `activeBoardId` if the currently-selected board still exists in the refreshed list.

## Priority
**P0 — core loop broken**

_Discovered in manual testing session 2026-03-29. See `docs/analysis/2026-03-29_manual_testing_consolidated_findings.md` (BUG-C2)._

Issue #508, opened 2026-03-29
BUG: Fresh user sees other users' automation queue data (data isolation failure)


Chris0Jeky:
## Summary
A fresh-registered user sees all automation queue items from other users' sessions, including items created days before their registration. This is a P0 data isolation failure.

## Reproduction
1. Register a brand-new account
2. Navigate to `/workspace/automation/queue`
3. Observe items from previous dates (e.g., 2026-03-27) with Completed/Failed statuses

## Expected
New user sees an empty queue (0 items).

## Actual
9 total items visible (6 Completed, 3 Failed) from 27/03/2026, belonging to other sessions. Internal error messages exposed in failure details (e.g., `BoardId is required to triage capture items into proposals`).

## Root Cause Hypothesis
The LLM queue list endpoint likely fetches all records without filtering by the authenticated user's `UserId`. A claims-derived `userId` predicate is missing from the list query in `LlmQueueController` (or its backing repository).

## Fix
Add `WHERE UserId = @currentUserId` (claims-derived) to the queue list query. Add a cross-user isolation integration test to `Api.Tests`.

## Priority
**P0 — must fix before any external user onboarding**

_Discovered in manual testing session 2026-03-29. See `docs/analysis/2026-03-29_manual_testing_consolidated_findings.md` (BUG-C1)._

PR #816, opened 2026-04-10
TST-44: Frontend store integration tests

### LINKED ISSUE #711: TST-44: Frontend store integration tests — stores + real API module interaction
## Context

Frontend stores are tested with mocked API modules, and API modules are tested with mocked HTTP. But the boundary between stores and API modules — especially error mapping, response transformation, and state reconciliation — is not tested as a unit. Additionally, several stores lack tests entirely.

### Stores With No Tests
- **captureStore** — the core capture workflow store; manages capture creation, queue status, triage state
- **workspaceStore** — workspace mode management, active board selection, mode persistence
- **archiveStore** — archive listing, restore operations, restore status tracking
- **chatStore** (if it exists separately from chatApi) — message accumulation, session management, tool-call streaming

#### captureStore
1. Create capture → API succeeds → store reflects new capture in list
2. Create capture → API returns 400 (validation error) → store shows error, no phantom capture
3. Create capture → API returns 500 → store shows network error, capture not lost
4. Fetch capture list → includes proper pagination/sorting
5. Capture triage status polling → store reflects Processing → Completed transition

#### boardStore
6. Create board → optimistic update → API confirms → final state matches
7. Create board → API fails → optimistic update rolled back
8. Card drag-drop → optimistic move → API confirms → position stable after re-render
9. Card drag-drop → API 409 Conflict → card snaps back to original position
10. Board data refresh while user is editing a card → edit not lost
11. Column reorder → optimistic → API confirms → all cards maintain correct column association
12. Delete board → confirmation → API succeeds → board removed from list, redirect to another board
13. **Board auto-switching bug regression (#509)**: creating a second board does not auto-switch away from current board

#### queueStore
14. Queue polling → items transition through Pending → Processing → Completed states
15. Queue item disappears (deleted server-side) → store handles gracefully (no phantom entry)
16. **Data isolation regression (#508)**: queue data only shows authenticated user's items

#### notificationStore
17. Real-time notification arrival (via SignalR or polling) → store increments unread count
18. Mark all read → all notifications show as read, unread count drops to 0
19. Board-scoped notification filtering → only shows notifications for selected board

#### sessionStore
20. Token expiry → store detects 401 → redirects to login
21. OAuth code exchange → store receives token → stores in local storage → redirects to intended page
22. Logout → store clears token, clears all other stores' user-specific state

## Implementation Notes

- Use Vitest with Pinia test utils (\`setActivePinia(createPinia())\`)
- For store + API integration: don't mock the API module, mock the HTTP layer (msw or vi.fn on fetch)
- This allows testing the actual store → API → HTTP chain, catching mismatches in response shapes
- For optimistic update tests: verify intermediate state (before API response) and final state (after response)

## Edge Cases to Stress

- **Stale state reconciliation**: Store has cached data, API returns newer data → store updates correctly
- **Concurrent mutations**: Two users editing the same board → SignalR event triggers store refresh
- **Network offline then online**: API calls fail with network error → retry when online → state recovers
- **Large datasets**: Board with 500 cards → store pagination/virtualization doesn't lose cards
- **Error message mapping**: API returns \`errorCode\` + \`message\` → store maps to user-friendly error string

Chris0Jeky:
## Summary
- Extends captureStore integration tests with API 500 error handling, multi-param query forwarding, cancelItem action, and triage polling lifecycle coverage
- Adds boardStore integration tests for card snap-back on 409 conflict, deleteCard, and column CRUD (create/delete)
- Adds notificationStore integration tests for unread count derivation, markAllRead count behavior, single-read decrement, real-time notification injection, and board-scoped unread filtering
- Adds sessionStore integration tests for changePassword action and token expiry detection via isAuthenticated computed
- Adds queueStore integration tests for loading state transitions and cancel-item isolation
- Creates archiveApi integration tests covering listing with filters, restore operations, URL encoding for special characters, error propagation, and conflict handling
- Covers regression scenarios for #508 (data isolation) and #509 (board auto-switching)

Closes #711

## Test plan
- [x] All new store integration tests pass with vitest (1952 tests, 162 files, 0 failures)
- [x] npm run typecheck passes
- [x] No existing tests broken

### DIFF
--- a/frontend/taskdeck-web/src/tests/store/archiveApi.integration.spec.ts
+++ b/frontend/taskdeck-web/src/tests/store/archiveApi.integration.spec.ts
@@ -64,8 +64,7 @@ describe('archiveApi — integration (mocked HTTP)', () => {
 
       await archiveApi.getItems({ entityType: 'card' })
 
-      const calledUrl = vi.mocked(http.get).mock.calls[0][0] as string
-      expect(calledUrl).toContain('entityType=card')
+      expect(http.get).toHaveBeenCalledWith(expect.stringContaining('entityType=card'))
     })
 
     it('appends boardId filter to the query string', async () => {
@@ -73,8 +72,7 @@ describe('archiveApi — integration (mocked HTTP)', () => {
 
       await archiveApi.getItems({ boardId: 'board-xyz' })
 
-      const calledUrl = vi.mocked(http.get).mock.calls[0][0] as string
-      expect(calledUrl).toContain('boardId=board-xyz')
+      expect(http.get).toHaveBeenCalledWith(expect.stringContaining('boardId=board-xyz'))
     })
 
     it('appends status filter to the query string', async () => {
@@ -82,8 +80,7 @@ describe('archiveApi — integration (mocked HTTP)', () => {
 
       await archiveApi.getItems({ status: 'Available' })
 
-      const calledUrl = vi.mocked(http.get).mock.calls[0][0] as string
-      expect(calledUrl).toContain('status=Available')
+      expect(http.get).toHaveBeenCalledWith(expect.stringContaining('status=Available'))
     })
 
     it('combines multiple filters in the query string', async () => {
@@ -91,10 +88,9 @@ describe('archiveApi — integration (mocked HTTP)', () => {
 
       await archiveApi.getItems({ entityType: 'card', boardId: 'board-1', limit: 50 })
 
-      const calledUrl = vi.mocked(http.get).mock.calls[0][0] as string
-      expect(calledUrl).toContain('entityType=card')
-      expect(calledUrl).toContain('boardId=board-1')
-      expect(calledUrl).toContain('limit=50')
+      expect(http.get).toHaveBeenCalledWith(expect.stringContaining('entityType=card'))
+      expect(http.get).toHaveBeenCalledWith(expect.stringContaining('boardId=board-1'))
+      expect(http.get).toHaveBeenCalledWith(expect.stringContaining('limit=50'))
     })
 
     it('propagates errors from the HTTP layer', async () => {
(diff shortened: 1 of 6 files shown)


Chris0Jeky:
## Self-review findings

Reviewed all 6 changed files against the adversarial checklist:

### Mock level
All tests mock at the HTTP layer (`vi.mock('../../api/http')`), never at the store or API module level. This is correct -- it tests the full store -> API module -> http chain and catches shape mismatches.

### State transition verification
Every test verifies both:
1. The HTTP call shape (URL, method, payload)
2. The resulting store state after the action completes

No test just calls an action without assertions.

### Error path coverage
Each store action tested has both success and failure paths:
- captureStore: 500 error, cancel failure, triage failure
- boardStore: 409 snap-back, delete failure
- notificationStore: fetch error, mark-read error
- sessionStore: changePassword failure, expired token detection
- queueStore: fetch error, loading state consistency

### Test isolation
- `beforeEach` creates fresh Pinia + clears mocks in every suite
- `afterEach` restores real timers where fake timers are used
- No shared mutable state between tests

### Boundary testing accuracy
- archiveApi tests live in `tests/store/` since there is no archiveStore; the archive functionality goes directly from ArchiveView -> archiveApi -> http
- Notification unread count is derived via `filter(!isRead)` since the store does not expose a computed property -- this correctly tests the underlying data mutation

**Verdict: No issues found. All tests are clean.**

Chris0Jeky (commented):
## Post-bot review and fixes

After reading all 6 test files and cross-referencing against the actual store/API implementations, here are the issues I found and am fixing:

### Bot comment fixes

1. **archiveApi: manual URL extraction pattern** (Gemini) — Lines 67-68, 76-77, 85-86, 94-97, and similar patterns in captureStore. Replaced `mock.calls[0][0]` extraction with idiomatic `expect(http.get).toHaveBeenCalledWith(expect.stringContaining(...))`.

2. **archiveApi restoreItem URL encoding test** — Same pattern for `http.post` mock call extraction on line 144-146.

3. **captureStore pollTriageCompletion: missing call count assertion** (Gemini) — Added assertion that `http.get` was called the expected number of times (3) to verify polling actually executed the right number of attempts.

4. **boardStore 409 snap-back test** (Copilot) — The test IS valid. After reading `cardStore.ts`, `moveCard` calls the API first and only updates local state after success. On 409, the card genuinely stays in its original column. Renamed the comment from "snap-back" to clarify there's no optimistic mutation to roll back — state was never changed.

5. **captureStore line 282 misleading comment** (Copilot) — The `getErrorDisplay` mock always returns the fallback, unconditionally. Fixed the comment to reflect reality.

### Additional issues found in adversarial review

6. **captureStore query parameter tests partly duplicate existing coverage** — The `fetchItems` describe block already tests status forwarding (line 97-105). The new query parameter tests add `boardId` and combined params, which are valuable, but noted overlap.

7. **notificationStore unread count tests test Array.filter, not a store getter** — These document expected behavior but the store doesn't expose an `unreadCount` computed. The "adding a new unread notification locally" test exercises pure array manipulation, not a store action. Added a comment clarifying the intent.

8. **captureStore fetchItems query parameter tests** — Applied the same `toHaveBeenCalledWith(expect.stringContaining(...))` idiom.

Issue #655, opened 2026-04-01
MCP-04: Phase 4 — MCP production hardening (deferred to v0.4.0+)


Chris0Jeky:
## Overview

Phase 4 of MCP server implementation. Production-grade hardening deferred until v0.4.0+ demand justifies the investment.

**Tracker**: #648 | **Spike**: `docs/spikes/SPIKE_619_COMPLETED.md` §13 Phase 4

## Scope

1. Structured logging and observability for MCP requests (OpenTelemetry integration)
2. Resource subscriptions (bridge SignalR for live board updates)
3. Prompt templates (if user demand — spike §8 recommends deferral)
4. OAuth 2.1 support (when cloud auth infrastructure exists from CLD wave)
5. API key management web UI (settings surface)
6. MCP spec version negotiation
7. Performance optimization (response caching for resources)
8. Scope-based API key permissions (per-board access restriction)

## Key Design Constraints

- Only invest in this when real external agent usage justifies the cost
- OAuth 2.1 depends on cloud auth infrastructure (GitHub OAuth from #539)
- Resource subscriptions require SignalR scale-out readiness (#105)

## Estimated Scope

2–4 weeks (deferred)

## Blocked By

- MCP-03 (#TBD — HTTP transport + auth)

Chris0Jeky:
**Progress update (2026-04-16):**

Two additional scope items from MCP-04 are now delivered:

1. **Structured logging and observability** (PR #879): `McpOperationLogger` for structured ILogger output, `McpTelemetryMiddleware` with OpenTelemetry spans/counters/histograms, `LogSanitizer` for sensitive data protection. 4 new test files.

2. **API key management UI** (PR #877): Frontend settings page at `/workspace/settings/api-keys` with full key lifecycle (list active/revoked, create with optional expiry, one-time key display + clipboard copy, revoke with confirmation). Uses Td* design system primitives. 19 tests.

**Delivered so far from #655 scope:**
- [x] Scope item 1: Structured logging/observability (#879)
- [x] Scope item 5: API key management UI (#877)

**Remaining:**
- [ ] Rate limiting dashboard/visibility
- [ ] Key rotation workflow
- [ ] Usage analytics per key

Chris0Jeky:
[...]

PR #879, opened 2026-04-16
feat: MCP structured logging and observability (#655)

Chris0Jeky:
## Summary
- Adds `McpOperationLogger` for structured ILogger output on MCP tool calls and resource access
- Adds `McpTelemetryMiddleware` with OpenTelemetry ActivitySource spans, request/error counters, and duration histograms
- Integrates with existing `TaskdeckTelemetry` ActivitySource and tag conventions
- Sensitive data protection: API keys and request bodies are never logged
- 4 new test files covering logging behavior and data protection

Sub-task of #655 (MCP-04 production hardening, scope item 1: structured logging/observability)

## Test plan
- [ ] `dotnet build backend/Taskdeck.sln -c Release` passes
- [ ] `dotnet test backend/Taskdeck.sln -c Release -m:1` passes
- [ ] New MCP logging tests pass
- [ ] Verify no PII/key leaks in log output


Chris0Jeky:
## Adversarial Self-Review

Full diff read. Findings below, rated by severity.

---

### 1. [MEDIUM] `McpApiKeyId` tag declared but never used

`TaskdeckTelemetryTags.McpApiKeyId` (`"taskdeck.mcp.api_key_id"`) is declared in the new tags but never set by any code in this PR. This is dead code. More importantly, if a future contributor sees this tag and starts populating it with the raw API key string instead of a key ID, it would be a **PII/credential leak**. Either remove the unused tag or add a doc comment clarifying it should only hold a hashed/opaque key identifier, never the raw value.

---

### 2. [MEDIUM] `exception.Message` passed to `Activity.SetStatus` may contain user data

In `McpOperationScope.Fail(Exception)`:
```csharp
_activity?.SetStatus(ActivityStatusCode.Error, exception.Message);
```
Exception messages from domain logic (e.g., "Board 'My Private Board' not found for user X") could contain user content or entity names that end up in your trace backend. Consider using a sanitized or type-only description (e.g., `exception.GetType().Name`) rather than the raw message, or explicitly document this as acceptable given the current exception patterns.

---

### 3. [MEDIUM] `McpOperationLogger` registered as Singleton three times

The diff adds `AddSingleton<McpOperationLogger>()` in three separate places in `Program.cs`:
1. Standalone MCP HTTP mode (`mcpHttpBuilder.Services`)
2. Stdio mode (`services`)
3. Co-hosted mode (`builder.Services`)

Each registration is in a different code path (only one runs per process), so there is no functional bug. However, the triple registration is a maintenance smell. If the constructor signature changes, three call sites must be updated. Consider extracting a shared `AddMcpTelemetry()` extension method.

---

### 4. [MEDIUM] Standalone MCP HTTP mode has no `CorrelationIdMiddleware`

The `McpTelemetryMiddleware` reads `CorrelationIdMiddleware.ItemKey` from `context.Items`, but in the standalone MCP HTTP mode (`mcpHttpApp`), `CorrelationIdMiddleware` is never registered in the pipeline. The code correctly falls back to `context.TraceIdentifier`, so there is no crash, but the correlation ID in logs will be a server-generated value rather than the client-supplied `X-Request-Id` header. This inconsistency between co-hosted and standalone modes could confuse operators debugging cross-service traces. Consider adding `CorrelationIdMiddleware` to the standalone pipeline too, or document the discrepancy.

---

### 5. [LOW] Metric naming does not follow OpenTelemetry semantic conventions

OTel semantic conventions use dots for namespace separation and recommend `{unit}` suffixes for duration metrics (e.g., `http.server.request.duration` with unit `s`). The current names:
- `taskdeck.mcp.requests` -- OK as a counter
- `taskdeck.mcp.request.duration` -- unit is `"ms"`, but OTel convention prefers seconds (`"s"`). Exporters that assume seconds (e.g., Prometheus `_seconds` suffix) will produce confusing output
- `taskdeck.mcp.errors` -- OK as a counter

This is a soft convention, not a hard rule, but aligning now avoids confusion when integrating with standard OTel dashboards/alerts.

---

### 6. [LOW] Metric cardinality from `McpOperationName` tag

Both counters and histograms include `McpOperationName` as a tag dimension. If the set of MCP tool/resource names grows significantly, this creates high-cardinality metric series. Currently the tool set is small and fixed (ReadTools, WriteTools, ProposalTools, Board/Capture/ProposalResources), so this is fine for now. Worth noting as a future concern if dynamic tool registration is ever added.

---

### 7. [LOW] `ActivitySource` lifecycle is correct

`TaskdeckTelemetry.McpActivitySource` and `McpMeter` are `static readonly` fields on a static class, making them true process-lifetime singletons. This is the correct pattern -- no per-request allocation of `ActivitySource` or `Meter`. Good.

---

### 8. [LOW] `Dispose` without `Complete`/`Fail` silently succeeds

`McpOperationScope.Dispose()` calls `Complete()` if neither `Complete` nor `Fail` was called. This means if a caller forgets to handle an error path, the operation is recorded as successful. The comment says "defensive," but a `_logger.LogWarning` in this path would help detect the bug in development. Current behavior is safe but slightly misleading.

---

### 9. [LOW] `string.Concat(span, string)` allocation in argument truncation

```csharp
var truncated = arguments.Length > MaxArgumentLogLength
    ? string.Concat(arguments.AsSpan(0, MaxArgumentLogLength), "...[truncated]")
    : arguments;
```
This allocates a new string on every debug-logged call with long arguments. Given the `IsEnabled(LogLevel.Debug)` guard and that this only runs for tool calls (not high-frequency), the allocation is negligible. Noting for completeness only.

---

### 10. [LOW] No test for `Fail(string)` followed by `Dispose` idempotency

Tests cover `Complete` called twice and `Dispose` without explicit complete, but there is no test for `Fail(string)` or `Fail(Exception)` followed by `Dispose` verifying the failure is not overwritten by the Dispose-path `Complete()`. The code is correct (the `_completed` flag guards this), but a test would prevent regressions.

---

### Summary

| Severity | Count | Action needed? |
|----------|-------|----------------|
| CRITICAL | 0 | -- |
| HIGH | 0 | -- |
| MEDIUM | 4 | Recommended before merge |
| LOW | 6 | Nice to have / future |

**No blocking issues found.** The code correctly guards against telemetry failures breaking MCP operations, never logs API key values at Info+ level, and uses static singletons for ActivitySource/Meter. The MEDIUM items are worth addressing for production hygiene (unused potentially-dangerous tag, exception message leakage, correlation ID gap in standalone mode, and metric unit convention).

Chris0Jeky:
## Round 2 Adversarial Review

Full diff read. All bot comments (CodeQL, Copilot, Gemini, Codex) reviewed. CI is green (all checks pass). Findings below.

---

### CRITICAL / HIGH

None found beyond the CodeQL findings below.

### CodeQL Findings (must fix)

**CodeQL cs/log-forging: Log entries created from user input** (3 instances, all in `McpTelemetryMiddleware.cs`)

- **Line 61** (`_logger.LogInformation` for request start): `context.Request.Path.Value` is user-controlled and passed as a structured parameter `{Path}`. While structured logging frameworks treat parameters as data (not interpolated), the Path value could contain newlines or control characters that forge log entries in plain-text sinks.
- **Line 86** (`_logger.LogInformation` for request completion): Same issue -- user-controlled path value flows through.
- **Line 111** (`_logger.LogError` for request failure): Same pattern.

**Fix**: Sanitize `context.Request.Path.Value` before logging (strip newlines, control chars) or log a fixed `"/mcp"` since the middleware already gates on that prefix. The path is only useful to distinguish `/mcp` vs `/mcp/sse` sub-paths, so sanitizing is the right approach.

---

### Round-1 MEDIUM findings (carry-forward, all need fix)

1. **Unused `McpApiKeyId` tag** (`TaskdeckTelemetryTags.cs:20`): Declared but never set. Remove it -- it is a PII trap for future contributors.

2. **`exception.Message` in `Activity.SetStatus`** (`McpOperationLogger.cs:166`, `McpTelemetryMiddleware.cs:96`): Exception messages may contain user content or entity names. Also `Fail(string errorMessage)` at line 211 passes user-controlled error text to `SetStatus`. **Fix**: Use `exception.GetType().Name` as the status description instead of the raw message. For the string overload, use a fixed `"ApplicationError"` description.

3. **Triple registration of `McpOperationLogger`** (`Program.cs:108, 205, 330`): Three separate `AddSingleton<McpOperationLogger>()` calls. Only one runs per process, but this is a maintenance smell. **Fix**: Extract a shared `AddMcpTelemetry()` extension method.

4. **Missing `CorrelationIdMiddleware` in standalone MCP HTTP mode** (`Program.cs:130-133`): Standalone MCP HTTP host registers `ApiKeyMiddleware` and `McpTelemetryMiddleware` but never `CorrelationIdMiddleware`, so correlation IDs from client `X-Request-Id` headers are ignored. **Fix**: Add `CorrelationIdMiddleware` before `ApiKeyMiddleware` in the standalone host.

---

### Additional findings from bot reviews (new)

5. **[MEDIUM] Missing `McpRequestDurationMs` in exception path** (`McpTelemetryMiddleware.cs:92-117`): The catch block records error counters but never records `McpRequestDurationMs`, biasing latency metrics toward successes only. Flagged by Copilot, Gemini, and Codex. **Fix**: Record duration histogram in the catch block too.

6. **[LOW] Middleware ordering: telemetry after auth** (`PipelineConfiguration.cs:89-93`): `McpTelemetryMiddleware` runs after `ApiKeyMiddleware`, so 401 responses from invalid API keys are invisible to telemetry. Both Codex and Gemini flagged this. The round-1 review noted this is intentional (user ID availability). Documenting as acceptable tradeoff -- unauthorized requests are still visible in ASP.NET Core's built-in request logging.

7. **[LOW] `Fail(Exception)` then `Dispose` idempotency not tested**: The `_completed` flag guards correctly, but no test covers the `Fail` -> `Dispose` path to prevent regression. Will add test.

---

### Summary

| Severity | Count | Action |
|----------|-------|--------|
| CodeQL (must fix) | 3 | Sanitize user input in log statements |
| MEDIUM (carry-forward) | 5 | Fix all |
| LOW | 2 | Fix #7 (test), document #6 |

Proceeding to fix all CodeQL and MEDIUM findings.

PR #902, opened 2026-04-22
feat: add Gitleaks secrets detection to CI pipeline

### LINKED ISSUE #871: CI-02: Add secrets detection to CI (Gitleaks)
## Context
Audit finding: No automated secrets detection. Hardcoded secrets could be committed without detection.

## Acceptance Criteria
- [ ] Gitleaks (or equivalent) configured as pre-commit hook or CI check
- [ ] Scans all new commits in PR for API keys, tokens, passwords
- [ ] Custom allowlist for known safe patterns (test fixtures, documentation)
- [ ] Non-blocking in ci-extended, blocking in ci-release
- [ ] `.gitleaksignore` file for false positive suppression

## Effort: M (3 hours)

Chris0Jeky:
## Summary
- Add `.gitleaks.toml` configuration extending default rules with project-specific allowlists for test fixtures, documentation examples, development placeholder secrets, and CI config strings
- Add `.gitleaksignore` for per-finding false positive suppression with documentation on how to add new entries
- Add `reusable-gitleaks.yml` workflow with configurable scan-mode (PR commits vs full history) and fail-on-findings toggle, pinned to `gitleaks/gitleaks-action@v2.3.7`
- Wire non-blocking Gitleaks scan into `ci-extended.yml` (advisory, does not gate PR merges)
- Wire blocking Gitleaks scan into `ci-release.yml` (fails the release on any detected secrets, scans full history)
- Update CI topology documentation in `ci-required.yml` header

## Design decisions
- **Non-blocking in ci-extended, blocking in ci-release**: Avoids disrupting regular development workflow while ensuring no secrets ship in releases
- **Action pinned to v2.3.7**: Supply chain safety — avoids floating major/minor tags that could be silently replaced
- **Path-based allowlists**: Test directories (`backend/tests/`, `frontend/taskdeck-web/tests/`), docs, and CI workflows are globally allowlisted since they contain only fixture/placeholder values
- **Regex allowlists for known-safe values**: Explicitly lists development JWT secrets, test OAuth fixtures, fake LLM API keys, and documentation placeholder patterns
- **SARIF artifact upload**: Gitleaks report is always uploaded for post-hoc review even when the scan passes

Closes #871

## Test plan
- [ ] Gitleaks action runs on PR creation via ci-extended
- [ ] Known safe patterns (test fixtures, dev secrets) not flagged due to allowlists
- [ ] Standard secret patterns (real API keys, tokens) would be detected
- [ ] Non-blocking in ci-extended (continue-on-error)
- [ ] Blocking in ci-release (fail-on-findings: true)
- [ ] `.gitleaksignore` is read and would suppress fingerprinted findings
- [ ] SARIF report artifact is uploaded on completion


Chris0Jeky:
## Adversarial Self-Review

### 1. Allowlist breadth analysis

**Path allowlists**: The `docs/.*` blanket exclusion is the broadest one. It suppresses all findings in documentation files, which could theoretically hide a real secret accidentally pasted into a doc. However, docs are never deployed as code, and the directory contains many legitimate placeholder/example values (runbook Redis passwords, example API key formats, config reference examples). The risk is **low** and the noise reduction is significant.

The `backend/tests/.*` and `frontend/taskdeck-web/tests/.*` exclusions are similarly broad but justified: test fixtures use deterministic placeholder secrets by design.

The `.github/workflows/.*` exclusion covers CI config which uses env var references (`${TASKDECK_...}`) and CI-only placeholder values.

**Regex allowlists**: Each regex targets a specific known-safe value. None are overly broad -- the broadest is `\$\{TASKDECK_.*\}` which matches Docker Compose env var references (not actual secrets). No real API key would match any of these patterns.

**Verdict**: Allowlists are appropriately scoped. No changes needed.

### 2. Default rule coverage

The config uses `[extend] useDefault = true`, which includes all standard Gitleaks rules for:
- AWS access keys, Azure keys, GCP service account keys
- GitHub tokens (ghp_, gho_, ghs_, ghb_)
- Generic API keys, Bearer tokens, Basic auth
- Private keys (RSA, DSA, EC, PGP)
- Database connection strings with passwords
- JWT tokens, Slack tokens, Stripe keys, etc.

This covers all common secret types referenced in the issue's acceptance criteria.

### 3. Action version pinning (supply chain safety)

The action is pinned to `gitleaks/gitleaks-action@v2.3.9` (specific patch version, not floating `v2`). This prevents silent supply-chain compromise via tag mutation. Note: for maximum safety, consider pinning to the full SHA hash in a future hardening pass, though tag-based pinning at the patch level is consistent with the project's existing convention (e.g., `actions/checkout@v6`, `rhysd/actionlint@v1.7.12`).

### 4. Bypass vectors

- **Commit squashing**: Not a concern. The PR scan mode scans the diff between PR head and base, not individual commits. Squashing does not hide content from the diff.
- **Force-push to replace commits**: The ci-extended scan runs on `synchronize` events, so force-pushed commits are re-scanned.
- **Release without PR**: The ci-release scan uses `scan-mode: full` (full history), so even content that bypassed PR scanning would be caught at release time.

### 5. Existing secrets in repo history

The `appsettings.Development.json` file contains `TaskdeckDevelopmentOnlySecretKeyChangeMe123!` which is explicitly allowlisted. No real production secrets were found in the codebase during the pre-implementation scan. The full-history release scan may surface historical findings from early commits -- these should be added to `.gitleaksignore` with fingerprints if they are confirmed false positives.

### 6. Fixed during review (commit `39ec7f68`)

- **Removed spurious `[[rules]]` block**: The custom rule for empty API key fields was unnecessary (default rules don't flag empty values) and its catch-all allowlist (`'''.*'''`) could confuse the rule engine.
- **Updated action pin**: `v2.3.7` to `v2.3.9` (latest available release).

### Remaining items (non-blocking)

- The first time the full-history scan runs in ci-release, it may surface historical findings that need `.gitleaksignore` fingerprints. This is expected and can be addressed incrementally.
- Consider adding Gitleaks to ci-nightly for regression coverage (not in scope for this issue).

Chris0Jeky:
## Adversarial Review of PR #902 (Gitleaks Secrets Detection)

### Critical Issues

**1. `scan-mode: full` does not actually scan full history (P1 -- confirms bot finding)**

Both the "PR commits" and "full repo" steps (`reusable-gitleaks.yml` lines 54-72) invoke the identical action (`gitleaks/gitleaks-action@v2.3.9`) with no differentiating inputs. The `gitleaks-action` v2 determines its scan scope entirely from the GitHub event context (`push` scans pushed commits, `pull_request` scans the PR diff). The `scan-mode` input to this reusable workflow is purely cosmetic -- it selects which `if` branch runs, but both branches execute the exact same action with the exact same behavior.

On a `release: published` event (which ci-release.yml triggers on), the action may scan nothing or error, because `release` is not a recognized event type for `gitleaks-action`. This means the "blocking release gate" claim is hollow.

**Fix**: Replace the "full repo" step with a direct CLI invocation: install gitleaks, run `gitleaks detect --source . --config .gitleaks.toml --report-format sarif --report-path results.sarif`. This gives deterministic full-history scanning regardless of the triggering event.

**2. Release event not supported by gitleaks-action (P1 -- confirms bot finding)**

`ci-release.yml` triggers on `release: types: [published]`. The `gitleaks/gitleaks-action` does not handle `release` events -- it will likely fall through to a no-op or error path. The release gate is unreliable.

**Fix**: Use CLI-based gitleaks in the full-scan path (as above). This is event-agnostic.

### Moderate Issues

**3. Overly broad path allowlists weaken the scanner**

The `.gitleaks.toml` path allowlist includes:
- `docs/.*` -- blanket suppression of all documentation files
- `.github/workflows/.*` -- blanket suppression of all workflow files
- `scripts/deploy/.*` -- blanket suppression of all deploy scripts

These are exactly the kinds of files where secrets get accidentally committed (hardcoded API keys in deploy scripts, tokens pasted into runbooks, secrets in workflow env blocks). The allowlist should be narrowed to specific safe files, or these broad path exclusions should be removed entirely -- the regex allowlist already covers known false positives.

**4. Overly generic regex allowlists**

Several regex patterns are so generic they could mask real secrets as substrings:
- `test-client`, `test-secret` -- could match parts of real OAuth credentials
- `password123` -- a common weak password that could be a real credential
- `your-secret-key-here.*` -- the `.*` suffix makes this match anything starting with that prefix
- `your-secret` -- extremely generic

These should be anchored (e.g., `^test-client$`) or scoped to specific files using per-rule allowlists instead of global regexes.

**5. Action version pinned to mutable tag, not SHA**

`gitleaks/gitleaks-action@v2.3.9` uses a Git tag that the upstream maintainer can force-push. For a security-focused tool, this is ironic -- a supply-chain attacker who compromises the gitleaks repo could replace the tag contents. Pin to a full commit SHA.

(The self-review acknowledged this but deferred it. For a security tool specifically, this should be addressed now, not later.)

### Minor Issues

**6. Step summary heredoc indentation**

Lines 86-99 of `reusable-gitleaks.yml` use an indented heredoc (`cat <<'SUMMARY'`), but the content inside is also indented. This produces markdown with leading whitespace in the GitHub Actions step summary, which may render incorrectly.

**7. `.gitleaksignore` is effectively empty**

The file contains only explanatory comments with no actual fingerprints. The comments describe categories of findings that "should" be suppressed but provide no real fingerprints. This is fine for initial setup but the comments are misleading -- they suggest findings have been reviewed when no fingerprints are actually present.

**8. `actions/upload-artifact` pinned to `@v7` (mutable tag)**

Same tag-vs-SHA concern as the gitleaks action itself, though this is consistent with the project's existing convention for GitHub-owned actions.

### Summary

The two P1 issues are real: the "full history" scan mode is cosmetic, and the release event is unsupported by the action. The fix is straightforward -- use the gitleaks CLI directly for full-history scans. The allowlist breadth is a secondary concern that reduces the scanner's effectiveness but doesn't break anything.

Chris0Jeky:
## Follow-up: Fixes Applied (commit a43edbb5)

All issues from the adversarial review and both P1 bot findings have been addressed:

### P1 Fixes (Bot Findings)

1. **Full-history scan now uses gitleaks CLI directly** (`reusable-gitleaks.yml` lines 77-96)
   - Replaced the `gitleaks/gitleaks-action` step for `scan-mode: full` with explicit CLI installation (`gitleaks v8.30.1`) and `gitleaks detect --source . --config .gitleaks.toml --report-format sarif --report-path results.sarif --verbose --exit-code 1`
   - This is event-agnostic -- works correctly on `release: published`, tag push, `workflow_dispatch`, and any other trigger
   - The PR-mode step still uses the action (which correctly handles `pull_request` events)

2. **Release event handling resolved** -- by using CLI for full-history mode, the ci-release.yml caller no longer depends on the action's event-aware behavior

### Action Pinning

3. **gitleaks-action pinned to commit SHA** `ff98106e4c7b2bc287b24eaf42907196329070c7` (v2.3.9) instead of mutable tag

### Allowlist Tightening

4. **Removed broad path exclusions** from `.gitleaks.toml`:
   - Removed `docs/.*` (docs can contain accidentally-pasted secrets)
   - Removed `scripts/deploy/.*` (deploy scripts can contain hardcoded keys)
   - Removed `.github/workflows/.*` (workflows can contain secrets in env blocks)
   - Test directories remain excluded (deterministic fixture values by design)

5. **Removed overly generic regex patterns**:
   - Removed `test-client`, `test-secret` (could mask real OAuth creds)
   - Removed `password123`, `wrong-password` (could mask real weak passwords)
   - Removed `your-secret` (too generic)
   - Removed trailing `.*` from `your-secret-key-here` pattern
   - Tightened env var regex from `\$\{TASKDECK_.*\}` to `\$\{TASKDECK_[A-Z_]+\}` (prevents runaway matching)

### Minor Fixes

6. **Fixed heredoc indentation** in step summary -- content is now left-aligned so GitHub renders the markdown table correctly

7. **Cleaned up `.gitleaksignore`** -- removed misleading category comments that implied fingerprints existed; replaced with a single note explaining the file is empty pending the first full-history scan

### Not Fixed (Pre-existing, Unrelated)

- **Frontend unit test failure** (`ActivityView.spec.ts > fetches entity history from discoverable selectors`): This is a pre-existing flaky test that passes on main and on Windows. The test expects `board-1` but receives `board-2`, indicating a test isolation issue. Not caused by this PR's changes.

PR #904, opened 2026-04-22
fix: replace sync-over-async in WorkspaceService with proper await

### LINKED ISSUE #847: PERF-11: Fix synchronous I/O blocking in WorkspaceService
## Context
Audit finding: `WorkspaceService.GetHomeAsync()` uses `.Result` on async tasks, blocking the thread pool. Under concurrent load (10+ DAU), this causes thread pool starvation.

## Acceptance Criteria
- [ ] `.Result` calls replaced with `await Task.WhenAll()` pattern
- [ ] No sync-over-async patterns in WorkspaceService
- [ ] Backend test suite passes
- [ ] Verify async behavior in existing workspace tests

## Effort: S (30 minutes)

Chris0Jeky:
## Summary
- Replace `.Result` calls with `await` in `GetOnboardingForPreferenceAsync` after `Task.WhenAll` completes
- Parallelize 7 independent database queries in `GetHomeAsync` using `Task.WhenAll` (previously sequential)
- Parallelize 5 independent database queries in `GetTodayAsync` using `Task.WhenAll` (previously sequential)
- Eliminates sync-over-async patterns that can cause thread pool starvation under concurrent load

Closes #847

## Test plan
- [x] All `.Result` calls removed from WorkspaceService (verified via grep)
- [x] Async call chain verified end-to-end
- [x] All 22 WorkspaceService unit tests pass
- [x] Full backend test suite passes (2038 tests, 0 failures)

### DIFF
--- a/backend/src/Taskdeck.Application/Services/WorkspaceService.cs
+++ b/backend/src/Taskdeck.Application/Services/WorkspaceService.cs
@@ -27,26 +27,45 @@ public class WorkspaceService : IWorkspaceService
         if (userId == Guid.Empty)
             return Result.Failure<WorkspaceHomeDto>(ErrorCodes.ValidationError, "User ID cannot be empty");
 
-        var preference = await EnsurePreferenceAsync(userId, cancellationToken);
-        var captureSummary = await _unitOfWork.LlmQueue.GetCaptureSummaryByUserAsync(userId, cancellationToken);
         var recentCutoff = DateTimeOffset.UtcNow.Subtract(RecentBoardWindow);
-        var capturesNeedingTriage = captureSummary.NewCount + captureSummary.FailedCount;
-        var capturesInProgress = captureSummary.TriagingCount;
-        var capturesReadyForFollowUp = captureSummary.TriagedCount;
-        var proposalsPendingReview = await _unitOfWork.AutomationProposals.CountPendingReviewByUserIdAsync(userId, cancellationToken);
-        var hasReviewedProposal = await _unitOfWork.AutomationProposals.HasReviewedByUserIdAsync(userId, cancellationToken);
-        var totalBoards = await _unitOfWork.Boards.CountReadableByUserIdAsync(userId, includeArchived: false, cancellationToken);
-        var recentBoardsCount = await _unitOfWork.Boards.CountReadableUpdatedSinceAsync(
+
+        // Launch all independent queries in parallel
+        var preferenceTask = EnsurePreferenceAsync(userId, cancellationToken);
+        var captureSummaryTask = _unitOfWork.LlmQueue.GetCaptureSummaryByUserAsync(userId, cancellationToken);
+        var proposalsPendingReviewTask = _unitOfWork.AutomationProposals.CountPendingReviewByUserIdAsync(userId, cancellationToken);
+        var hasReviewedProposalTask = _unitOfWork.AutomationProposals.HasReviewedByUserIdAsync(userId, cancellationToken);
+        var totalBoardsTask = _unitOfWork.Boards.CountReadableByUserIdAsync(userId, includeArchived: false, cancellationToken);
+        var recentBoardsCountTask = _unitOfWork.Boards.CountReadableUpdatedSinceAsync(
             userId,
             recentCutoff,
             includeArchived: false,
             cancellationToken);
-        var recentBoardCandidates = (await _unitOfWork.Boards.GetRecentReadableByUserIdAsync(
-                userId,
-                RecentBoardLimit,
-                includeArchived: false,
-                cancellationToken))
-            .ToList();
+        var recentBoardCandidatesTask = _unitOfWork.Boards.GetRecentReadableByUserIdAsync(
+            userId,
+            RecentBoardLimit,
+            includeArchived: false,
+            cancellationToken);
+
+        await Task.WhenAll(
+            preferenceTask,
+            captureSummaryTask,
+            proposalsPendingReviewTask,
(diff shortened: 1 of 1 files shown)


Chris0Jeky:
## Adversarial Self-Review

### Changes reviewed
1. **GetOnboardingForPreferenceAsync**: `.Result` calls replaced with `await` after `Task.WhenAll`. This is safe since `Task.WhenAll` has already completed all tasks -- the `await` will return synchronously.
2. **GetHomeAsync**: 7 independent queries now run in parallel via `Task.WhenAll` instead of sequentially.
3. **GetTodayAsync**: 5 independent queries now run in parallel via `Task.WhenAll` instead of sequentially.

### Potential concerns evaluated

**Task.WhenAll failure modes**: If one task fails, `Task.WhenAll` will propagate the first exception. The other tasks may still run to completion in the background, but their results won't be accessed. This is the correct behavior -- if any query fails, the entire operation should fail. No partial-success state is exposed.

**Deadlock risk**: None. All methods use `await` (not `.Result` or `.Wait()`), and ASP.NET Core doesn't have a `SynchronizationContext`, so there's no capture-and-deadlock risk.

**SQLite concurrency**: SQLite serializes writes but allows concurrent reads. All parallelized queries are reads, so this is safe. The `EnsurePreferenceAsync` call (which may do a write on first run) is also safe as a Task because it will simply serialize at the SQLite level if needed.

**Return type correctness**: All task variables have correct types matching their `await` expressions. Verified the `.ToList()` call on `recentBoardCandidatesTask` and `accessibleBoardsTask` is preserved.

### Backend-wide sync-over-async scan

Grep for `.Result`, `.GetAwaiter().GetResult()`, and `.Wait()` across `backend/src/`:

- **`RedisBackplaneHealthCheck.cs:111`**: `_lazyConnection.Value.GetAwaiter().GetResult().Dispose()` -- This is inside a synchronous `Dispose()` method (required by `IDisposable`). Acceptable trade-off; no async alternative without implementing `IAsyncDisposable`. Not in a hot path. **No action needed.**

No other sync-over-async patterns found in production code.

### Verdict
No issues found. The change is clean, correctly preserves behavior, and all 2038 backend tests pass.

Chris0Jeky:
## Adversarial Review -- Critical DbContext Thread-Safety Bug

### Summary
This PR introduces **two new `Task.WhenAll` parallelizations** in `GetHomeAsync` and `GetTodayAsync` that are **unsafe**. EF Core's `DbContext` is not thread-safe, and all repositories in `_unitOfWork` share the same scoped `DbContext` instance. Running multiple EF queries in parallel on the same context will cause `InvalidOperationException` ("A second operation was started on this context instance before a previous operation completed") under any real concurrency.

Both `chatgpt-codex-connector` and `gemini-code-assist` flagged this correctly as P1/Critical.

### What is safe
- **`GetOnboardingForPreferenceAsync`**: The `.Result` to `await` change is correct and desirable. The `Task.WhenAll` there was **pre-existing** (not introduced by this PR). Note: that pre-existing `Task.WhenAll` has the same thread-safety issue, but that is a separate concern and out of scope for this PR.

### What must be fixed
1. **`GetHomeAsync` (lines 32-56)**: Revert `Task.WhenAll` parallelization. Keep queries sequential with `await`.
2. **`GetTodayAsync` (lines 116-131)**: Same -- revert `Task.WhenAll` parallelization. Keep sequential `await`.

### Why the self-review missed this
The self-review comment on this PR states "SQLite serializes writes but allows concurrent reads. All parallelized queries are reads, so this is safe." This conflates SQLite-level concurrency (which does allow concurrent readers) with EF Core DbContext thread-safety (which does not). Even if the underlying database supports concurrent reads, the `DbContext` itself -- its change tracker, connection management, and query pipeline -- is not designed for concurrent use from multiple threads. The exception happens at the EF Core layer, not the SQLite layer.

### Correct fix approach
Replace `Task.WhenAll` with sequential `await` calls in both methods. This preserves the good part of the PR (replacing `.Result` with `await` in `GetOnboardingForPreferenceAsync`) while removing the thread-safety hazard.

### Pre-existing issue (out of scope)
`GetOnboardingForPreferenceAsync` (lines 368-375) already had a `Task.WhenAll` with the same DbContext-sharing problem before this PR. That should be fixed in a follow-up.

I will push a fix commit momentarily.

Chris0Jeky:
## Fix Applied (55915d5a)

Pushed a fix commit that addresses all four critical bot findings:

### Changes made
1. **GetHomeAsync**: Reverted `Task.WhenAll` parallelization back to sequential `await` calls. The method now matches the original main branch implementation.
2. **GetTodayAsync**: Same -- reverted `Task.WhenAll` to sequential `await` calls.
3. **GetOnboardingForPreferenceAsync**: Replaced the **pre-existing** `Task.WhenAll` + `.Result` pattern with sequential `await` calls. This fixes both the sync-over-async issue (the original PR goal) and the DbContext thread-safety bug that predated this PR.

### What was preserved
The `.Result` to `await` improvement in `GetOnboardingForPreferenceAsync` is retained -- this was the valid part of the original change.

### Net diff from main
The only change vs main is in `GetOnboardingForPreferenceAsync`: three `Task` variables + `Task.WhenAll` + `.Result` accesses replaced with three sequential `await` calls. `GetHomeAsync` and `GetTodayAsync` are identical to main.

### Verification
- Build: 0 errors, 19 warnings (pre-existing)
- Application tests: 2146 passed, 0 failed
- Architecture tests: 8 passed, 0 failed
- Workspace API tests: 11 passed, 0 failed

PR #915, opened 2026-04-22
feat: add SAST scanning to CI pipeline (CI-01)

### LINKED ISSUE #870: CI-01: Add SAST scanning to CI (Semgrep or equivalent)
## Context
Audit finding: No static application security testing (SAST) in CI. Security issues are found manually rather than automatically.

## Acceptance Criteria
- [ ] Semgrep (or equivalent SAST tool) configured for C# and TypeScript/Vue
- [ ] Runs in ci-extended lane (label: security) and ci-nightly
- [ ] Reports findings as CI step summary
- [ ] Non-blocking initially (advisory), with path to gating on high/critical
- [ ] Custom rules for Taskdeck-specific patterns (e.g., missing [Authorize])

## Effort: M (4 hours)

Chris0Jeky:
## Summary

- Add Semgrep-based SAST scanning to the CI pipeline, covering both C# and TypeScript/Vue codebases
- Create custom Taskdeck rules in `.semgrep/` for project-specific security patterns (missing `[Authorize]`, raw SQL injection, logging secrets, architecture violations, XSS, insecure token storage)
- Wire scanning into ci-extended (triggered by `security` label) and ci-nightly (unconditional) as non-blocking advisory checks
- Add `scripts/ci/summarize-sast-findings.mjs` to produce GitHub step summaries and machine-readable JSON output
- Add ADR-0031 documenting the Semgrep choice, alternatives considered, and enforcement roadmap

## Design Decisions

- **Advisory first**: `enforce-findings: false` by default. Flip to `true` after baseline triage to gate on ERROR-level findings.
- **Reusable workflow pattern**: Follows ADR-0013 topology with `reusable-sast-scanning.yml` called from both lane workflows.
- **pip install over container**: Uses `setup-python` + `pip install semgrep` instead of `semgrep/semgrep` container to allow `setup-node` for the summary script in the same job.
- **Registry + custom rules**: `p/csharp`, `p/typescript`, `p/jwt` from Semgrep registry plus 10 custom rules in `.semgrep/`.

## Test Plan

- [ ] Verify workflow YAML is valid (actionlint in ci-extended will check this)
- [ ] Trigger ci-extended manually or via `security` label to confirm Semgrep runs end-to-end
- [ ] Check GitHub step summary renders correctly with finding counts and collapsible details
- [ ] Verify `.semgrepignore` excludes build artifacts, test fixtures, and generated code
- [ ] Confirm artifacts are uploaded (semgrep-results.json, summary.md, summary.json)
- [ ] Validate custom rules catch intended patterns (e.g., HealthController without [Authorize] should be excluded by metavariable-regex)

Closes #870


Chris0Jeky:
## Adversarial Self-Review

### Issues found and fixed (commit 8b521b2c)

1. **Semgrep version not pinned** -- `pip install semgrep` without a version means nightly could break unpredictably. Fixed: pinned to `semgrep==1.113.0`.

2. **Stdout/stderr redirect swallowed errors** -- The summarize step used `> file 2>&1` which hid both script output and error messages from CI logs. Fixed: removed redirect so errors appear in workflow logs while `--output-markdown` writes the file.

3. **Redundant `--json` flag** -- Used both `--json` (stdout) and `--json-output` (file). The `--json-output` flag implicitly enables JSON mode, so `--json` was redundant and could cause double JSON output. Fixed: removed `--json`.

4. **ADR-0031 text mismatch** -- ADR mentioned "semgrep/semgrep:latest container image" but the workflow uses pip install. Fixed: updated ADR text.

### Remaining considerations (not blocking)

5. **False positive risk on missing-authorize rule** -- The `metavariable-regex` excludes `Health` and `Auth$` but not other intentionally-public controllers (if any exist in the future). This is acceptable because the rule is WARNING-level and advisory -- false positives are surfaced for human triage, not gating.

6. **Semgrep registry rule noise** -- The `p/csharp`, `p/typescript`, and `p/jwt` community rulesets may produce findings on patterns that are acceptable in Taskdeck. The first nightly run should be triaged to identify rules that need `.semgrepignore` additions or `nosemgrep` inline annotations.

7. **No test for the summarization script** -- The existing `summarize-dependency-security-signals.mjs` has a companion `.test.mjs` file. A follow-up PR could add `summarize-sast-findings.test.mjs` for parity, though the script is straightforward I/O.

8. **`p/csharp` and `p/typescript` registry scope** -- These rulesets include rules beyond security (e.g., correctness, best-practice). If the volume is too high, consider switching to `p/csharp-security` and `p/typescript-security` subsets.

9. **Scan target scope** -- `.semgrepignore` excludes `scripts/` entirely. This means CI helper scripts aren't scanned. Acceptable since they run in CI only, but worth noting.

Chris0Jeky:
## Adversarial Review -- PR #915 (SAST Scanning)

I performed a thorough adversarial review of all changed files, evaluated all 8 bot review comments, and pushed fixes for the legitimate findings. Summary below.

---

### Issues Found and Fixed (commit 0c0c00a4)

**1. Controller exclusion regex is wrong (severity: medium)**
`.semgrep/taskdeck-csharp.yml` line 38

The regex `^(?!Health|Auth$).*Controller$` does not exclude `AuthController`. The `Auth$` lookahead only matches if the entire string is literally "Auth", which it never is -- the class name is `AuthController`. Fixed to `^(?!(Health|Auth)Controller$).*Controller$`.

*Note:* This is currently moot because `AuthController` inherits from `AuthenticatedControllerBase` (not `ControllerBase`), so the Semgrep pattern wouldn't match it anyway. But the regex is semantically wrong and would bite someone adding a new Auth-prefixed controller inheriting from `ControllerBase`.

**Gemini raised this -- legitimate, fixed.**

**2. Unhandled-fetch rule produces false positives and has broken `fix` template (severity: high)**
`.semgrep/taskdeck-typescript.yml` lines 57-60

Two problems:
- The `pattern: fetch(...).then(...)` matches even when followed by `.catch()`, producing false positives.
- The `fix:` template uses literal `...` which Semgrep would emit as-is, producing broken code like `fetch(...).then(...).catch(...)`.

Fixed by switching to `patterns` with `pattern-not: fetch(...).then(...).catch(...)` and removing the `fix` entirely (this is an advisory WARNING rule; auto-fix is not appropriate).

**Gemini raised this -- legitimate, fixed.**

**3. Enforcement bypass on scanner failure (severity: critical)**
`scripts/ci/summarize-sast-findings.mjs`

When enforcement is enabled and Semgrep crashes (exit code >= 2) or produces no/invalid JSON output, the script silently creates an empty result set and reports "No SAST findings. The codebase passed all configured rules." This means enforcement mode can be trivially bypassed by any condition that causes Semgrep to fail.

Fixed by tracking `scanIncomplete` (parse failures or exit codes other than 0 or 1) and treating it as an enforcement failure. The summary now reports "Scan incomplete" with the exit code instead of a false pass.

**Codex P1 raised this -- legitimate, fixed.**

**4. Misleading "passed" on parse failure (severity: medium)**
`scripts/ci/summarize-sast-findings.mjs`

When `readFileSync`/`JSON.parse` fails, the catch block created `{ results: [], errors: [] }` which downstream reports as a clean scan. Fixed by injecting a parse-failure error message into the `errors` array so the summary always shows the scan error section.

**Codex P2 raised this -- legitimate, fixed.**

**5. Pipe characters break markdown table (severity: low)**
`scripts/ci/summarize-sast-findings.mjs` line 167

Semgrep messages containing `|` would break the markdown table rendering. Fixed by escaping `|` to `\|` before inserting into table cells.

**Gemini raised this -- legitimate, fixed.**

**6. Status messages pollute stdout (severity: low)**
`scripts/ci/summarize-sast-findings.mjs` lines 244, 249

The "Markdown summary written to" and "JSON summary written to" messages used `console.log` (stdout) alongside the markdown output. While the current workflow uses the file for `GITHUB_STEP_SUMMARY`, not stdout, the script's own doc comment says stdout is for piping to `GITHUB_STEP_SUMMARY`. Switched to `console.error` for consistency and defensive correctness.

**Gemini raised this -- legitimate, fixed.**

**7. ADR mentions "container image" but implementation uses pip install (severity: low)**
`docs/decisions/ADR-0031-sast-scanning-semgrep.md` line 55

The Consequences section references "Semgrep container image adds ~1 min to job startup" but the actual implementation uses `pip install semgrep==1.113.0`. Fixed to "pip install".

**Gemini raised this -- legitimate, fixed.**

---

### Bot Comments Evaluated but NOT Acted On

**Gemini's ADR line 32 comment ("update ADR to reflect pip install"):** The ADR line 32 already correctly says "Installs Semgrep via pip (version-pinned)." The stale "container image" reference was only in the Consequences section (line 55), which I fixed. The line Gemini pointed at was already correct.

---

### Additional Findings from My Own Review

**A. No `--exclude` flag for semgrep scan (observation, not a fix)**
The workflow uses `.semgrepignore` to control scan scope, which is the recommended approach. The `--exclude` CLI flag is redundant with `.semgrepignore`. No issue here.

**B. Semgrep exit code semantics are correctly handled**
Semgrep exit codes: 0 = clean scan, 1 = findings found, >= 2 = error. The script now correctly treats 0 and 1 as "scan completed" and anything else as "scan failed". The workflow uses `set +e` to capture the exit code without failing the step prematurely. This is correct.

**C. Workflow permissions are minimal (good)**
`permissions: contents: read` is the minimum needed. No write permissions exposed to the SAST job.

**D. No SEMGREP_APP_TOKEN or registry auth required**
The workflow uses `SEMGREP_SEND_METRICS: "off"` and public registry rules (`p/csharp`, `p/typescript`, `p/jwt`). No auth token needed. Correct for open-source usage.

**E. Test files are scanned (acceptable)**
`.semgrepignore` excludes test fixtures and snapshots but not test `.cs`/`.ts` files themselves. This is acceptable -- security rules like SQL injection or innerHTML should still fire on test code that might be copy-pasted into production. The console.log rule already excludes test files via its own `paths` config.

**F. The TD-CS-003 (logging sensitive data) rule may have false positives**
The metavariable regex `(?i).*(password|secret|token|apikey|...).*` is quite broad. A variable named `tokenCount` or `resetPasswordResult` would trigger it. This is acknowledged in the ADR ("False positives are possible") and the rule is WARNING severity (not ERROR), so it won't block enforcement. Acceptable for now; tune after baseline triage.

**G. `hasFlag` function is safe from collision**
`args.includes('--enforce')` will not match `--enforce-findings` since they are separate tokens in the argv array. No issue.

---

### Verdict

All 7 bot findings were legitimate and have been fixed. The most critical fix is the enforcement bypass on scanner failure (#3), which could have allowed a security gate to silently pass without a valid scan result. The fixes are in commit 0c0c00a4.

The PR is now ready for merge after CI passes on the fix commit.

PR #919, opened 2026-04-22
feat: add session timeout warning before JWT expiry (FE-20)

### LINKED ISSUE #861: FE-20: Session timeout warning before JWT expiry
## Context
Audit finding (LOW): JWT expires silently. User is redirected to login without warning, potentially losing unsaved work.

## Acceptance Criteria
- [ ] Toast shown 5 minutes before token expiry
- [ ] "Extend Session" button re-authenticates silently (if possible) or warns to save work
- [ ] Countdown visible in toast
- [ ] No warning spam (shown once per expiry cycle)

## Effort: S (3 hours)

Chris0Jeky:
## Summary

- Adds a `useSessionTimeout` composable that monitors JWT expiry and shows a persistent warning banner 5 minutes before the token expires
- Warning includes a live countdown timer (mm:ss format), "Extend Session" button (attempts `/auth/refresh`, graceful fallback toast on failure), and "Dismiss" button
- Per-token deduplication prevents warning spam after dismissal
- Skips entirely in demo mode and unauthenticated states
- `SessionTimeoutWarning.vue` component wired into `App.vue` for global visibility
- `authApi.refreshToken()` added (backend endpoint not yet implemented; composable handles failure gracefully)

Closes #861

## Test plan

- [x] 19 vitest tests covering all composable logic paths (timer scheduling, countdown, dismiss, dedup, extend success/failure/idempotency, demo mode, logout, teardown, token reschedule)
- [x] TypeScript typecheck passes (`npm run typecheck`)
- [x] ESLint passes (`npm run lint`) -- 0 new warnings
- [x] Production build passes (`npm run build`)
- [x] Full test suite passes (2571 tests, 211 files)
- [ ] Manual: verify warning appears ~5min before real JWT expiry
- [ ] Manual: verify Extend Session button shows fallback toast (no backend refresh endpoint yet)
- [ ] Manual: verify Dismiss prevents re-show for same session
- [ ] Manual: verify demo mode never shows warning

### DIFF
--- a/frontend/taskdeck-web/src/App.vue
+++ b/frontend/taskdeck-web/src/App.vue
@@ -2,6 +2,7 @@
 import { computed, onMounted, watch } from 'vue'
 import { useRoute } from 'vue-router'
 import ToastContainer from './components/common/ToastContainer.vue'
+import SessionTimeoutWarning from './components/common/SessionTimeoutWarning.vue'
 import AppShell from './components/shell/AppShell.vue'
 import ErrorBoundary from './components/ErrorBoundary.vue'
 import { useSessionStore } from './store/sessionStore'
@@ -48,6 +49,7 @@ watch(
       <router-view />
     </ErrorBoundary>
     <ToastContainer />
+    <SessionTimeoutWarning />
   </div>
 </template>
 
--- a/frontend/taskdeck-web/src/api/authApi.ts
+++ b/frontend/taskdeck-web/src/api/authApi.ts
@@ -79,7 +79,13 @@ export const authApi = {
   },
 
   async refreshToken(): Promise<AuthResponse> {
-    const { data } = await http.post<AuthResponse>('/auth/refresh')
+    // skipAuth401: prevent the 401 interceptor from redirecting to /login
+    // before the caller's catch block can handle the error gracefully.
+    // skipRetry: refresh is not idempotent and should fail fast.
+    const { data } = await http.post<AuthResponse>('/auth/refresh', null, {
+      skipAuth401: true,
+      skipRetry: true,
+    } as Record<string, unknown>)
     return data
   },
 }
--- a/frontend/taskdeck-web/src/api/authApi.ts
+++ b/frontend/taskdeck-web/src/api/authApi.ts
@@ -77,4 +77,9 @@ export const authApi = {
   async disableMfa(request: MfaVerifyRequest): Promise<void> {
     await http.post('/auth/mfa/disable', request)
   },
+
+  async refreshToken(): Promise<AuthResponse> {
+    const { data } = await http.post<AuthResponse>('/auth/refresh')
+    return data
+  },
 }
--- a/frontend/taskdeck-web/src/api/http.ts
+++ b/frontend/taskdeck-web/src/api/http.ts
@@ -53,8 +53,11 @@ http.interceptors.response.use(
     if (error.response) {
       console.error('API Error:', error.response.data)
 
-      // Handle 401 - clear session and redirect to login (skip in demo mode)
-      if (error.response.status === 401 && !isDemoMode) {
+      // Handle 401 - clear session and redirect to login (skip in demo mode).
+      // Callers can set `skipAuth401` on the request config to suppress this
+      // behaviour (e.g. token refresh attempts that want to handle 401 locally).
+      const skipAuth401 = (error.config as Record<string, unknown> | undefined)?.skipAuth401 === true
+      if (error.response.status === 401 && !isDemoMode && !skipAuth401) {
         tokenStorage.clearAll()
         const pathname = window.location.pathname
         const currentPath = `${pathname}${window.location.search}`
--- a/frontend/taskdeck-web/src/components/common/SessionTimeoutWarning.vue
+++ b/frontend/taskdeck-web/src/components/common/SessionTimeoutWarning.vue
@@ -1,0 +1,95 @@
+<script setup lang="ts">
+import { computed } from 'vue'
+import { useSessionTimeout } from '../../composables/useSessionTimeout'
+
+const { showWarning, secondsRemaining, extending, dismiss, extend } = useSessionTimeout()
+
+const formattedTime = computed(() => {
+  const secs = secondsRemaining.value
+  if (secs === null || secs <= 0) return '0:00'
+  const minutes = Math.floor(secs / 60)
+  const seconds = secs % 60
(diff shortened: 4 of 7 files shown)


Chris0Jeky:
## Adversarial Self-Review

### Issues found and fixed

1. **Watcher leak in teardown()** -- `teardown()` cleared timers and state but did not stop the Vue `watch()`. When the composable was used outside a component (tests) or teardown was called manually, the watcher would continue firing. Fixed in commit 4146d5d6: `teardown()` now calls `stopWatch()`.

### Reviewed and acceptable

2. **Multiple tabs** -- Each tab has its own composable instance. If a user extends the session in one tab, other tabs will not automatically pick up the new token. This is expected behavior since there is no cross-tab sync mechanism yet (and the backend refresh endpoint does not exist). When/if refresh is implemented, a `BroadcastChannel` or `storage` event listener could sync tokens across tabs. Not blocking for this PR.

3. **Token refresh during countdown** -- If the token is refreshed externally (e.g., by another mechanism), the watcher catches the `session.token` change and reschedules the warning timer correctly. Test: "reschedules warning when token changes to a new one".

4. **Demo mode** -- The composable explicitly checks `session.isDemo` and skips entirely. Test: "does not show warning when in demo mode".

5. **Backend refresh endpoint missing** -- The `authApi.refreshToken()` call will 404 until the backend implements `/auth/refresh`. The composable gracefully catches the error and shows a toast: "Could not extend session. Please save your work and log in again." The warning banner remains visible so the user can still save work. Test: "extend shows toast on failure and keeps warning visible".

6. **Dynamic imports in extend()** -- The non-test path uses `await import(...)` for authApi, sessionStore, and tokenStorage to avoid circular imports. These are resolved from the same Vite bundle, so they work correctly in both dev and production builds.

7. **Countdown accuracy** -- Uses `setInterval(1000)` for display ticking but computes seconds from `nowFn()` vs stored `tokenExpiryMs`, so drift accumulation is impossible -- each tick recalculates from the real clock.

8. **Accessibility** -- Warning uses `role="alert"`, `aria-live="assertive"` on the container and `aria-live="polite"` on the countdown span. Dismiss button has `aria-label`.

### Test coverage assessment

19 tests cover all composable paths. The component (`SessionTimeoutWarning.vue`) does not have its own test file -- it is a thin rendering layer over the composable. Adding a component-level test would improve coverage but is not blocking.

### Remaining work (not in scope)

- Backend `/auth/refresh` endpoint (separate issue)
- Cross-tab token sync via BroadcastChannel
- Component-level test for SessionTimeoutWarning.vue

Chris0Jeky:
## Adversarial Review -- PR #919 (Session Timeout Warning)

I have read the full diff, all 4 bot inline comments, and done my own independent review. Summary of findings below.

---

### CRITICAL: No backend `/auth/refresh` endpoint exists

The new `authApi.refreshToken()` calls `POST /auth/refresh`, but **this endpoint does not exist** in `AuthController.cs` (or anywhere in the backend). The "Extend Session" button will always fail -- returning either a 404 or triggering the HTTP client's 401 interceptor (see next issue). This renders the extend functionality dead-on-arrival.

**Action required:** Either (a) add a `/auth/refresh` endpoint to the backend in a separate PR that ships before or alongside this one, or (b) remove the "Extend Session" button until the endpoint exists and replace it with an informational message directing users to save their work and re-login.

---

### HIGH: 401 interceptor will hijack refresh failure (agrees with chatgpt-codex-connector bot)

The bot correctly identified that `http.ts` response interceptor unconditionally clears `tokenStorage` and redirects to `/login` on any 401. If the refresh call fails with 401 (likely, since the endpoint doesn't exist or the token is near-expiry), the interceptor redirects **before** the composable's `catch` block can show the friendly toast. This silently interrupts the user's work -- the exact scenario the feature is supposed to prevent.

**Action required:** If the refresh endpoint is implemented, the refresh call should either bypass the 401 interceptor (e.g. via a `skipAuth401` flag on the request config) or use a separate axios instance.

---

### HIGH: Extend logic bypasses `setSession`, causing incomplete state updates (agrees with both bots)

Both bots correctly flagged this. In `extend()`, when `!deps?.refreshToken` (production path), the code manually sets `store.token` and `store.expiresAt` but **skips**: `userId`, `username`, `email`, `defaultRole`, and `persistSessionSnapshot()`. The store's `setSession(data: AuthResponse)` method handles all of this centrally, but it is not exported from the store's return block.

**Fix:** Add a `refreshSession` action to `sessionStore.ts` that calls `authApi.refreshToken()` and then passes the response through `setSession()`. The composable should call `await session.refreshSession()` instead of manually mutating store state.

---

### MEDIUM: Inline type duplication (agrees with gemini-code-assist bot)

Line 45 inlines the full `{ token: string; user: { id: string; ... } }` type instead of importing `AuthResponse` from `../types/auth`. This creates a maintenance burden and can silently drift out of sync.

**Fix:** Import and use `AuthResponse` type for the `refreshToken` dependency.

---

### LOW: Nested `aria-live` regions may cause double announcements

The container `div` has `aria-live="assertive"` and the countdown `span` inside it has `aria-live="polite"`. Nested live regions can cause screen readers to announce changes at both levels. Consider removing the inner `aria-live` on the span since the outer container already handles it, or move the `aria-live="polite"` to a visually-hidden element that only contains the countdown text.

---

### Observations (no fix needed now)

1. **Timer cleanup is solid.** The watcher properly calls `resetState()` on token change, `onUnmounted` is correctly guarded by `getCurrentInstance()`, and `teardown()` stops the watcher. No timer leak risk.
2. **JWT parsing is safe for this use case.** The `atob` concern is a non-issue here -- the code correctly normalizes base64url to base64 before calling `atob`, and JWT payloads only contain ASCII JSON for standard claims like `exp`.
3. **Test coverage is good** for the composable logic (timer scheduling, countdown, dismiss, extend, teardown, token change, demo mode). Missing: a test for the production `!deps?.refreshToken` branch, and a test for the `extend()` error path checking that the warning remains visible AND a toast was shown.
4. **`COUNTDOWN_TICK_MS` is declared and used correctly** -- no dead constant.

---

### Summary

The core timing/warning/dismiss logic is well-implemented and well-tested. However, the "Extend Session" feature is non-functional because the backend endpoint does not exist, and even if it did, the HTTP interceptor would hijack the error path. The extend logic also bypasses the store's centralized state management.

**Recommendation:** Before merging, either implement the backend refresh endpoint or remove the extend button. Fix the store state bypass and interceptor interaction regardless.

PR #966, opened 2026-04-23
Add DbContext connection resilience (timeout and retry)

### LINKED ISSUE #952: PERF-14: Add DbContext connection resilience (timeout and retry)
## Context

The 2026-04-16 audit (docs/AUDIT.md, Backend section) flagged: "No connection timeout or retry policy on DbContext" in `DependencyInjection.cs`. While SQLite rarely has connection issues, adding configurable command timeouts and basic retry policy improves resilience for the PostgreSQL migration path (ADR-0023).

## Acceptance Criteria

- [ ] Add configurable `CommandTimeout` to DbContext options (default: 30s)
- [ ] Add basic retry policy via `EnableRetryOnFailure` or equivalent for the SQLite provider where supported
- [ ] Settings should be configurable via `appsettings.json` (new `DatabaseSettings` section)
- [ ] Add settings validation via the existing `ValidateOnStart` pattern
- [ ] Add unit tests for the configuration
- [ ] Full backend solution builds and all tests pass

## Suggested Branch Name

`fix/perf-14-dbcontext-resilience`

## Priority

Priority III — hardening

Chris0Jeky:
## Summary

- Adds `DatabaseSettings` class with `CommandTimeoutSeconds` (default: 30, range 1-300) and `MaxRetryCount` (default: 3, range 0-10) with data annotation validation
- Applies `CommandTimeout` to SQLite `DbContextOptionsBuilder` from configuration
- Documents that `EnableRetryOnFailure` is not supported by SQLite and will be enabled on PostgreSQL migration
- Registers settings via `RegisterValidatedOptions` for fail-fast startup validation
- Adds `Database` section to `appsettings.json`
- Adds 7 validation tests covering defaults, boundary values, and out-of-range rejection

Closes #952

## Test plan

- [x] All 5,014 existing tests pass (0 failures)
- [x] 7 new `DatabaseSettings` validation tests pass
- [x] Integration test `App_StartsSuccessfully_WithDefaultConfiguration` confirms no startup regression
- [ ] Verify CI passes on PR


Chris0Jeky:
## Self-review: Adversarial analysis

### Defaults
- **CommandTimeoutSeconds = 30**: Sensible for a local-first SQLite app. 30s is the EF Core default and matches the LLM provider timeout. Acceptable.
- **MaxRetryCount = 3**: Standard retry count. Since this is currently a no-op for SQLite, the value only matters as documentation of intent for future PostgreSQL migration.

### Validation correctness
- Range [1, 300] for CommandTimeoutSeconds prevents 0 (which would mean "no timeout" in some ADO.NET providers, potentially dangerous) and caps at 5 minutes.
- Range [0, 10] for MaxRetryCount allows disabling retries (0) and caps at a reasonable maximum.
- Data annotations are validated via `ValidateDataAnnotations().ValidateOnStart()`, consistent with all other settings.
- No cross-property validation needed (the two properties are independent), so no `IValidateOptions<DatabaseSettings>` validator was created. This is correct.

### SQLite limitation handling
- The code correctly avoids calling `EnableRetryOnFailure` on the SQLite provider (which would throw at runtime).
- The `MaxRetryCount` property is validated and documented as future-use for PostgreSQL migration.
- Inline comments in `DependencyInjection.cs` explain the limitation and provide the exact code to add when migrating.

### Existing test impact
- All 5,014 tests pass (0 failures).
- The integration test `App_StartsSuccessfully_WithDefaultConfiguration` confirms the new `Database` section doesn't break startup.
- The `CommandTimeout(30)` now being explicitly set matches SQLite's existing default behavior, so no behavioral change.

### Configuration reference
- Added `Database` section to `docs/platform/CONFIGURATION_REFERENCE.md` with table of contents entry.

### No issues found.

Chris0Jeky:
## Adversarial Code Review

### Summary
Clean, well-structured change. Follows existing codebase conventions precisely. I found one substantive issue (misleading no-op property) and several minor observations.

---

### Issue 1 (Medium): MaxRetryCount is a validated, configured, defaulted no-op

`MaxRetryCount` has a default of 3, validation annotations, tests, appsettings entry, and CONFIGURATION_REFERENCE documentation -- but it does absolutely nothing at runtime. The code explicitly says so:

> SQLite does not support `EnableRetryOnFailure`; this setting will take effect when the project migrates to PostgreSQL...

**Why this matters**: An operator who sets `MaxRetryCount: 5` in appsettings.json will believe their database operations have retry resilience. They don't. This is a false sense of security. The setting is validated (fail-fast on bad values), documented, and tested -- all signals that suggest it is functional.

**Recommendation**: Either (a) remove `MaxRetryCount` entirely and add it when PostgreSQL support actually ships, or (b) at minimum, log a warning at startup if `MaxRetryCount > 0` saying "MaxRetryCount is configured but has no effect with the current SQLite provider." Option (a) is preferred -- YAGNI. Dead configuration is a liability, not preparation.

### Observation 2 (Low): Dual binding pattern is consistent but fragile

`DependencyInjection.cs` reads `DatabaseSettings` manually via `configuration.GetSection("Database").Get<DatabaseSettings>()`, bypassing the `IOptions<T>` pipeline. This is the same pattern used by `CacheSettings` in the same file. The `ValidateOnStart()` registration in `OptionsValidationRegistration.cs` catches invalid config before any DbContext is resolved, so the ordering is safe. But the two independent bindings could silently diverge if someone adds an `IConfigureOptions<DatabaseSettings>` hook later (it would only apply to the `IOptions<T>` path, not the manual read). This is an existing codebase pattern, not something introduced by this PR, so no action needed here.

### Observation 3 (Low): 30s timeout for local-first SQLite

The 30s default matches EF Core's built-in default, which is reasonable. For a local-first SQLite app where queries should complete in milliseconds, 30s is generous but not harmful -- it protects against pathological cases (locked DB file, large migrations) without causing false timeouts. Acceptable.

### Observation 4 (Info): Migration timeout

`Database.Migrate()` is called in `Program.cs` and `PipelineConfiguration.cs`. These will be subject to the configured `CommandTimeoutSeconds`. For SQLite with the current schema, 30s is more than sufficient. If someone lowers the timeout to 1s for aggressive protection, they could inadvertently break migrations. The CONFIGURATION_REFERENCE documentation does not mention this interaction. Consider adding a note.

### Observation 5 (Info): No deploy/.env.example update

The `Database` section has sensible defaults and is optional, so omitting it from `.env.example` is defensible. The CONFIGURATION_REFERENCE.md update is sufficient.

### What looks good
- Layer placement: `DatabaseSettings` in Application layer, consumed by Infrastructure -- respects clean architecture boundaries.
- Consistent with `CacheSettings` dual-binding pattern.
- Test coverage is thorough (default validation, boundary values for both properties).
- Integration test confirms app starts with new config section.
- CONFIGURATION_REFERENCE.md updated with table of contents entry.
- SQLite limitation called out in code comments, XML docs, and reference docs.

### Verdict
**Approve with one requested change**: Remove `MaxRetryCount` (or add a startup log warning that it is a no-op with SQLite). The rest is solid.

Chris0Jeky:
@copilot resolve the merge conflicts in this pull request

PR #967, opened 2026-04-23
Add audit trail retention policy with configurable cleanup

### LINKED ISSUE #956: OPS-31: Audit trail retention policy with configurable cleanup
## Context

The 2026-04-16 audit (docs/AUDIT.md, Security section) flagged: "Audit trail retention unbounded — grows indefinitely." This is a production concern as the AuditLog table will grow without bounds.

## Acceptance Criteria

- [ ] Add \`AuditRetentionSettings\` with configurable \`MaxRetentionDays\` (default: 90) and \`CleanupBatchSize\` (default: 1000)
- [ ] Add settings validation via existing \`ValidateOnStart\` pattern
- [ ] Add \`AuditRetentionWorker\` as a \`BackgroundService\` that runs daily cleanup
- [ ] Worker deletes audit entries older than \`MaxRetentionDays\` in batches
- [ ] Worker logs summary of deleted entries
- [ ] Add unit/integration tests for the cleanup logic
- [ ] Document the configuration in \`docs/platform/CONFIGURATION_REFERENCE.md\`
- [ ] Full backend solution builds and all tests pass

## Suggested Branch Name

\`feat/ops-31-audit-retention-policy\`

## Priority

Priority III — operational hardening

Chris0Jeky:
## Summary

Implements configurable audit log retention policy with automatic cleanup (#956 / OPS-31).

- **AuditRetentionSettings** in Application layer with data-annotation validation: `MaxRetentionDays` (default 90), `CleanupBatchSize` (default 1000), `CleanupIntervalHours` (default 24)
- **DeleteOldEntriesAsync** on `IAuditLogRepository` — uses parameterized raw SQL DELETE with batch-size limits (supports both SQLite LIMIT and SQL Server TOP)
- **AuditRetentionWorker** background service following existing worker patterns (heartbeat, cancellation, telemetry, redacted error logging)
- Registered in `WorkerRegistration` and `OptionsValidationRegistration` with `ValidateOnStart`
- Default config section added to `appsettings.json`

## Test plan

- [x] Settings validation tests (defaults pass, boundary values, out-of-range rejection for all 3 properties)
- [x] Worker unit tests (deletion + logging, debug log when nothing deleted, cutoff date calculation, cancellation propagation)
- [x] Repository integration tests against real SQLite (old entry deletion, recent entry preservation, batch size handling, zero-result case, graceful empty completion)
- [x] Full test suite passes (5,025 tests, 0 failures)
- [x] `dotnet build -c Release` — 0 errors

Closes #956

### DIFF
--- a/backend/src/Taskdeck.Api/Extensions/OptionsValidationRegistration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/OptionsValidationRegistration.cs
@@ -40,6 +40,7 @@ public static class OptionsValidationRegistration
         // ── Settings from WorkerRegistration ────────────────────────────────
 
         services.RegisterValidatedOptions<WorkerSettings>(configuration, "Workers");
+        services.RegisterValidatedOptions<AuditRetentionSettings>(configuration, "AuditRetention");
 
         // ── Settings from CorsRegistration (Cache is used in infrastructure) ─
 
--- a/backend/src/Taskdeck.Api/Extensions/WorkerRegistration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/WorkerRegistration.cs
@@ -39,10 +39,14 @@ public static class WorkerRegistration
             };
         });
 
+        var auditRetentionSettings = configuration.GetSection("AuditRetention").Get<AuditRetentionSettings>() ?? new AuditRetentionSettings();
+        services.AddSingleton(auditRetentionSettings);
+
         services.AddSingleton<WorkerHeartbeatRegistry>();
         services.AddHostedService<LlmQueueToProposalWorker>();
         services.AddHostedService<ProposalHousekeepingWorker>();
         services.AddHostedService<OutboundWebhookDeliveryWorker>();
+        services.AddHostedService<AuditRetentionWorker>();
 
         return services;
     }
--- a/backend/src/Taskdeck.Api/Workers/AuditRetentionWorker.cs
+++ b/backend/src/Taskdeck.Api/Workers/AuditRetentionWorker.cs
@@ -1,4 +1,3 @@
-using Microsoft.Extensions.Options;
 using Taskdeck.Application.Interfaces;
 using Taskdeck.Application.Services;
 using Taskdeck.Api.Telemetry;
--- a/backend/src/Taskdeck.Api/Workers/AuditRetentionWorker.cs
+++ b/backend/src/Taskdeck.Api/Workers/AuditRetentionWorker.cs
@@ -1,0 +1,97 @@
+using Microsoft.Extensions.Options;
+using Taskdeck.Application.Interfaces;
+using Taskdeck.Application.Services;
+using Taskdeck.Api.Telemetry;
+
+namespace Taskdeck.Api.Workers;
+
+/// <summary>
+/// Background worker that periodically deletes audit log entries
+/// older than the configured retention period.
+/// </summary>
+public class AuditRetentionWorker : BackgroundService
+{
+    private readonly IServiceScopeFactory _scopeFactory;
+    private readonly AuditRetentionSettings _settings;
+    private readonly WorkerHeartbeatRegistry _workerHeartbeatRegistry;
+    private readonly ILogger<AuditRetentionWorker> _logger;
+
+    public AuditRetentionWorker(
+        IServiceScopeFactory scopeFactory,
+        AuditRetentionSettings settings,
+        WorkerHeartbeatRegistry workerHeartbeatRegistry,
+        ILogger<AuditRetentionWorker> logger)
+    {
+        _scopeFactory = scopeFactory;
+        _settings = settings;
+        _workerHeartbeatRegistry = workerHeartbeatRegistry;
+        _logger = logger;
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        _logger.LogInformation(
+            "AuditRetentionWorker starting (retention={RetentionDays}d, batch={BatchSize}, interval={IntervalHours}h)",
+            _settings.MaxRetentionDays,
+            _settings.CleanupBatchSize,
+            _settings.CleanupIntervalHours);
+
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            _workerHeartbeatRegistry.ReportHeartbeat(nameof(AuditRetentionWorker));
+
+            try
+            {
+                await CleanupOldEntriesAsync(stoppingToken);
+            }
+            catch (OperationCanceledException) when (stoppingToken.IsCancellationRequested)
+            {
+                break;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(
+                    "Error in AuditRetentionWorker iteration. {ExceptionSummary}",
+                    SensitiveDataRedactor.SummarizeException(ex));
+            }
+
+            await Task.Delay(TimeSpan.FromHours(_settings.CleanupIntervalHours), stoppingToken);
+        }
+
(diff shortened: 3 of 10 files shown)


Chris0Jeky:
## Adversarial Self-Review

### 1. SQL Deletion Safety (parameterization / injection)

**SAFE.** The `DeleteOldEntriesAsync` method uses `ExecuteSqlRawAsync` with `{0}` and `{1}` positional placeholders and passes values via the `object[]` params array. EF Core translates these into proper parameterized queries (e.g., `@p0`, `@p1` in the generated SQL). No string interpolation or concatenation of user input into the SQL string. The `olderThan` and `batchSize` values originate from validated settings (data annotations enforce ranges) and the worker's own `DateTimeOffset.UtcNow` calculation, so even the inputs are safe.

### 2. Could batch deletion block other queries too long?

**Low risk with default settings.** The default batch size is 1,000 rows. SQLite uses database-level locking (not row-level), so each DELETE statement will hold a write lock for the duration of that batch. With 1,000 rows, this should complete in milliseconds. The batch loop breaks as soon as fewer rows are deleted than the batch size, so it does not hold locks in a tight loop. Between batches, other writers can acquire the lock. For very large backlogs (e.g., millions of old entries), operators can reduce `CleanupBatchSize` further. The 24-hour default interval means this only runs once per day. **Acceptable for production.**

### 3. Cancellation handling

**Properly handled.** The worker's `ExecuteAsync` loop catches `OperationCanceledException` when `stoppingToken.IsCancellationRequested` and breaks cleanly. The `CleanupOldEntriesAsync` method passes the cancellation token through to `ExecuteSqlRawAsync`. The batch loop also checks `cancellationToken.IsCancellationRequested` at the top of each iteration, so mid-cleanup cancellation will exit after the current batch completes. No resources are leaked -- `using var scope` ensures cleanup.

### 4. Default values for production

**Sensible.**
- `MaxRetentionDays = 90`: 3 months is a standard audit retention default. Allows compliance reviews while keeping DB size manageable.
- `CleanupBatchSize = 1000`: Conservative enough to avoid lock contention on SQLite. Large enough to be efficient for typical audit volumes.
- `CleanupIntervalHours = 24`: Daily cleanup is appropriate -- audit entries accumulate slowly and don't need real-time pruning.

### 5. Unused import

The `AuditRetentionWorker.cs` has `using Microsoft.Extensions.Options;` which is not used (the settings are injected as a direct singleton, not via `IOptions<T>`). This is cosmetic only, not functional.

### Verdict

No blocking issues found. The implementation follows existing worker patterns, the SQL is safe, and the defaults are production-appropriate. The unused import is cosmetic.

Chris0Jeky:
Fixed the unused `using Microsoft.Extensions.Options;` import found during self-review (commit fd4c9c20).

Chris0Jeky:
## Adversarial Code Review

### CRITICAL: DateTimeOffset SQLite string comparison is unreliable

**File**: `AuditLogRepository.cs` `DeleteOldEntriesAsync`

The method passes a raw `DateTimeOffset` object to `ExecuteSqlRawAsync`:

```csharp
var deleted = await _context.Database.ExecuteSqlRawAsync(
    sql,
    new object[] { olderThan, batchSize },
    cancellationToken);
```

SQLite has no native DateTimeOffset type. EF Core stores DateTimeOffset as a string in the format `"yyyy-MM-dd HH:mm:ss.fffffff+HH:mm"`. The `Timestamp < {0}` comparison in SQLite is therefore a **string comparison**. If the parameterized DateTimeOffset value is formatted differently than the stored values (e.g., different precision, different offset format), the comparison will produce wrong results -- either deleting too many rows or too few.

This same codebase already handles this correctly in `OAuthAuthCodeRepository`:

```csharp
var cutoffStr = cutoff.ToString("yyyy-MM-dd HH:mm:ss.fffffff+00:00");
```

The fix: format the DateTimeOffset as a string matching EF Core's SQLite storage format before passing it as a parameter, just like `OAuthAuthCodeRepository.DeleteExpiredAsync` does.

**Risk**: This could cause the worker to **silently delete zero rows on every run** (no-op retention policy) or, worse, delete rows it shouldn't based on string ordering.

**Note**: The integration tests happen to pass because both the UPDATE (to set old timestamps) and the DELETE use the same `ExecuteSqlRawAsync` path with raw `DateTimeOffset`, so they have matching format inconsistencies. This means the tests do not catch the real-world scenario where timestamps were written by EF Core's LINQ pipeline (which formats them as strings).

---

### CRITICAL: SQL Server `DELETE TOP(@p)` will fail at runtime

**File**: `AuditLogRepository.cs` `DeleteOldEntriesAsync`

```csharp
: "DELETE TOP({1}) FROM AuditLogs WHERE Timestamp < {0}";
```

When EF Core converts `{1}` to a parameter `@p1`, this generates:

```sql
DELETE TOP(@p1) FROM AuditLogs WHERE Timestamp < @p0
```

SQL Server does **not** allow a parameterized value directly inside `TOP()` without a subquery or variable. The correct pattern is to use a subquery or CTE as Gemini suggested:

```sql
WITH CTE AS (SELECT TOP(@p1) Id FROM AuditLogs WHERE Timestamp < @p0 ORDER BY Timestamp ASC)
DELETE FROM CTE
```

Since this is a SQLite-first project and SQL Server support is future-facing, the severity is lower in practice. But if anyone deploys against SQL Server, this will throw a runtime exception.

---

### MEDIUM: Heartbeat starvation during large batch cleanup (agrees with Gemini)

The batch loop runs entirely inside `DeleteOldEntriesAsync` in the repository layer. If there are millions of old entries (e.g., first deployment of retention on an existing database), the loop could run for minutes without reporting a heartbeat. The heartbeat is only reported once per `ExecuteAsync` iteration (every 24 hours by default, but also before each cleanup call).

The fix: either move the batch loop to the worker level (so heartbeats can be reported between batches), or add a batch iteration cap with a "will continue next cycle" semantic.

---

### LOW: Missing ORDER BY on batch deletion (agrees with Gemini)

Without `ORDER BY Timestamp ASC` in the subquery, SQLite may delete rows in non-deterministic order. This is cosmetically undesirable and can lead to suboptimal index utilization, though it does not affect correctness (all matching rows will eventually be deleted).

---

### OBSERVATIONS (no fix required)

- **SQL injection safety**: Confirmed safe. `ExecuteSqlRawAsync` with `{0}`/`{1}` positional placeholders creates proper parameterized queries. No string interpolation or concatenation.
- **Batch loop termination**: Correct. The loop breaks when `deleted < batchSize`, and also checks `cancellationToken.IsCancellationRequested` each iteration. Cannot infinite-loop.
- **BackgroundService lifecycle**: Properly handled. `OperationCanceledException` is caught when `stoppingToken.IsCancellationRequested`. The `Task.Delay` also respects the token, so shutdown is clean.
- **Settings validation**: Range annotations are correct. `MaxRetentionDays` minimum is 1 (not 0), preventing accidental delete-everything. Validation tests cover all boundaries.
- **Clock handling**: `DateTimeOffset.UtcNow` is used consistently. No timezone confusion risk.
- **Concurrency**: SQLite uses database-level locking. Each batch DELETE acquires and releases the write lock, allowing interleaved reads. Acceptable for default batch size of 1000.
- **Logging**: Appropriate. Information-level for deletes, Debug for no-ops. No log spam risk (runs once per 24h default).
- **DI registration**: Double-registered (singleton instance in `WorkerRegistration` + `IOptions<T>` in `OptionsValidationRegistration`) -- this is consistent with other settings in this codebase, so it's fine.
- **Test coverage**: Good. Both unit tests (with fakes) and integration tests (against real SQLite) cover the main paths. The cancellation test and boundary tests are solid.

Issue #996, opened 2026-04-25
PAPER-00 · Master tracker · Paper & Graphite, Ember Edition overhaul


Chris0Jeky:
# PAPER-00 · Master Tracker · Paper & Graphite, Ember Edition Overhaul

## Thesis

Adopt the **Paper & Graphite, Ember Edition** visual + interaction system across the Taskdeck frontend. The system encodes the product's core thesis — capture → review → apply with no silent mutations — directly in materials (cream paper, hairline rules, italic serif, single seal-red ember accent) rather than helper labels. Trust signals are physical: applied stamps emboss, undo windows literally erase as the 6-hour grace closes.

Source of truth: `design_handoff_taskdeck_paper/` checked into the repo (README + JSX prototypes + `paper/tokens.css`).

## Strategic intent

- Replace the current Obsidian-toned shell across all primary surfaces with Paper-Light as canonical, Paper-at-Night as the dark companion (strict mirror of metaphor).
- Make Review the centerpiece: implement the *deep* version with provenance, side-effects, conflicts, history, confidence dial, undo timeline, similar-past decisions.
- Replace every `loading…` / spinner state in LLM flows with **Ink Bleed** — a 5-phase, 4.6s motion that dries to a stamp.
- Keep behavior unchanged: this is presentation + interaction polish, not a backend rewrite. All capture / proposal / approval semantics remain the same.

## Non-goals

- No backend changes beyond what surface needs require (e.g., new fields the dossier needs).
- No new product features beyond what the design implies (e.g., "Seal day", "A line for tomorrow", undo timer).
- No removal of the existing `--td-*` token system in this overhaul. Paper sits parallel under `.paper` / `.paper-night` scope until migration is ready to flip canonical.
- Multiplayer/CRDT explicitly out of scope.

### Phase 1 · Foundation (must land first, blocks everything else)
- **PAPER-01** Tokens + theme classes + fonts + paper substrate
- **PAPER-02** Component primitives (Stamp, HLBtn, Tagstamp, Card, kbd, hairline icons)

### Phase 2 · Shell
- **PAPER-03** Sidebar + TopBar in Paper

### Phase 3 · Surfaces (parallelizable after Phase 1+2)
- **PAPER-04** Home / morning reset
- **PAPER-05** Board / kanban with index card variant
- **PAPER-06** Review *deep* — central surface
- **PAPER-07** Inbox / capture (composer + nib)
- **PAPER-08** Today / end-of-day dossier
[...]

PR #1029, opened 2026-04-25
PAPER-11 · Narrow companions — foundation + composable

### LINKED ISSUE #1007: PAPER-11 · Narrow companions (375 phone, 768 tablet)
# PAPER-11 · Narrow companions — 375 phone + 768 tablet

Part of the Paper overhaul (master tracker: PAPER-00). **Blocked by:** PAPER-03 + at least one surface (Board or Review).

## Goal
Match the narrow specifications so Paper survives at phone (375) and tablet (768) widths.

### Phone (≤480px, target 375)
- Single-column.
- Sidebar collapses to bottom tab bar with 4 letterform glyphs (H · T · R · I).
- Stamps shrink to 48px.
- Type scales down 10% via `--paper-scale: .9`.

### Tablet (≤1024px, target 768)
- Sidebar collapses to icon-only rail (60px).
- Boards reduce to 2 visible columns with horizontal scroll, snap points.

## Implementation
- Narrow-mode CSS in `paper-tokens.css` under media queries.
- `PaperSidebar.vue` reads viewport and renders `rail` / `bottombar` variant.
- `PaperBoardView.vue` adds horizontal snap scroll at tablet width.
- Type scale via `--paper-scale` consumed by `.tk-h1/h2/h3/lede/body`.

## Tests
- Playwright: `page.setViewportSize` 375/768; non-empty + no console errors.
- vitest: PaperSidebar renders bottom-bar variant when matchMedia simulates phone.

## Adversarial review
- [ ] No fixed widths leak from desktop CSS.
- [ ] Bottom-bar respects iOS safe area inset.
- [ ] Type scale doesn't cascade twice (no compounding 0.9 × 0.9).
- [ ] Tablet rail still respects active-route ember accent.

Chris0Jeky:
## Summary

Foundation slice of issue #1007 (master tracker #996). Lands the responsive media-query rules + reactive viewport composable; per-surface narrow tweaks land in follow-ups once the surface branches merge.

- \`paper-tokens.css\` adds phone (≤480px) and tablet (≤1024px) breakpoints scoped under \`.paper\` / \`.paper-night\`. Phone sets \`--paper-scale: 0.9\` (consumed by \`.tk-*\` utilities via the existing \`calc(... * var(--paper-scale, 1))\` math, no compounding) and shrinks stamps to 48px. Tablet flips \`--paper-density\` to \`compact\`. Introduces \`--paper-safe-bottom: env(safe-area-inset-bottom, 0px)\` for the phone bottom-bar variant.
- \`useViewportMode\` composable returns a reactive \`'desktop' | 'tablet' | 'phone'\` ref driven by \`matchMedia\` listeners; teardown on unmount via \`onBeforeUnmount\`.
- 5 vitest cases cover detection, live media changes, and listener cleanup.

## Out of scope (filed as follow-ups against #1007)

The README spec also calls for: sidebar bottom-bar variant on phone (4 letterform glyphs H · T · R · I), icon-only 60px rail on tablet, and board horizontal-snap scroll. These need to land as patches on the respective surface PRs (#1012, #1014, etc.). Filing per-surface follow-up issues to track.

## Test plan

- [x] \`npm run typecheck\` clean
- [x] \`npm run lint\` no new warnings
- [x] vitest 5 / 5 passing
- [ ] Manual smoke at 375 / 768 / desktop after surface PRs merge

Closes #1007 (foundation scope only — see follow-ups list).

### DIFF
--- a/frontend/taskdeck-web/src/composables/useViewportMode.ts
+++ b/frontend/taskdeck-web/src/composables/useViewportMode.ts
@@ -1,23 +1,24 @@
-import { onBeforeUnmount, readonly, ref, type Ref } from 'vue'
+import { onScopeDispose, readonly, ref, type Ref } from 'vue'
 
 export type ViewportMode = 'desktop' | 'tablet' | 'phone'
 
 const PHONE_QUERY = '(max-width: 480px)'
 const TABLET_QUERY = '(max-width: 1024px)'
 
-function detect(): ViewportMode {
-  if (typeof window === 'undefined' || !window.matchMedia) return 'desktop'
-  if (window.matchMedia(PHONE_QUERY).matches) return 'phone'
-  if (window.matchMedia(TABLET_QUERY).matches) return 'tablet'
+function classify(phoneMatches: boolean, tabletMatches: boolean): ViewportMode {
+  if (phoneMatches) return 'phone'
+  if (tabletMatches) return 'tablet'
   return 'desktop'
 }
 
 /**
  * Reactive viewport mode for Paper narrow companions.
- * Listens to matchMedia changes and tears down on unmount.
+ * Listens to matchMedia changes and tears down via onScopeDispose so the
+ * composable can be used inside components, Pinia stores, or any other
+ * reactive effect scope.
  */
 export function useViewportMode(): { mode: Readonly<Ref<ViewportMode>> } {
-  const mode = ref<ViewportMode>(detect())
+  const mode = ref<ViewportMode>('desktop')
 
   if (typeof window === 'undefined' || !window.matchMedia) {
     return { mode: readonly(mode) }
@@ -27,13 +28,16 @@ export function useViewportMode(): { mode: Readonly<Ref<ViewportMode>> } {
   const tabletMq = window.matchMedia(TABLET_QUERY)
 
   const update = () => {
-    mode.value = detect()
+    mode.value = classify(phoneMq.matches, tabletMq.matches)
   }
+
+  // Initialise from the cached MQL state — no extra matchMedia() calls.
+  update()
 
   phoneMq.addEventListener?.('change', update)
   tabletMq.addEventListener?.('change', update)
 
-  onBeforeUnmount(() => {
+  onScopeDispose(() => {
     phoneMq.removeEventListener?.('change', update)
     tabletMq.removeEventListener?.('change', update)
   })
--- a/frontend/taskdeck-web/src/composables/useViewportMode.ts
+++ b/frontend/taskdeck-web/src/composables/useViewportMode.ts
@@ -1,0 +1,42 @@
+import { onBeforeUnmount, readonly, ref, type Ref } from 'vue'
+
+export type ViewportMode = 'desktop' | 'tablet' | 'phone'
+
+const PHONE_QUERY = '(max-width: 480px)'
+const TABLET_QUERY = '(max-width: 1024px)'
+
+function detect(): ViewportMode {
+  if (typeof window === 'undefined' || !window.matchMedia) return 'desktop'
+  if (window.matchMedia(PHONE_QUERY).matches) return 'phone'
+  if (window.matchMedia(TABLET_QUERY).matches) return 'tablet'
+  return 'desktop'
+}
+
+/**
+ * Reactive viewport mode for Paper narrow companions.
+ * Listens to matchMedia changes and tears down on unmount.
+ */
+export function useViewportMode(): { mode: Readonly<Ref<ViewportMode>> } {
+  const mode = ref<ViewportMode>(detect())
+
+  if (typeof window === 'undefined' || !window.matchMedia) {
+    return { mode: readonly(mode) }
+  }
+
+  const phoneMq = window.matchMedia(PHONE_QUERY)
+  const tabletMq = window.matchMedia(TABLET_QUERY)
+
+  const update = () => {
+    mode.value = detect()
+  }
+
+  phoneMq.addEventListener?.('change', update)
+  tabletMq.addEventListener?.('change', update)
+
+  onBeforeUnmount(() => {
+    phoneMq.removeEventListener?.('change', update)
+    tabletMq.removeEventListener?.('change', update)
+  })
+
+  return { mode: readonly(mode) }
+}
--- a/frontend/taskdeck-web/src/paper-tokens.css
+++ b/frontend/taskdeck-web/src/paper-tokens.css
@@ -432,9 +432,18 @@
 /* =========================================================
    Narrow companions — PAPER-11
    Phone (≤480px) and tablet (≤1024px) responsive rules.
-   The `--paper-scale` token (default 1) is consumed by .tk-* utilities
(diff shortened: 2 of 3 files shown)


Chris0Jeky:
# Adversarial review (PAPER-11)

Reviewed `paper/11-narrow` against issue #1007 checklist, the spec at `design_handoff_taskdeck_paper/README.md` §11 + `paper/surface-narrow.jsx`, and existing CI feedback (gemini-code-assist).

## Scope verification

- Foundation slice only: media-query rules + reactive composable. Per-surface tweaks (phone bottom-bar, tablet 60px rail, board horizontal-snap) are explicitly deferred to follow-up patches on the relevant surface PRs (#1012, #1014, etc.). Confirmed `PaperSidebar` is untouched in this PR, so its existing ≤640px off-canvas drawer behavior is unaffected and no tablet-rail regression is possible.
- All four required CI jobs (`Frontend Unit`, `E2E Smoke`, `Docs Governance`, `Backend Unit`) are green.

## Findings

### F1 · `onBeforeUnmount` should be `onScopeDispose` — confidence high
**File:** `frontend/taskdeck-web/src/composables/useViewportMode.ts:1,36`
The composable currently uses `onBeforeUnmount`, which only fires inside a component's `setup`. `onScopeDispose` is the canonical Vue 3 idiom for composables — it works inside any reactive `EffectScope` (Pinia stores, custom scopes, async setups) and still fires on component unmount. Already flagged by gemini-code-assist (medium). Will fix.

### F2 · `update()` re-creates MQL objects on every resize — confidence high
**File:** `frontend/taskdeck-web/src/composables/useViewportMode.ts:29-31` (calls `detect()` at lines 8-13)
The `change` listener calls `detect()`, which invokes `window.matchMedia(query)` twice per event. Since `phoneMq` and `tabletMq` are already in closure scope, the listener should read their `.matches` directly. Avoids redundant browser API calls and makes the SSR-safe early-return in `detect()` redundant for the post-mount hot path. Already flagged by gemini-code-assist (medium). Will fix.

### F3 · `--paper-density` and `--paper-safe-bottom` introduced but unconsumed — confidence med (~80)
**File:** `frontend/taskdeck-web/src/paper-tokens.css:439-448`
Both tokens are defined but nothing in the stylesheet (or in the rest of the frontend) reads them yet. The PR body acknowledges this as foundation-only, but the in-file CSS comment block currently only documents `--paper-scale`. A forward-hook note next to the new tokens would prevent a future contributor from assuming they are dead code and removing them. Will add a comment line.

### F4 · No regression test for multi-cycle mount/unmount listener cleanup — confidence med (~80)
**File:** `frontend/taskdeck-web/src/tests/composables/useViewportMode.spec.ts`
The current "cleans up listeners on unmount" case mounts and unmounts once. A multi-cycle test (mount → unmount → mount → unmount, asserting listener count returns to zero each cycle and is exactly 1 between mounts) is cheap insurance against future refactors that might leak listeners across remounts (e.g. hoisting the MQL into module scope). Will add.

### F5 · `detect()` initial call creates two MQLs — confidence low (<75)
**File:** `frontend/taskdeck-web/src/composables/useViewportMode.ts:8-13`
On first call `detect()` invokes `window.matchMedia` for both queries (up to twice on phones, once on tablets, twice on desktop). Browsers cache MQLs internally so this is effectively free, and it only happens once per `useViewportMode()` call. Deferred — fixing it would tangle with F2 and gain nothing measurable.

## Spec-vs-implementation cross-checks (no findings)

- **Fixed widths leaking at 375px:** grep across `paper-tokens.css` shows the only fixed widths are `.stamp` (overridden in the ≤480 media query to 48px), `.pkbd min-width: 18px`, `.hl-icon` 14/16/20px, `.pstatus::before` 6px dot. None breaks layout at 375px. ✓
- **Type scale double-multiply:** `--paper-scale` is set once on `.paper` / `.paper-night`. Each leaf utility (`.tk-display`, `.tk-h1`, `.tk-body`, …) sets `font-size: calc(var(--t-X) * var(--paper-scale, 1))` using absolute `px` calcs — no `em` cascade, no compounding. Children that don't set their own `font-size` simply inherit the already-scaled root value. Confirmed clean. ✓
- **`env(safe-area-inset-bottom, 0px)` fallback:** present and correct. ✓
- **SSR / no-`matchMedia` safety:** `useViewportMode` returns `'desktop'` and skips listener wiring when `typeof window === 'undefined' || !window.matchMedia`. Verified. ✓
- **Test mock advances state and asserts string values:** the live-changes test asserts `'desktop' → 'tablet' → 'phone'` after `mq.set(...)` + `nextTick()`. Genuine assertions. ✓
- **PaperSidebar tablet variant regression:** PaperSidebar is unchanged in this PR; its existing ≤640px drawer rules continue to apply. The spec's tablet 60px icon rail is correctly deferred to a per-surface follow-up. ✓

## Bot comments status

- gemini-code-assist (3 inline comments on `useViewportMode.ts`): F1 + F2 cover both substantive items. The third inline (`onScopeDispose` inside the cleanup body) is the same change as F1.

Pushing fixes for F1–F4 next; will follow up with a "Adversarial review fixes" comment listing commit hashes.

Chris0Jeky:
# Adversarial review fixes

Pushed three commits to `paper/11-narrow` addressing findings F1–F4 from [the review above](https://github.com/Chris0Jeky/Taskdeck/pull/1029#issuecomment-4320133919). F5 deferred per its own confidence note.

| Finding | Commit | What changed |
|---|---|---|
| F1 + F2 | `0574b200` | `useViewportMode` now uses `onScopeDispose` for teardown (works inside any reactive `EffectScope`, not just component setup) and the `change` listener reads `phoneMq.matches` / `tabletMq.matches` from the closed-over `MediaQueryList` handles via a small `classify` helper instead of re-calling `window.matchMedia()` on every event. Initial value is also derived from those handles, so the standalone `detect()` helper is gone. |
| F3 | `f4cbc810` | Expanded the comment block at the head of the PAPER-11 narrow-companion section in `paper-tokens.css` to explicitly document `--paper-safe-bottom` and `--paper-density` as forward hooks for the per-surface follow-up patches against #1007. No behaviour change. |
| F4 | `c183a86b` | Added two cases to `useViewportMode.spec.ts`: (a) three-cycle mount/unmount that asserts listener count returns to zero each cycle and is exactly 1 between mounts; (b) a no-`matchMedia` case that verifies the composable falls back to `'desktop'` without throwing. Also promoted the mocked MQL `.matches` to a live getter so the mock now mirrors real browser semantics — the previous snapshot-style mock only worked because the old implementation re-queried `matchMedia` on each event. |

## Verification

- `npm run typecheck` — clean.
- `npm run lint` — 6 warnings, all pre-existing in `AgentRunDetailView.vue` / `AgentRunsView.vue` / `AgentsView.vue`; no new warnings from this PR.
- `npx vitest --run src/tests/composables/useViewportMode.spec.ts` — 7 / 7 passing (5 original + 2 new).

## Bot-comment status

The three gemini-code-assist inline comments on `useViewportMode.ts` are all addressed by F1 + F2 in commit `0574b200`. No deferred bot findings.

Issue #1004, opened 2026-04-25
PAPER-08 · Today / End-of-day dossier in Paper


Chris0Jeky:
# PAPER-08 · Today / End-of-day dossier

Part of the Paper overhaul (master tracker: PAPER-00). **Blocked by:** PAPER-01, PAPER-02, PAPER-03.

## Goal
Replace `TodayView` with the Paper end-of-day dossier — the closing read of the workday.

## Spec (paper/surface-today.jsx + README §10)

1. **Cover** — paper-2→paper gradient, 56px serif italic headline ("Today, you moved nine cards."), tk-lede paragraph, "Seal day" CTA (ember), dossier serial top-right.
2. **5 stat cards** — cards moved · proposals applied · captures triaged · longest focus · overdue. 38px serif italic number, mono eyebrow, sub-line, 2px top accent in tone.
3. **§I Cadence** — 24-hour activity strip (24 bars, peak hour ember) + first/peak/last mini-stats.
4. **§II Ledger** — full event log of the day in `.rule-ledger`: serial `L-NNN`, time, who pill, what, dot tone.
5. **§III Decisions** — 4-up grid of today's proposals with verdict tagstamps (APPLIED green, REJECTED rust, DEFERRED ember).
6. **§IV Boards touched** — list of boards with move counts and proposal pills. Untouched boards dimmed.
7. **§V Carry-over** — overdue cards in rust-bordered card. "Pin to tomorrow's morning" CTA.
8. **§VI Streak** — 90-day grid, 30 cols, ember intensity per day, today highlighted.
9. **§VII A line for tomorrow** — italic serif textarea, autosaved.

## Backend gaps (file as sub-issues)
- Daily cadence aggregation per-hour buckets.
- Streak query (last 90 days, sealed flag per day).
- "Seal day" snapshot action.
- "A line for tomorrow" per-day storage.

## Implementation
- Create `views/paper/PaperTodayView.vue` + section subcomponents.
- Cadence and streak as plain SVG (no chart library).
- "Seal day" CTA dispatches `todayStore.sealDay()`.

## Tests
- vitest: stat number formatting; streak grid 90 cells.
- vitest: "A line for tomorrow" autosaves with debounce.
- Playwright: end-of-day flow smoke.

## Adversarial review
- [ ] Cadence chart respects `prefers-reduced-motion`.
- [ ] "Seal day" idempotent — second click while sealed is a no-op.
- [ ] Streak intensity bucketed deterministically.
[...]

PR #1031, opened 2026-04-25
Add cadence aggregation backend for Today view (#1015)

### LINKED ISSUE #1015: paper-today-backend-gap-cadence-aggregation
Required by PAPER-08 (#1004). Provide a per-hour activity bucket aggregation for the current user/workspace/day to drive the Today Cadence strip (24 SVG bars + first/peak/last action). Should include event counts per hour and timestamps for first/peak/last actions. Currently stubbed via useTodayDossier with mock data.

Chris0Jeky:
## Summary

Implements the backend cadence aggregation endpoint required by PAPER-08 (#1004) for the Today Cadence strip.

- **Domain**: `CadenceBucket` and `CadenceSnapshot` value objects in `Taskdeck.Domain.Common` with hour validation (0-23) and an `Empty()` factory for zero-activity days
- **Application**: `ICadenceService` interface + `CadenceService` implementation that queries `IAuditLogRepository` for a user's audit entries on a given UTC day, buckets them by hour, and computes first/peak/last action timestamps. `CadenceResponse` DTO for API serialization
- **API**: `TodayController` with `GET /api/today/cadence?date=2026-04-25` endpoint, following `AuthenticatedControllerBase` + `Result<T>` patterns
- **DI**: Wired in `ApplicationServiceRegistration.cs` as scoped, injecting `IAuditLogRepository` from `IUnitOfWork`
- **Tests**: 13 domain tests (validation, equality, empty, boundary) + 10 application tests (validation, empty day, single event, multi-hour aggregation, peak ties, midnight boundary, date normalization, all-24-hours)

No new database migration needed -- reuses existing `IAuditLogRepository.QueryAsync` which already supports user+date-range filtering.

Closes #1015

## Test plan

- [x] `dotnet build backend/Taskdeck.sln -c Release` -- 0 errors
- [x] `dotnet test --filter "FullyQualifiedName~Cadence"` -- 23 tests pass (13 domain + 10 application)
- [x] Architecture tests pass (layer boundary validation)
- [ ] CI green on PR

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/TodayController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/TodayController.cs
@@ -31,6 +31,7 @@ public class TodayController : AuthenticatedControllerBase
     /// Returns 24 hourly buckets with event counts plus first/peak/last action timestamps.
     /// </summary>
     /// <param name="date">Date to aggregate (ISO 8601). Defaults to today (UTC).</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
     /// <returns>Cadence snapshot for the day.</returns>
     /// <response code="200">Cadence snapshot returned successfully.</response>
     /// <response code="400">Invalid date parameter.</response>
--- a/backend/src/Taskdeck.Api/Controllers/TodayController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/TodayController.cs
@@ -1,0 +1,65 @@
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Taskdeck.Api.Contracts;
+using Taskdeck.Api.Extensions;
+using Taskdeck.Application.DTOs;
+using Taskdeck.Application.Interfaces;
+
+namespace Taskdeck.Api.Controllers;
+
+/// <summary>
+/// Today-view endpoints: cadence aggregation and daily dossier data.
+/// </summary>
+[ApiController]
+[Authorize]
+[Route("api/[controller]")]
+[Produces("application/json")]
+public class TodayController : AuthenticatedControllerBase
+{
+    private readonly ICadenceService _cadenceService;
+
+    public TodayController(
+        ICadenceService cadenceService,
+        IUserContext userContext)
+        : base(userContext)
+    {
+        _cadenceService = cadenceService;
+    }
+
+    /// <summary>
+    /// Get the per-hour cadence snapshot for the authenticated user on the specified date.
+    /// Returns 24 hourly buckets with event counts plus first/peak/last action timestamps.
+    /// </summary>
+    /// <param name="date">Date to aggregate (ISO 8601). Defaults to today (UTC).</param>
+    /// <returns>Cadence snapshot for the day.</returns>
+    /// <response code="200">Cadence snapshot returned successfully.</response>
+    /// <response code="400">Invalid date parameter.</response>
+    /// <response code="401">Authentication required.</response>
+    [HttpGet("cadence")]
+    [ProducesResponseType(typeof(CadenceResponse), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(ApiErrorResponse), StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(typeof(ApiErrorResponse), StatusCodes.Status401Unauthorized)]
+    public async Task<IActionResult> GetCadence(
+        [FromQuery] DateTimeOffset? date,
+        CancellationToken cancellationToken)
+    {
+        if (!TryGetCurrentUserId(out var userId, out var errorResult))
+            return errorResult!;
+
+        var targetDate = date ?? DateTimeOffset.UtcNow;
+
+        var result = await _cadenceService.GetDailyCadenceAsync(userId, targetDate, cancellationToken);
+
+        if (!result.IsSuccess)
+            return result.ToErrorActionResult();
+
+        var snapshot = result.Value;
+        var response = new CadenceResponse(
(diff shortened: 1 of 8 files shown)


Chris0Jeky:
## Adversarial Self-Review

Reviewed all 7 changed files. Findings:

### No issues found (verified correct)

1. **Layer boundaries**: Domain has no infrastructure dependencies. Application depends only on Domain. API depends on Application. Architecture tests pass.
2. **Validation**: `CadenceBucket` validates hour range (0-23) and non-negative event count. `CadenceService` validates non-empty userId. Controller uses `TryGetCurrentUserId` for auth.
3. **Day boundary calculation**: `dayEnd = dayStart.AddDays(1).AddTicks(-1)` is correct for the inclusive `<=` comparison in `QueryAsync`.
4. **Peak tie-breaking**: Loop iterates 0-23 with strict `>` comparison, so ties correctly go to the earliest hour. Tested.
5. **CancellationToken**: Properly threaded from controller through service to repository.
6. **DI registration**: Uses factory lambda to resolve `IAuditLogRepository` from `IUnitOfWork.AuditLogs`, matching the pattern used by `BoardMetricsService` and `ForecastingService`.
7. **Empty day**: Returns `CadenceSnapshot.Empty()` with all 24 zero-count buckets, not an error, so the frontend always gets 24 bars.

### Observations (acceptable, no fix needed)

1. **No workspace scoping**: Issue mentions "workspace/day" but audit log entries are entity-scoped, not workspace-scoped. This is consistent with the existing data model -- the user's audit trail is inherently their workspace activity.
2. **Reflection in tests**: `CreateAuditLog` uses reflection to set `Timestamp` (private setter). This is a common pattern in the codebase for testing time-dependent behavior (the AuditLog constructor hardcodes `DateTimeOffset.UtcNow`).
3. **10,000 entry cap**: The `MaxDailyEntries` limit prevents unbounded memory usage. If a user genuinely exceeds this, the least-recent entries would be excluded from the cadence, which is an acceptable degradation.

### No security concerns

- Endpoint requires `[Authorize]`
- User can only query their own cadence (userId comes from JWT claims via `TryGetCurrentUserId`, not from query params)
- No user-controlled input reaches raw SQL

Chris0Jeky:
## Adversarial Review -- PR #1031 (cadence aggregation)

Reviewed all 8 changed files. Overall the PR is clean and well-structured. Three issues found plus the Gemini bot finding. Fixing all of them in follow-up commits.

### Issues Found

**1. [Medium] `CadenceSnapshot.Empty()` allocates on every call (Gemini bot finding)**
`CadenceSnapshot.cs:38-46` -- `Empty()` creates 24 `CadenceBucket` objects and a new `CadenceSnapshot` every time. Since the empty snapshot is immutable and constant, it should be cached as a `static readonly` field.

**Fix:** Cache as `private static readonly Lazy<CadenceSnapshot>` and return the cached value.

---

**2. [Low] No invariant validation on `CadenceSnapshot.Buckets`**
`CadenceSnapshot.cs:28-33` -- The positional record constructor accepts any `IReadOnlyList<CadenceBucket>` without validating that it contains exactly 24 buckets or that `Buckets` is non-null. A caller could construct a snapshot with 0, 5, or 30 buckets, violating the domain invariant documented in the XML comment ("24 SVG bars").

**Fix:** Add a `Validate()` method called from a custom constructor (or use an `init` block) that enforces `Buckets.Count == 24` and non-null.

---

**3. [Low] Missing XML doc `<param>` tag for `cancellationToken` in `TodayController`**
`TodayController.cs:44` -- Produces compiler warning `CS1573: Parameter 'cancellationToken' has no matching param tag`. The other parameters have doc tags but this one is missing.

**Fix:** Add `<param name="cancellationToken">` tag.

---

### Verified Non-Issues

- **`CadenceBucket` record `with` bypass**: Properties use `{ get; }` (no `init`), so the `with` expression cannot modify values -- constructor validation is not bypassable.
- **`dayEnd = AddTicks(-1)` semantics**: Repository uses `<=` comparison, and `DateTimeOffset` has tick precision, so this correctly covers the full day window.
- **DI wiring via `IUnitOfWork.AuditLogs`**: Matches the established pattern for other services.
- **Clean architecture boundaries**: Domain types in Domain, DTOs in Application, controller in Api. All correct.
- **Test quality**: Good coverage of edge cases (midnight boundary, peak ties, non-UTC normalization, empty day). The reflection-based timestamp setter in tests is a pragmatic choice given the private setter on `AuditLog.Timestamp`.

Chris0Jeky:

### IMPORTANT

**1. Missing `PeakHour` range validation in `CadenceSnapshot` constructor (Confidence: 85)**
`CadenceSnapshot` accepts any `int?` for `peakHour` without checking 0-23. `CadenceBucket` validates hour range, but the snapshot does not. A caller could construct `new CadenceSnapshot(buckets, first, peakHour: 99, last)` and hold an invalid hour value. Currently safe because `CadenceService` only passes valid indices, but domain value objects should be self-validating.

**Fix:** Add `if (peakHour.HasValue && (peakHour.Value < 0 || peakHour.Value > 23)) throw new ArgumentOutOfRangeException(...)` in the constructor.

**2. Silently truncated results at 10,000 entries (Confidence: 82)**
`CadenceService` passes `limit: 10_000` to `QueryAsync`. If exceeded (automation-generated audit entries), the cadence is silently truncated — and since results are ordered `DESC`, the **oldest** entries are dropped, skewing `FirstActionAt` and early-hour buckets.

**Fix:** Add an `isTruncated` flag to the response, or change query order to `ASC`.

### Verified Non-Issues
- Security: `[Authorize]` + `TryGetCurrentUserId` from JWT claims — no cross-user leakage
- Timezone: correctly normalizes to UTC, hour extraction is consistent
- Clean architecture boundaries respected
- CancellationToken properly threaded
- Empty day returns `CadenceSnapshot.Empty()` with all 24 zero-count buckets

Issue #1017, opened 2026-04-25
paper-today-backend-gap-seal-day-action


Chris0Jeky:
Required by PAPER-08 (#1004). Provide a 'seal day' snapshot action that freezes the day's ledger/dossier and is idempotent (second invoke is a no-op). Should record sealedAt timestamp on the daily snapshot. Currently stubbed via useTodayDossier with local mock state.

PR #1032, opened 2026-04-25
Add 90-day streak query for Today view

### LINKED ISSUE #1016: paper-today-backend-gap-streak-query
Required by PAPER-08 (#1004). Provide a 90-day streak query returning sealed/unsealed flag and intensity bucket per day for the current user. Drives the Today Streak grid (30 cols x 3 rows). Currently stubbed via useTodayDossier with mock data.

Chris0Jeky:
## Summary
- Adds `StreakDay` and `StreakResult` domain value objects for daily activity tracking with intensity buckets (0-4) and sealed flag
- Implements `StreakService` that queries `IAuditLogRepository` for user audit entries, groups by date, computes quartile-based intensity buckets, and calculates current/longest streak lengths
- Adds `GET /api/today/streak?days=90` endpoint on new `TodayController` (authenticated, claims-first)
- Wires `IStreakService` into DI registration
- 57 tests: 23 domain (StreakDay/StreakResult validation, equality, edge cases) + 34 application (StreakService: empty history, single day, continuous streak, gap handling, intensity bucketing boundaries, day count validation)

## Notes
- `IsSealed` defaults to `false` for all days -- will be driven by #1017 seal entity when available
- Intensity bucketing uses quartile-based approach: count/max ratio mapped to buckets 1-4, with 0 for no activity
- Day count parameter validated to 1-365 range

Closes #1016
Required by PAPER-08 (#1004)

## Test plan
- [x] `dotnet build backend/Taskdeck.sln -c Release` -- 0 errors
- [x] `dotnet test --filter "FullyQualifiedName~Streak"` -- 57 tests pass (23 domain + 34 application)
- [ ] Adversarial self-review of all changed files

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/TodayController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/TodayController.cs
@@ -31,6 +31,7 @@ public class TodayController : AuthenticatedControllerBase
     /// Get streak data (daily activity intensity and sealed status) for the authenticated user.
     /// </summary>
     /// <param name="days">Number of days to include (1-365, default 90).</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
     /// <returns>Streak data with daily intensity buckets and streak lengths.</returns>
     /// <response code="200">Streak data returned successfully.</response>
     /// <response code="400">Invalid days parameter.</response>
--- a/backend/src/Taskdeck.Api/Controllers/TodayController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/TodayController.cs
@@ -39,12 +39,14 @@ public class TodayController : AuthenticatedControllerBase
     [ProducesResponseType(typeof(StreakResponse), StatusCodes.Status200OK)]
     [ProducesResponseType(typeof(ApiErrorResponse), StatusCodes.Status400BadRequest)]
     [ProducesResponseType(typeof(ApiErrorResponse), StatusCodes.Status401Unauthorized)]
-    public async Task<IActionResult> GetStreak([FromQuery] int days = 90)
+    public async Task<IActionResult> GetStreak(
+        [FromQuery] int days = 90,
+        CancellationToken cancellationToken = default)
     {
         if (!TryGetCurrentUserId(out var userId, out var errorResult))
             return errorResult!;
 
-        var result = await _streakService.GetStreakAsync(userId, days);
+        var result = await _streakService.GetStreakAsync(userId, days, cancellationToken);
 
         if (!result.IsSuccess)
             return result.ToErrorActionResult();
--- a/backend/src/Taskdeck.Api/Controllers/TodayController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/TodayController.cs
@@ -1,0 +1,61 @@
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Taskdeck.Api.Contracts;
+using Taskdeck.Api.Extensions;
+using Taskdeck.Application.DTOs;
+using Taskdeck.Application.Interfaces;
+using Taskdeck.Application.Services;
+
+namespace Taskdeck.Api.Controllers;
+
+/// <summary>
+/// Today view endpoints: streak data, dossier, and daily summaries.
+/// </summary>
+[ApiController]
+[Authorize]
+[Route("api/[controller]")]
+[Produces("application/json")]
+public class TodayController : AuthenticatedControllerBase
+{
+    private readonly IStreakService _streakService;
+
+    public TodayController(
+        IStreakService streakService,
+        IUserContext userContext)
+        : base(userContext)
+    {
+        _streakService = streakService;
+    }
+
+    /// <summary>
+    /// Get streak data (daily activity intensity and sealed status) for the authenticated user.
+    /// </summary>
+    /// <param name="days">Number of days to include (1-365, default 90).</param>
+    /// <returns>Streak data with daily intensity buckets and streak lengths.</returns>
+    /// <response code="200">Streak data returned successfully.</response>
+    /// <response code="400">Invalid days parameter.</response>
+    /// <response code="401">Authentication required.</response>
+    [HttpGet("streak")]
+    [ProducesResponseType(typeof(StreakResponse), StatusCodes.Status200OK)]
+    [ProducesResponseType(typeof(ApiErrorResponse), StatusCodes.Status400BadRequest)]
+    [ProducesResponseType(typeof(ApiErrorResponse), StatusCodes.Status401Unauthorized)]
+    public async Task<IActionResult> GetStreak([FromQuery] int days = 90)
+    {
+        if (!TryGetCurrentUserId(out var userId, out var errorResult))
+            return errorResult!;
+
+        var result = await _streakService.GetStreakAsync(userId, days);
+
+        if (!result.IsSuccess)
+            return result.ToErrorActionResult();
+
+        var streakResult = result.Value;
+        var response = new StreakResponse(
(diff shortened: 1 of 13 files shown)


Chris0Jeky:
## Adversarial Self-Review Findings

### Fixed
1. **Missing CancellationToken in TodayController.GetStreak** (fixed in a7fdad9e): The endpoint did not accept or forward a `CancellationToken`, meaning HTTP request cancellation would not propagate to the audit log query. Now accepts `CancellationToken cancellationToken = default` and forwards to `GetStreakAsync`.

### Verified Correct
2. **Date arithmetic**: `today.AddDays(-(dayCount - 1))` correctly produces exactly `dayCount` days inclusive. For dayCount=1, only today. For dayCount=90, today minus 89 days through today = 90 days.
3. **Time range boundaries**: `from` uses `TimeOnly.MinValue` (00:00:00.0) and `to` uses `TimeOnly.MaxValue` (23:59:59.9999999), ensuring full-day coverage at both ends.
4. **UTC date grouping**: `DateOnly.FromDateTime(a.Timestamp.UtcDateTime.Date)` correctly extracts the UTC date regardless of the original offset on the audit log timestamp.
5. **Intensity bucketing boundaries**: Verified the ratio-based bucketing is consistent:
   - ratio <= 0.25 => bucket 1
   - ratio <= 0.50 => bucket 2  
   - ratio <= 0.75 => bucket 3
   - ratio > 0.75 => bucket 4
   
   At exactly 0.25 (25/100), the day lands in bucket 1, not 2. This is correct -- the `<=` makes boundaries inclusive of the lower bucket.
6. **Current streak counting**: Iterates backwards from last element (today), breaks on first zero-intensity day. No off-by-one -- starts at `days.Count - 1` and decrements to 0.
7. **Longest streak counting**: Forward scan with running counter, resets on zero. Correctly tracks maximum during iteration.
8. **Security**: Controller inherits `AuthenticatedControllerBase`, has `[Authorize]`, uses `TryGetCurrentUserId` from JWT claims. No user-supplied identity in the query path -- userId comes exclusively from claims. Architecture tests pass.
9. **StreakResult invariant**: `currentStreakLength > longestStreakLength` is rejected at construction time, preventing logically impossible states.

### Accepted Limitations
10. **Audit log limit**: Query uses `limit: 100_000`. For 365 days of very heavy activity this could theoretically truncate results, but 100K is a reasonable practical cap for a single user's audit entries over one year.
11. **Reflection in tests**: `CreateAuditLog` uses reflection to set `Timestamp` (private setter). This is fragile but acceptable for test infrastructure -- the alternative would require modifying the domain entity's API surface.
12. **IsSealed always false**: Documented as pending #1017 seal entity. No bug, just a known stub.

Chris0Jeky:
## Adversarial Review -- PR #1032 (paper/1016-streak-query)

### Issues Found and Fixed

**HIGH -- Memory risk: loading full AuditLog entities for streak computation**
`StreakService.GetStreakAsync` was calling `QueryAsync` with `limit: 100_000`, which loads full `AuditLog` entities into memory. Each entity carries `Changes` (up to 4KB string), `EntityType`, `EntityId`, `Action`, `UserId`, `Timestamp`, plus `Entity` base class fields (`Id`, `CreatedAt`, `UpdatedAt`) and a `User` navigation property. At 100k rows this could consume hundreds of MB.

**Fix:** Added `IAuditLogRepository.CountByDateAsync(from, to, userId)` which returns `IReadOnlyList<DailyAuditCount>` -- a lightweight `(DateOnly Date, int Count)` projection. The implementation uses server-side `GROUP BY` (raw SQL with `SUBSTR` for SQLite, EF LINQ for others), returning at most 365 rows regardless of audit log volume. The `IX_AuditLogs_UserId_Timestamp` composite index covers the WHERE clause. `StreakService` now calls this instead of `QueryAsync`.

**MEDIUM -- Correctness risk: silent data truncation at 100k**
If a user exceeded 100k audit entries in the window, streak calculation would silently operate on incomplete data. This is now moot: `CountByDateAsync` has no row limit -- it aggregates on the server and returns one row per active day (max 365).

**MEDIUM -- `DateTime.UtcNow` directly in service (testability)**
No `IClock`, `IDateTimeProvider`, or `TimeProvider` abstraction exists anywhere in the codebase. Direct `DateTime.UtcNow` / `DateTimeOffset.UtcNow` usage is the established pattern across 40+ call sites in `Taskdeck.Application.Services`. Introducing a clock abstraction would be a cross-cutting change beyond this PR's scope. **Acceptable for now** -- flagged as future improvement if date-boundary testing becomes necessary.

### Additional Fix

**LOW -- CS1573 warning in TodayController**
Missing `<param name="cancellationToken">` in XML doc for `GetStreak`. Added the missing tag.

### Files Changed by This Review

| File | Change |
|------|--------|
| `IAuditLogRepository.cs` | Added `DailyAuditCount` record and `CountByDateAsync` method |
| `AuditLogRepository.cs` | Implemented `CountByDateAsync` with SQLite raw SQL + EF LINQ paths |
| `StreakService.cs` | Replaced `QueryAsync(limit: 100_000)` with `CountByDateAsync` |
| `StreakServiceTests.cs` | Rewrote mocks from `QueryAsync`/`AuditLog` to `CountByDateAsync`/`DailyAuditCount` |
| `AuditRetentionWorkerTests.cs` | Added `CountByDateAsync` stub to `FakeAuditLogRepository` |
| `TodayController.cs` | Added missing XML doc param tag |

### Verification

- `dotnet build backend/Taskdeck.sln -c Release` -- 0 errors
- `dotnet test --filter "FullyQualifiedName~Streak"` -- 57/57 passed (23 Domain + 34 Application)
- `dotnet test --filter "FullyQualifiedName~AuditRetentionWorker"` -- 4/4 passed

### Remaining Notes

- The pure computation methods (`ComputeDays`, `ComputeIntensityBucket`, `ComputeCurrentStreak`, `ComputeLongestStreak`) are well-structured, thoroughly tested, and correct.
- Domain value objects `StreakDay` and `StreakResult` have proper validation and equality semantics.
- Controller properly uses `[Authorize]`, `AuthenticatedControllerBase`, `CancellationToken` forwarding, and `Result<T>` error mapping.

Chris0Jeky:

### CRITICAL

**1. Non-SQLite LINQ path: `DateTimeOffset.Date` GROUP BY will fail on SQL Server (Confidence: 95)**
`AuditLogRepository.CountByDateAsync` non-SQLite branch uses `.GroupBy(al => al.Timestamp.Date)`. EF Core for SQL Server cannot translate `DateTimeOffset.Date` in GROUP BY — runtime crash. SQLite path works via raw SQL, but the generic fallback is broken.

**Fix:** Use `.GroupBy(al => new { al.Timestamp.Year, al.Timestamp.Month, al.Timestamp.Day })` or add `throw new NotSupportedException(...)` for non-SQLite since the project only uses SQLite currently.

### IMPORTANT

**2. Value objects placed in `Entities/` folder (Confidence: 82)** — `StreakDay`/`StreakResult` are pure value objects (no `Id`, no persistence) placed in the `Entities/` directory. Consider `ValueObjects/` or a comment clarifying convention.

**3. `AuditLog.UserId` is nullable but comparison undocumented (Confidence: 83)** — System-generated entries with `UserId = null` are silently excluded. Correct behavior, but not documented in the interface contract.

**4. SQLite `SUBSTR(Timestamp, 1, 10)` assumes fixed format (Confidence: 85)** — Fragile if EF Core changes DateTimeOffset serialization. Add cross-reference to existing `DeleteOldEntriesAsync` comment that documents this.

**5. No API integration test for the endpoint (Confidence: 84)** — Unit tests are thorough (61), but no HTTP pipeline test for `TodayController.GetStreak`.

### Verified Non-Issues
- Security: claims-first identity, 1-365 day count validation, no auth bypass
- Domain invariants: `currentStreak <= longestStreak` enforced at construction
- SQL parameterized (no injection), index covers query
- Clean architecture boundaries respected

PR #1034, opened 2026-04-25
Add card history ledger for proposal review (#1023)

### LINKED ISSUE #1023: paper-review-backend-gap-history
Tracking issue for per-card history support behind PAPER-06 (#1002).

The § V History section needs every touch on the affected card surfaced as a ledger row with serial / event / age / status.

## Required shape
```ts
interface HistoryRow {
  serial: string           // e.g. '#014'
  event: string            // human-readable event description
  age: string              // pre-formatted ('11:42', 'yest 16:04', 'Mon 11:00')
  status: 'pending' | 'applied' | 'past'
}
```

## Where this is wired
- Frontend: `frontend/taskdeck-web/src/composables/usePaperReviewSelectors.ts` (DEMO_HISTORY).
- See `design_handoff_taskdeck_paper/paper/surface-review-deep.jsx` § V.

Refs PAPER-06 (#1002).

Chris0Jeky:
## Summary
- Add `CardHistoryStatus` enum and `CardHistoryRow` value object in the Domain layer
- Add `ICardHistoryService` / `CardHistoryService` in the Application layer that collects audit log entries and proposal operations for all cards affected by a proposal, formats them as numbered history rows with relative-time age strings
- Add `GET /api/automation/proposals/{id}/history` endpoint with board-level authorization
- 42 tests: 17 domain (value object construction, validation, equality) + 25 application (service validation, empty/single/multi-card history, serial numbering, age formatting with UTC timezone handling, status classification, deduplication)

## Test plan
- [x] Domain tests: CardHistoryRow construction, blank field rejection, invalid status, equality, toString
- [x] Service tests: validation (empty IDs, missing proposal), empty history, single card with mixed audit/proposal entries, multi-card combination, sequential serial numbering, age formatting (same day, midnight boundary, yesterday, this week, older, non-UTC conversion), status classification (pending/applied/past), non-card operation handling, self-referencing proposal deduplication
- [x] `dotnet build backend/Taskdeck.sln -c Release` passes (0 errors)
- [x] `dotnet test --filter "FullyQualifiedName~CardHistory"` passes (42/42)

Closes #1023


Chris0Jeky:
## Adversarial Self-Review

### Issues found and fixed

1. **Locale-sensitive age formatting** (FIXED in fe5c670d): `ddd` and `MMM` format strings used `CultureInfo.CurrentCulture` by default. On a non-English locale server, "Mon" could appear as "Lun", "Apr" as "avr". Fixed by explicitly using `CultureInfo.InvariantCulture` for all date/time format operations.

2. **Unbounded total history rows** (FIXED in fe5c670d): While each card's audit entries are capped at 200, a proposal with N distinct card targets could produce N*200 entries total. Added `MaxTotalHistoryRows = 500` cap applied after sorting (so newest entries are always preserved).

### Issues reviewed and accepted

3. **`GetLatestByOperationTargetAsync` returns only the most recent proposal per card**: The repository API only exposes the latest related proposal, not all historical proposals that targeted each card. This means the ledger may not show every proposal touch. Acceptable for v1 since the audit log already captures the actual mutations those proposals caused. A future enhancement could add a `GetAllByOperationTargetAsync` query if the UI needs more complete proposal lineage.

4. **Midnight boundary / timezone handling**: All comparisons and formatting consistently use UTC (`timestamp.UtcDateTime`, `now.UtcDateTime`). Tests cover the midnight boundary case (00:01 on day N vs 23:59 on day N-1) and non-UTC offset conversion. No issues found.

5. **Serial numbering off-by-one**: Serial starts at `FormatSerial(i + 1)` for 0-based index `i`, producing `#001` for the first row. Tests verify this explicitly. No off-by-one.

6. **Card access authorization**: The endpoint uses `AuthorizeProposalAsync` which checks board-level read permission. Since cards belong to boards, board-level auth is sufficient -- if you can read the board, you can read its card history. No authorization gap.

7. **Audit log query performance**: Each card query is bounded at `MaxAuditEntriesPerCard = 200`. The number of cards is bounded by the proposal's operations count (typically small, single digits). Total output is further capped at 500 rows. Query count is O(cards) where cards = distinct card targets in the proposal's operations.

Chris0Jeky:
## Adversarial Review: CardHistoryService — All 5 Gemini findings addressed

### Fixes applied

**1. HIGH: Duplicate proposal entries (line 99)**
When multiple cards shared the same related proposal, it appeared multiple times. Fixed by initializing `var seenProposalIds = new HashSet<Guid> { proposalId }` and using `seenProposalIds.Add()` as the guard. This also subsumes the existing self-reference check (`relatedProposal.Id != proposalId`). Added a dedicated test (`MultiCard_ShouldDeduplicateSharedRelatedProposal`) that verifies only 1 applied entry appears when 2 cards return the same related proposal.

**2. MEDIUM: Unused `userId` parameter (line 30)**
Removed `userId` from `ICardHistoryService.GetCardHistoryForProposalAsync`, the service implementation, and the controller call site. Auth is fully handled by the controller's `AuthorizeProposalAsync` — the service has no use for it. Updated all 14 test call sites and removed the now-unnecessary userId validation test.

**3. MEDIUM: Double GUID parsing (line 54)**
Replaced the LINQ chain (`Where(TryParse) + Select(Parse)`) with a single `foreach` loop that calls `Guid.TryParse` once, producing the parsed `Guid` via the `out` parameter. Also reuses a `HashSet<Guid>` for distinctness instead of `.Distinct()`.

**4. MEDIUM: N+1 query for audit logs (line 80)**
No batch `GetByEntitiesAsync` exists on `IAuditLogRepository`. Added an explicit comment documenting this as a known N+1, acceptable for the typical 1–5 cards per proposal, with a pointer to add a batch method if this becomes a hot path.

**5. MEDIUM: Brittle `string.Contains` on raw JSON (line 229)**
Replaced all `string.Contains` checks with `JsonDocument.Parse` + `TryGetProperty`. This correctly checks for JSON property names rather than substring matching on values. Added a `JsonException` catch so malformed JSON falls through to the generic `"updated"` message. Also added `"labels"` (plural) as an additional property name check alongside `"label"`.

### Additional finding

No additional bugs found beyond the 5 reported items. The `FormatAge`, `FormatSerial`, `FormatAuditEvent`, `FormatOperationEvent`, and `FormatProposalStatus` helpers are clean. The controller endpoint correctly uses `requireWriteAccess: false` for this read-only endpoint. DI registration is correct.

### Verification

- `dotnet build backend/Taskdeck.sln -c Release` — 0 errors
- `dotnet test --filter "FullyQualifiedName~CardHistory"` — 42 tests pass (17 domain, 25 application including new dedup test)

Chris0Jeky:

### CRITICAL

**1. Case-sensitive `"Card"` vs `"card"` mismatch silently breaks related proposal lookups (Confidence: 95)**
`CardHistoryService` passes `"Card"` (PascalCase) to `GetLatestByOperationTargetAsync`, but production operations store `TargetType` as `"card"` (lowercase). The repository does exact `==` comparison. Result: the query **always returns null**, making the "Applied" status marking dead code. Tests pass because they mock the repository.

**Fix:** Change `"Card"` to `"card"` in the `GetLatestByOperationTargetAsync` call, or extract a shared constant.

### IMPORTANT

**2. `CardHistoryRow` domain entity is dead code (Confidence: 92)** — 64 lines with 17 tests, but never referenced by any Application or Api layer code. The service exclusively uses `CardHistoryRowDto`.

**3. Related proposals query returns only the single "latest" per card (Confidence: 85)** — A card affected by 5 proposals over its lifetime only shows the most recent one. Undercuts the "history ledger" value.

**4. `FormatOperationEvent` includes non-card operations (Confidence: 82)** — Pending entries include column operations mixed into card history.

**5. `FormatUpdateEvent` JSON property checking is case-sensitive (Confidence: 80)** — If audit log stores PascalCase properties, all checks miss and fall through to generic "updated" message.

### Verified Non-Issues
- Security: `AuthorizeProposalAsync` with `requireWriteAccess: false` is correct
- Serial numbering is correct (1-based, zero-padded)
- Age formatting uses `InvariantCulture` consistently
- Deduplication via `HashSet<Guid>` is correct
- JSON parsing catches `JsonException` — no crash on malformed data

PR #1037, opened 2026-04-25
Add seal-day backend action for daily snapshots

### LINKED ISSUE #1017: paper-today-backend-gap-seal-day-action

Chris0Jeky:
## Summary

Closes #1017. Implements the backend 'seal day' snapshot action required by PAPER-08.

- **Domain**: `DailySnapshot` entity with `Seal(DateTimeOffset)` (idempotent -- second invoke is a no-op), `IsSealed` computed property, validation (no future dates, non-empty userId)
- **Application**: `IDailySnapshotRepository` interface (user+date lookup, sealed-days range query), `IDailySealService`/`DailySealService` with get-or-create + idempotent seal logic, response DTOs
- **Infrastructure**: `DailySnapshotRepository`, EF Core configuration with unique index on `(UserId, Date)`, EF migration for `DailySnapshots` table, DI wiring in `UnitOfWork` and `DependencyInjection`
- **API**: `TodayController` with `POST /api/today/seal` (seal a day) and `GET /api/today/seal?date=...` (check seal status), claims-first identity via `AuthenticatedControllerBase`
- **Tests**: 10 domain tests (construction, seal idempotency, future date rejection, IsSealed), 10 service tests (seal new/existing/already-sealed, validation, status checks)

## Adversarial self-review findings

1. **Concurrency (two simultaneous seal requests)**: The unique index on `(UserId, Date)` provides database-level protection against duplicate snapshots. If two requests race to create a snapshot for the same user+date, one will get a `DbUpdateException` caught by `UnitOfWork.SaveChangesAsync` which wraps it in a `DomainException` with `Conflict` error code. The `Seal()` method itself is idempotent, so if both requests find an existing row, both succeed harmlessly.

2. **Date timezone handling**: The service uses `DateTimeOffset.UtcNow` consistently for the `now` parameter. The `DateOnly` comparison in the domain constructor uses `DateOnly.FromDateTime(now.UtcDateTime)` to ensure UTC-based date boundary. The `date` parameter is `DateOnly` (no timezone ambiguity). The API accepts `DateOnly` from the request body/query, which serializes as `YYYY-MM-DD` with no timezone component -- the caller decides which calendar day to seal.

3. **Migration correctness**: The migration creates the table with the correct columns, primary key, and unique composite index on `(UserId, Date)`. The `Down()` method drops the table cleanly.

4. **Idempotency guarantee**: Enforced at two levels -- domain (`Seal()` is a no-op if already sealed) and service (returns `WasAlreadySealed: true` without mutation). The `UpdatedAt` concurrency token prevents lost updates if the row changes between read and write.

5. **Missing edge case**: No explicit transaction wrapping in `SealDayAsync` for the create-then-seal path. With EF Core change tracking, both the `AddAsync` and the implicit `Seal()` mutation are persisted in a single `SaveChangesAsync` call, which is atomic. This is safe.

## Test plan

- [x] Domain: DailySnapshot creation with valid/invalid inputs
- [x] Domain: Seal idempotency (multiple calls preserve original timestamp)
- [x] Domain: Future date rejection
- [x] Domain: IsSealed computed property
- [x] Service: Seal new day creates and seals snapshot
- [x] Service: Seal existing unsealed day seals it
- [x] Service: Seal already-sealed day returns idempotent response
- [x] Service: Empty userId validation
- [x] Service: Future date validation
- [x] Service: GetSealStatus for missing/sealed/unsealed snapshots
- [x] Build: `dotnet build backend/Taskdeck.sln -c Release` passes (0 errors)
- [x] Tests: All 20 new tests pass


Chris0Jeky:
## Adversarial Self-Review

### Concurrency: two simultaneous seal requests

The unique index `IX_DailySnapshots_UserId_Date` on `(UserId, Date)` provides database-level protection. If two requests race to create a snapshot for the same user+date:

- **Both find null**: One `SaveChangesAsync` succeeds, the other gets a `DbUpdateException` from the unique constraint violation. The `UnitOfWork.SaveChangesAsync` wraps concurrency exceptions as `DomainException(Conflict)`, which maps to HTTP 409. The client can retry and will hit the idempotent path.
- **Both find existing row**: Both call `Seal()` which is idempotent. The `UpdatedAt` concurrency token means one save wins and the other gets a concurrency exception (also mapped to 409).
- **One finds null, one finds existing**: The creator's save wins; the other may get a unique constraint violation on create, or succeeds on the existing row.

**Verdict**: Safe. The worst case is a retriable 409 conflict, never data corruption.

### Date timezone edge cases

The `date` parameter is `DateOnly` (no timezone component). The future-date check uses `DateOnly.FromDateTime(now.UtcDateTime)` consistently. This means:
- A user in UTC+13 could seal "tomorrow" in UTC terms, but the system prevents it by comparing against UTC now.
- The caller decides which calendar day to seal -- the backend does not infer the user's timezone.

**Verdict**: Correct for a UTC-normalized system. If the product later needs timezone-aware day boundaries, the API contract (`DateOnly`) supports passing the user's local date without backend changes.

### Migration correctness

- Table schema matches entity properties exactly.
- Unique composite index prevents duplicate snapshots.
- `Down()` drops the table cleanly.
- Namespace matches the `Migrations/` folder convention.

**Verdict**: Correct.

### Idempotency guarantees

- Domain: `Seal()` checks `IsSealed` and returns early if true.
- Service: Detects `wasAlreadySealed` before calling `Seal()` and reports it in the response.
- Database: Single `SaveChangesAsync` call is atomic.

**Verdict**: Strong idempotency at all layers.

### No issues found that require fixing.

Chris0Jeky:
## Adversarial Review -- Findings & Fixes

Reviewed all 24 files in this PR. Addressing the 6 Gemini bot comments plus additional verification.

### Issues Found & Fixed

**HIGH: Missing DailySnapshot unique constraint handler in UnitOfWork.cs**
- `TryResolveRecoverableUniqueConflicts` handled Notifications and UserPreferences but NOT DailySnapshots.
- Concurrent seal requests hitting the `IX_DailySnapshots_UserId_Date` unique index would throw unhandled `DbUpdateException` (500).
- **Fix**: Added `TryResolveDuplicateDailySnapshotConflicts` and `IsDailySnapshotUniqueViolation` methods following the established pattern. Wired into `TryResolveRecoverableUniqueConflicts`.

**MEDIUM: Missing CancellationToken propagation (4 locations)**
- `IDailySealService.SealDayAsync` / `GetSealStatusAsync` -- no CancellationToken param.
- `DailySealService.SealDayAsync` / `GetSealStatusAsync` -- no CancellationToken param or propagation.
- `TodayController.SealDay` / `GetSealStatus` -- no CancellationToken accepted from ASP.NET pipeline.
- **Fix**: Added `CancellationToken cancellationToken = default` to interface and service methods. Controller now accepts `CancellationToken` from the framework and propagates it. All downstream async calls (`GetByUserAndDateAsync`, `AddAsync`, `SaveChangesAsync`) now receive the token. Tests updated to use `It.IsAny<CancellationToken>()`.

### Additional Checks (All OK)

- **Repository methods**: `IDailySnapshotRepository` already accepts `CancellationToken` -- no change needed.
- **EF Migration**: Table and `IX_DailySnapshots_UserId_Date` unique index are correctly defined.
- **Idempotency**: `DailySnapshot.Seal()` is a no-op when already sealed; `SealDayAsync` correctly reports `WasAlreadySealed=true` and preserves the original timestamp. Truly idempotent.
- **EF Configuration**: `DailySnapshotConfiguration` correctly defines the unique index and concurrency token.
- **DI Registration**: Both `DailySnapshotRepository` and `DailySealService` are properly registered.

### Build & Test Results
- `dotnet build backend/Taskdeck.sln -c Release` -- **0 errors**
- 28 seal/snapshot tests -- **all pass** (16 domain, 12 application)

Chris0Jeky:

### CRITICAL

**1. Concurrent seal race returns fabricated response data (Confidence: 95)**
When two concurrent seal requests race on the same `(UserId, Date)`:
- Both find no existing snapshot, both create and `AddAsync`
- Loser hits unique constraint, `TryResolveDuplicateDailySnapshotConflicts` detaches the entity
- Retry `SaveChangesAsync` succeeds (saves nothing)
- Service returns `DailySealResponse(snapshot.SealedAt!.Value, wasAlreadySealed: false)` using the **detached, in-memory** snapshot

The losing request reports `WasAlreadySealed: false` when it was actually sealed by the other request, and returns its own in-memory `SealedAt` timestamp rather than the actual persisted value.

**Fix:** After `SaveChangesAsync`, re-fetch the snapshot from the database. Or catch `DomainException(Conflict)` at the service level, re-query, and return accurate `WasAlreadySealed: true`.

### IMPORTANT

**2. No controller-level integration tests (Confidence: 85)** — Domain and service tests exist (28 tests), but no HTTP pipeline test verifying routing, claims extraction, or status code mapping for `POST /api/today/seal` and `GET /api/today/seal`.

### Verified Non-Issues
- Security: claims-first identity, future date prevention at both domain and service levels
- Idempotency correct for non-concurrent path (second seal is no-op preserving original timestamp)
- Migration: proper unique composite index on `(UserId, Date)`, clean `Down()` method
- CancellationToken threaded through all layers
- Error contracts: `ErrorCodes.ValidationError` mapped to 400 consistently

PR #1038, opened 2026-04-25
Add similar past decisions endpoint for proposal review context

### LINKED ISSUE #1024: paper-review-backend-gap-similar-past
Tracking issue for similar-past-decisions query behind PAPER-06 (#1002).

The right rail surfaces 3 prior similar decisions plus an aggregate apply rate to give the reviewer a quick base rate.

## Required shape
```ts
interface SimilarPastRow {
  serial: string
  title: string
  verdict: 'applied' | 'rejected'
  date: string                   // pre-formatted 'wk 14'
}
```

Server query: nearest neighbours on the proposal's action class, returning the latest 3 with verdicts. Apply-rate denominator can include rejected for honest signal.

## Where this is wired
- Frontend: `frontend/taskdeck-web/src/composables/usePaperReviewSelectors.ts` (DEMO_SIMILAR).
- See `design_handoff_taskdeck_paper/paper/surface-review-deep.jsx` (right rail similar-past).

Refs PAPER-06 (#1002).

Chris0Jeky:
## Summary
- Adds `GET /api/automation/proposals/{proposalId}/similar-past` endpoint that returns the 3 most recent prior decisions with the same action class (create/move/archive/update) and an aggregate apply rate
- Domain layer: `PastVerdict` enum, `SimilarPastDecision` value object with validation/truncation, `SimilarPastResult` with division-by-zero-safe `ComputeApplyRate`
- Application layer: `SimilarDecisionService` matches on first operation's action type, queries board-scoped then falls back to user-scoped, formats serials as `#001`/`#002`/`#003`, dates as ISO week (`wk NN`), computes rate across all matches (not just top 3)
- Infrastructure: `GetTerminalByActionTypeAsync` repository method with bounded lookback (200 proposals)
- 50 new tests (24 domain + 26 application) covering: empty history, all-applied, all-rejected, mixed rates, top-3 limiting, self-exclusion, board-to-user fallback, ISO week formatting, verdict mapping, division-by-zero

Closes #1024

## Test plan
- [x] Domain tests: value object creation, validation, title truncation, apply rate edge cases (zero denominator, negative inputs)
- [x] Service tests: no-history empty result, rate calculations (0.0, 1.0, mixed), top-3 limiting with full-set rate, action-type matching, board-scoped fallback, self-exclusion, ISO week formatting, serial formatting
- [x] Full solution build passes
- [x] Existing tests unbroken (fake repositories updated with new interface method)

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/AutomationProposalsController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/AutomationProposalsController.cs
@@ -26,16 +26,19 @@ public class AutomationProposalsController : AuthenticatedControllerBase
 
     private readonly IAutomationProposalService _proposalService;
     private readonly IAutomationExecutorService _executorService;
+    private readonly ISimilarDecisionService _similarDecisionService;
     private readonly BoardAuthorizationService _authorizationService;
 
     public AutomationProposalsController(
         IAutomationProposalService proposalService,
         IAutomationExecutorService executorService,
+        ISimilarDecisionService similarDecisionService,
         BoardAuthorizationService authorizationService,
         IUserContext userContext) : base(userContext)
     {
         _proposalService = proposalService;
         _executorService = executorService;
+        _similarDecisionService = similarDecisionService;
         _authorizationService = authorizationService;
     }
 
@@ -276,6 +279,24 @@ public class AutomationProposalsController : AuthenticatedControllerBase
 
         var result = await _proposalService.GetProposalDiffAsync(id, cancellationToken);
         return result.IsSuccess ? Ok(new { diff = result.Value }) : result.ToErrorActionResult();
+    }
+
+    /// <summary>
+    /// Gets similar past decisions for a proposal, including the latest 3 decisions
+    /// with the same action class and an aggregate apply rate.
+    /// </summary>
+    [HttpGet("{id}/similar-past")]
+    public async Task<IActionResult> GetSimilarPast(Guid id, CancellationToken cancellationToken = default)
+    {
+        if (!TryGetCurrentUserId(out var callerUserId, out var errorResult))
+            return errorResult!;
+
+        var auth = await AuthorizeProposalAsync(id, callerUserId, requireWriteAccess: false, cancellationToken);
+        if (auth.ErrorResult is not null)
+            return auth.ErrorResult;
+
+        var result = await _similarDecisionService.GetSimilarPastAsync(id, callerUserId, cancellationToken);
+        return result.IsSuccess ? Ok(result.Value) : result.ToErrorActionResult();
     }
 
     private async Task<(ProposalDto? Proposal, IActionResult? ErrorResult)> AuthorizeProposalAsync(
--- a/backend/src/Taskdeck.Api/Extensions/ApplicationServiceRegistration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/ApplicationServiceRegistration.cs
@@ -48,6 +48,7 @@ public static class ApplicationServiceRegistration
         services.AddScoped<IHistoryService>(sp => sp.GetRequiredService<HistoryService>());
         services.AddScoped<IAutomationProposalService, AutomationProposalService>();
         services.AddScoped<IProposalRevisionService, ProposalRevisionService>();
+        services.AddScoped<ISimilarDecisionService, SimilarDecisionService>();
         services.AddScoped<IAutomationPolicyEngine, AutomationPolicyEngine>();
         services.AddScoped<IAutomationPlannerService, AutomationPlannerService>();
         services.AddScoped<IAutomationExecutorService, AutomationExecutorService>();
--- a/backend/src/Taskdeck.Application/DTOs/SimilarPastDtos.cs
+++ b/backend/src/Taskdeck.Application/DTOs/SimilarPastDtos.cs
@@ -1,0 +1,17 @@
+namespace Taskdeck.Application.DTOs;
+
+/// <summary>
+/// DTO for a single similar past decision.
+/// </summary>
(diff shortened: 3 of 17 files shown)


Chris0Jeky:
## Adversarial Self-Review

### Division by zero
**Status: Safe.** `SimilarPastResult.ComputeApplyRate(0, 0)` returns `0.0` explicitly. Negative inputs throw `ArgumentOutOfRangeException`. Both paths are covered by domain tests.

### Action class matching
**Status: Exact match, which is correct.** Action types are fixed strings (`create`, `move`, `archive`, `update`) set by the tool executors. Fuzzy matching would be incorrect here since these are code-defined constants, not user input.

### Date formatting (ISO vs US weeks)
**Status: Correct.** Uses `System.Globalization.ISOWeek.GetWeekOfYear()` which follows ISO 8601 (Monday-start, first 4-day week rule). The `DateTime`-to-`DateTimeOffset` implicit conversion is safe because `DecidedAt` is set via `DateTime.UtcNow` (Kind=UTC). Test verifies: April 6, 2026 = wk 15, and Dec 29, 2025 = wk 1 (of 2026's first ISO week).

### Query performance
**Status: Bounded.** `LookbackLimit = 200` prevents unbounded queries. The repository uses a two-query pattern (IDs first, then full entities with operations) matching existing codebase patterns. The `ActionType` exact match on operations is efficient. The `OrderByDescending(DecidedAt ?? UpdatedAt)` with `Take(limit)` ensures bounded result sets.

### Authorization
**Status: Consistent.** Uses the same `AuthorizeProposalAsync` helper as all other proposal endpoints, checking board read permission. User ID derived from JWT claims via `TryGetCurrentUserId`. No authorization bypass paths found.

### Potential improvements (not blocking)
- The `using System.Globalization` import is needed for `ISOWeek` but could be a static import for clarity
- The board-to-user fallback makes 2 DB queries when the board has no history. This is acceptable for the expected query volume but could be optimized with a single query that tries board-scoped first and widens if needed
- The `FormatWeekDate` only shows the week number, not the year. For cross-year reviews, `wk 1` could be ambiguous (is it this year or last year?). Consider `wk 1 '26` format in a future iteration

Chris0Jeky:
## Adversarial Review -- PR #1038 (similar-past decisions)

### Issues Found and Fixed

**HIGH: Wrong userId filter in `AutomationProposalRepository.cs:240`**
`GetTerminalByActionTypeAsync` always filtered by `p.RequestedByUserId == userId`, which only returned proposals *created by* the caller. When a reviewer looks at similar past decisions for a board, they should see all decisions made on that board -- not just their own proposals. Fixed: when `boardId` is present, filter by board only (no userId); when `boardId` is null (user-scoped fallback), filter by userId.

**MEDIUM: UTC for week calculation in `SimilarDecisionService.cs:152`**
`dateTime.DateTime` can silently shift to the wrong date depending on server timezone offset (e.g. UTC+13 could shift a Jan 1 date backwards or forwards across a year boundary). Fixed: use `dateTime.UtcDateTime` for ISO week calculation.

**MEDIUM: Non-SARGable OrderBy in `AutomationProposalRepository.cs:248`**
`OrderByDescending(p => p.DecidedAt ?? p.UpdatedAt)` forces EF Core to emit a `COALESCE` expression that prevents index usage. Since terminal proposals (Applied/Rejected) should always have `DecidedAt` populated, fixed both the SQL query and the in-memory re-sort to use `.OrderByDescending(p => p.DecidedAt).ThenByDescending(p => p.UpdatedAt)`.

### Additional Checks (all clear)

- **Division-by-zero**: `ComputeApplyRate` handles the zero-denominator case explicitly (`total == 0 ? 0.0`), with guard clauses for negative inputs. Safe.
- **Action class matching correctness**: Uses first operation by `Sequence` ordering, which is correct for both single-op and multi-op proposals.
- **Query performance / lookback bound**: Bounded by `LookbackLimit = 200`. The operation-to-proposal join uses `.Distinct()` before joining. Acceptable.
- **Self-exclusion**: Current proposal is excluded at line 68-69 via `.Where(p => p.Id != proposalId)`. Correct.
- **Test coverage**: 50 tests covering domain value objects (24) and service logic (26), including edge cases for empty results, self-exclusion, fallback, serials, date formatting, and verdict mapping.

Chris0Jeky:

### IMPORTANT

**1. [SECURITY] User-scoped fallback leaks cross-board proposal titles (Confidence: 85)**
When board-scoped query returns empty, fallback queries the caller's proposals across ALL boards. Titles from other boards appear in the response. If another user on the current board can observe this (shared review UI, caching, SignalR), they could see cross-board titles.

**Fix:** Filter out proposals whose `BoardId` doesn't match current board in fallback, or redact titles to generic form.

**2. Missing `AsNoTracking()` on read-only repository query (Confidence: 88)**
`GetTerminalByActionTypeAsync` loads up to 200 proposals with operations into EF Core change tracker. Read-only query that would benefit from `AsNoTracking()` on both the ID query and entity+operations query.

**3. No index on `AutomationProposalOperations.ActionType` (Confidence: 82)**
The query filters operations by `ActionType` but no index exists. Operations table scan grows with data. **Fix:** Add composite index `(ActionType, ProposalId)`.

**4. `SimilarPastDecision` public constructor bypasses validation (Confidence: 83)**
`Create` factory validates inputs, but the positional record constructor is public and accepts anything. Same for `SimilarPastResult.ApplyRate` accepting NaN/negative.

**Fix:** Make primary constructor private, expose only `Create`.

### Verified Non-Issues
- Board-scoped query correctly excludes non-caller proposals (userId filter fix verified)
- 200-proposal lookback limit is bounded
- Apply rate computation handles division-by-zero
- Security: `AuthorizeProposalAsync` gating correct
- 50 tests with good coverage

PR #1036, opened 2026-04-25
Add confidence breakdown API for proposal review UI

Chris0Jeky:
## Summary

Implements the backend confidence breakdown system (#1021) so the paper deep-review header dial and right-rail bars can show a 4-component explanation of why a proposal is or isn't above the apply threshold.

- **Domain layer**: `ConfidenceComponent` and `ConfidenceBreakdown` value objects with full [0..1] range validation, NaN/Infinity rejection, equality, and defensive copy semantics
- **Application layer**: `IConfidenceBreakdownService` / `ConfidenceBreakdownService` computing 4 components (Pattern match, Reach, Reversibility, Recency) from proposal metadata with weighted overall score and threshold-relative note generation
- **API layer**: `GET /api/automation/proposals/{id}/confidence` endpoint with board-scoped authorization
- **Tests**: 30 domain tests + 33 service tests covering validation, edge cases, and integration scenarios

### Component computation
| Component | Weight | Signal |
|-----------|--------|--------|
| Pattern match | 0.30 | Proportion of operations using well-known action types |
| Reach | 0.20 | Inverse log of distinct affected entities (focused = high) |
| Reversibility | 0.35 | Blend of risk level baseline + action-type destructiveness |
| Recency | 0.15 | Remaining fraction of proposal expiry window |

### Response shape
```json
{
  "overall": 0.82,
  "components": [
    { "key": "Pattern match", "value": 1.0 },
    { "key": "Reach", "value": 0.67 },
    { "key": "Reversibility", "value": 0.9 },
    { "key": "Recency", "value": 0.95 }
  ],
  "note": null,
  "threshold": 0.7,
  "meetsThreshold": true
}
```

## Test plan
- [x] Domain: ConfidenceComponent validation (0..1 range, NaN, Infinity, empty key)
- [x] Domain: ConfidenceBreakdown validation (overall, threshold, null components, boundary values, MeetsThreshold)
- [x] Domain: Equality and hash code contracts
- [x] Service: Breakdown computation for different proposal types
- [x] Service: Each component computes in [0, 1] range
- [x] Service: Pattern match for known/unknown action types
- [x] Service: Reach for single vs multi-target proposals
- [x] Service: Reversibility for different risk levels and action types
- [x] Service: Overall weighted average correctness
- [x] Service: Note generation near/far from threshold
- [x] Service: Not-found proposal returns proper error
- [x] `dotnet build backend/Taskdeck.sln -c Release` passes
- [x] 63 new tests all pass

Closes #1021

PR #1039, opened 2026-04-25
Expose proposal provenance via API for Paper deep-Review

### LINKED ISSUE #1019: paper-review-backend-gap-provenance
Tracking issue for backend provenance support behind PAPER-06 (#1002).

The Paper deep-Review surface needs the proposal payload to expose the read-set / not-read / inferred fields so the § II Provenance section can render real data instead of the demo stub.

## Required shape (TypeScript reference)
```ts
type ProvenanceWeight = 'primary' | 'contextual' | 'excluded' | 'inferred'
interface ProvenanceRow {
  icon: string         // emoji glyph chosen server-side or via a stable map
  key: string          // e.g. 'card body', 'design-doc · #notion'
  value: string        // human-readable description
  weight: ProvenanceWeight
}
```

## Where this is wired
- Frontend: `frontend/taskdeck-web/src/composables/usePaperReviewSelectors.ts`
  currently returns `DEMO_PROVENANCE` whenever the active proposal is
  non-null. Once the backend ships the field, replace the demo branch with
  the proposal payload and flip `PAPER_REVIEW_DEMO_FILL` to `false`.
- See `design_handoff_taskdeck_paper/paper/surface-review-deep.jsx` § II.

Refs PAPER-06 (#1002).

Chris0Jeky:
## Summary

- Adds `GET /api/automation/proposals/{id}/provenance` endpoint that returns `ProvenanceRowDto[]` with icon, key, value, and weight fields for the Paper deep-Review section II Provenance surface
- Implements `ProvenanceQueryService` with stable icon mapping (30+ field names to emoji), weight bucketing (extractive high-confidence -> primary, low-confidence -> contextual, inferred -> inferred), and human-readable value strings
- Adds full EF Core persistence for `ProposalProvenance`, `ProvenanceField`, and `ProvenanceEvidenceLink` entities (DbSets, configurations, migration, repository)
- Authorization reuses existing `AuthorizeProposalAsync` for board read-access checks

## Details

**Application layer:**
- `ProvenanceRowDto` (Icon, Key, Value, Weight) in DTOs
- `IProvenanceQueryService` and `IProposalProvenanceRepository` interfaces
- `ProvenanceQueryService` with icon map, weight bucketing, value building, and quote truncation

**Infrastructure layer:**
- EF entity type configurations with indexes and cascading deletes
- `ProposalProvenanceRepository` with eager-loading of Fields and EvidenceLinks
- EF migration `AddProvenanceEntities` creating 3 tables

**API layer:**
- Endpoint added to existing `AutomationProposalsController`
- Returns empty array (not error) when no provenance exists for a proposal

## Test plan

- [x] 40 new unit tests covering icon mapping, weight bucketing, value generation, edge cases
- [x] All 132 provenance-related tests pass (54 domain + 71 application + 7 API migration)
- [x] Full solution build succeeds (0 errors)
- [x] Migration bootstrap tests pass after adding EF migration

Closes #1019


Chris0Jeky:
## Adversarial Self-Review

### Authorization

**OK.** The provenance endpoint reuses `AuthorizeProposalAsync` which checks both board-level read access (for board-scoped proposals) and proposal ownership (for non-board proposals). No bypass risk -- the same pattern is used by `GetProposal`, `GetProposalDiff`, and other read endpoints.

### Null handling

**OK.** When no provenance exists for a proposal, `GetProvenanceRowsAsync` returns `Result.Success` with an empty list (not a 404 or error). This matches the expected behavior: many proposals have no provenance yet, and the frontend should render an empty section rather than an error.

### Icon map completeness

**Acceptable.** The icon map covers 30+ field names including aliases (e.g., "due date" / "duedate" / "due_date"). Unknown field names fall back to a generic page icon. The map is case-insensitive. Risk: if new domain field names are added to `ProvenanceField` usage but not to the icon map, they silently get the fallback icon -- this is intentional (no crash, just a less specific icon).

### Weight mapping edge cases

**Documented limitation.** The `excluded` weight bucket is never generated from `ProvenanceField` data because the domain model has no concept of "excluded sources" (only extractive and inferred). The frontend demo stub shows "excluded" rows (e.g., "not read" / "other boards"). A future enhancement could add a dedicated excluded-sources field to `ProposalProvenance` or generate synthetic rows. This is explicitly documented in `MapWeight` comments.

### Confidence boundary at 0.7

**Deliberate.** The threshold `>= 0.7` for "primary" vs "contextual" is a stable heuristic boundary. The 0.7 value was chosen to align with the existing `ConfidenceBreakdown.threshold` default in the frontend (`DEMO_CONFIDENCE`). The boundary is tested in `MapWeight_CorrectlyBuckets` with values at 0.7 (primary) and 0.69 (contextual).

### Quote truncation

**OK.** Long extractive quotes are truncated to 117 chars + "..." to keep the value string scannable. The truncation boundary (120 chars) was chosen to fit typical card-width rendering. UTF-8 multi-byte characters could theoretically be split mid-codepoint by `AsSpan(0, 117)`, but C# `string.AsSpan` operates on chars (UTF-16 code units), not bytes, so surrogate pairs are the only risk -- and surrogate pairs in proposal field quotes are extremely unlikely.

### EF configuration

**OK.** The `ProposalProvenances.ProposalId` index is `unique: true` (one provenance chain per proposal). Cascading deletes from provenance -> fields -> evidence links are intentional and match the domain ownership hierarchy. The configuration uses `ValueGeneratedNever` for IDs, consistent with all other entity configurations in the project.

### Missing: no FK from ProposalProvenance to AutomationProposal

**Noted.** The `ProposalProvenances` table has a `ProposalId` column with a unique index but no formal FK constraint to `AutomationProposals`. This matches the existing domain design where `ProposalProvenance.ProposalId` is a plain Guid with validation in the constructor (non-empty), not a navigation property. Adding an FK would require a domain model change (adding a navigation property to `AutomationProposal`) which is out of scope for this issue. The unique index on `ProposalId` is sufficient for query performance and prevents duplicates.

### Frontend integration gap

**Expected.** The frontend composable `usePaperReviewSelectors.ts` still returns `DEMO_PROVENANCE` when `PAPER_REVIEW_DEMO_FILL` is true. A follow-up PR should: (1) add an API call to `GET /api/automation/proposals/{id}/provenance`, (2) map the response to `ProvenanceRow[]`, (3) flip the flag to prefer backend data when available.

Chris0Jeky:
## Adversarial Review - PR #1039 (paper/1019-provenance)

### Bot Comments Addressed

**MEDIUM (Gemini): Integer truncation for confidence** -- Fixed.
`ProvenanceQueryService.BuildValue` used `(int)(field.Confidence * 100)` which truncates (95.9% -> 95%). Changed to `(int)Math.Round(field.Confidence * 100)` so 95.9% correctly rounds to 96%. Added a dedicated test (`BuildValue_RoundsConfidenceInsteadOfTruncating`) that asserts 0.959 displays as "96%", not "95%".

**P2 (Codex): Missing FK for ProposalProvenance** -- Fixed via new migration.
`ProposalProvenances.ProposalId` had a unique index but no foreign key constraint to `AutomationProposals`, allowing orphaned rows. Added `HasOne<AutomationProposal>().WithOne().HasForeignKey<ProposalProvenance>(pp => pp.ProposalId).OnDelete(Cascade)` to `ProposalProvenanceConfiguration`, then generated a new migration (`20260425232031_AddProposalProvenanceForeignKey`). The existing migration was NOT edited.

### Additional Checks

| Area | Status | Notes |
|------|--------|-------|
| Icon map completeness | OK | 26 entries covering all common field names, case-insensitive lookup via `StringComparer.OrdinalIgnoreCase`, fallback to `DefaultIcon` for unknown fields. Solid. |
| Weight mapping edge cases | OK | "excluded" weight is documented as never generated from `ProvenanceField` data (frontend-injected only). All `ProvenanceKind` enum values (Extractive, Inferred) are handled. Boundary at 0.7 is correct and tested. |
| Authorization | OK | `GetProposalProvenance` endpoint uses `AuthorizeProposalAsync` which calls `CanReadBoardAsync` for board-scoped proposals and falls back to ownership check for unscoped ones. Same pattern as `GetProposal` and `GetProposalDiff`. |
| Null handling | OK | When no provenance exists, service returns `Result.Success` with empty list. Controller returns 200 with `[]`. |
| EF entity configurations | Fixed | ProvenanceField->ProposalProvenance and EvidenceLink->ProvenanceField FKs were correctly configured. ProposalProvenance->AutomationProposal FK was missing and is now fixed. |
| Domain validation | OK | All entities validate inputs thoroughly (empty Guids, string lengths, confidence range, enum validity, extractive quote requirements). |
| DI registration | OK | `IProposalProvenanceRepository` registered in Infrastructure DI, `IProvenanceQueryService` registered in Application services. |

### Commits Pushed

1. `33b3da61` Fix confidence percentage rounding in ProvenanceQueryService
2. `e1ff5d72` Add test for confidence rounding behavior in ProvenanceQueryService
3. `dbe8af69` Add FK from ProposalProvenance to AutomationProposal in EF config
4. `f0dddb03` Add EF Core migration for ProposalProvenance FK to AutomationProposals

### Verification

- `dotnet build backend/Taskdeck.sln -c Release` -- 0 errors, 12 warnings (pre-existing)
- `dotnet test --filter "FullyQualifiedName~Provenance"` -- all pass (72 application, 54 domain, 7 API)

Chris0Jeky:

### IMPORTANT

**1. `MapWeight` silently treats unknown `ProvenanceKind` values as Extractive (Confidence: 88)**
Only checks for `ProvenanceKind.Inferred` and falls through to extractive-style bucketing for everything else. If a third variant is added (e.g., `Excluded`, `Transformed`), it silently gets "primary"/"contextual" classification.

**Fix:** Use exhaustive `switch` expression with `_ => throw new ArgumentOutOfRangeException(...)`.

**2. Two-migration FK split creates partial-migration risk (Confidence: 82)**
`AddProvenanceEntities` creates table without FK, `AddProposalProvenanceForeignKey` adds it. If first runs but second fails, orphaned rows possible. Since neither has shipped, squash into one migration.

**3. Weight strings are raw magic strings (Confidence: 81)** — `"primary"`, `"contextual"`, `"inferred"` with no shared constants. Frontend/backend contract is fragile to typos.

**Fix:** Define `ProvenanceWeights.Primary`, etc. as static constants.

**4. No API integration test for the endpoint (Confidence: 80)** — Application-layer tests exist but no HTTP pipeline test verifying auth enforcement and response shape.

### Verified Non-Issues
- Security: reuses `AuthorizeProposalAsync` with `requireWriteAccess: false` — correct
- `Math.Round` for confidence display (truncation bug fixed)
- Domain validation thorough (ExtractiveQuote presence rules, confidence bounds)
- Cascading deletes correct across entity chain
- Icon map has fallback default

Chris0Jeky:
## Merge readiness update

Merged current `main` into this PR after `#1036` and `#1037` landed. Conflict resolution kept the provenance endpoint/DI/DbContext wiring from this PR and the confidence + daily seal registrations/entities from main.

Docs were corrected so the Paper backend status now reflects `#1031`-`#1039` as delivered/merge-ready and leaves `#1040` pending.

Verification:
- `dotnet build backend/Taskdeck.sln -c Release` - passed (existing warnings only)
- `dotnet test backend/Taskdeck.sln -c Release --no-build --filter FullyQualifiedName~ProvenanceQuery|FullyQualifiedName~ConfidenceBreakdown|FullyQualifiedName~Seal|FullyQualifiedName~DailySnapshot` - passed, 136 tests
- `node scripts/check-docs-governance.mjs` - passed
- GitHub CI required lanes - all passing

Current state: mergeable and clean.

Issue #977, opened 2026-04-25
RFAI-05: Confidence pipeline and Review evidence section


Chris0Jeky:
## Context

Week 5 from `taskdeck-12-week-roadmap-v4.md`.

Parent: #972
Depends on: #976

This issue makes proposal confidence and evidence visible enough for review-first automation to feel trustworthy, not opaque.

## Scope

- Combine verbalized confidence, provider confidence/logprob signals where available, and attribution/provenance verification scores into field-level confidence.
- Add Review UI evidence sections using source spans and EvidenceLink reason chips.
- Add self-consistency only for high-criticality low-confidence decisions, with explicit quotas/cost guardrails.
- Add Brier-score measurement on confidence buckets for the golden set.
- Preserve existing readable proposal cards, sticky actions, and board-centered review flow; this is an evidence depth pass, not another Review decomposition.

## Acceptance Criteria

- [ ] Review cards expose why a proposal exists, with source/evidence links where available.
- [ ] Confidence buckets are generated deterministically and can be scored against the golden set.
- [ ] High-criticality low-confidence self-consistency is gated by quotas and never silently applies work.
- [ ] Manual inbox-triage review evidence path is documented and can be tested.
- [ ] Brier score target and current result are reported by the eval harness.

## Suggested Verification

- Frontend Review component tests
- Backend confidence aggregation tests
- Golden eval smoke including confidence-bucket scoring
- Manual Review flow check recorded in docs when behavior lands

Chris0Jeky:
Delivered in PR #1062 (Paper Review deep-dive wired to backend APIs) merged 2026-05-16

PR #1062, opened 2026-05-16
RFAI-05: Wire Paper Review deep-dive to real backend APIs

Chris0Jeky:
## Summary
- Add `proposalDeepReviewApi.ts` with typed HTTP client for all 6 deep-review endpoints (provenance, confidence, side-effects, conflicts, history, similar-past)
- Rewrite `usePaperReviewSelectors` composable to fetch real data from backend instead of hardcoded demo stubs
- Graceful per-endpoint failure handling via `Promise.allSettled` — individual endpoint failures don't break the whole Review surface
- Generation counter prevents stale API responses from overwriting current proposal data on rapid switching
- DTO-to-frontend type mapping normalizes backend enum casing (PascalCase → lowercase)

## Test plan
- [x] 8 API module tests (endpoint calls, ID encoding, error propagation)
- [x] 10 composable tests (empty state, fetch on activate, weight/tone/status mapping, stale discard, failure isolation, apply rate computation)
- [x] 45 existing Paper Review view tests continue to pass
- [x] `vue-tsc --noEmit` typecheck passes clean
- [ ] CI (pending)

Refs #977

### DIFF
--- a/frontend/taskdeck-web/src/api/proposalDeepReviewApi.ts
+++ b/frontend/taskdeck-web/src/api/proposalDeepReviewApi.ts
@@ -62,49 +62,59 @@ export interface SimilarPastResultDto {
   applyRate: number
 }
 
+export interface RequestOptions {
+  signal?: AbortSignal
+}
+
 function encodedId(id: string): string {
   return encodeURIComponent(id)
 }
 
 export const proposalDeepReviewApi = {
-  async getProvenance(proposalId: string): Promise<ProvenanceRowDto[]> {
+  async getProvenance(proposalId: string, options?: RequestOptions): Promise<ProvenanceRowDto[]> {
     const { data } = await http.get<ProvenanceRowDto[]>(
       `/automation/proposals/${encodedId(proposalId)}/provenance`,
+      { signal: options?.signal },
     )
     return data
   },
 
-  async getConfidence(proposalId: string): Promise<ConfidenceBreakdownDto> {
+  async getConfidence(proposalId: string, options?: RequestOptions): Promise<ConfidenceBreakdownDto> {
     const { data } = await http.get<ConfidenceBreakdownDto>(
       `/automation/proposals/${encodedId(proposalId)}/confidence`,
+      { signal: options?.signal },
     )
     return data
   },
 
-  async getSideEffects(proposalId: string): Promise<ProposalSideEffectsDto> {
+  async getSideEffects(proposalId: string, options?: RequestOptions): Promise<ProposalSideEffectsDto> {
     const { data } = await http.get<ProposalSideEffectsDto>(
       `/automation/proposals/${encodedId(proposalId)}/side-effects`,
+      { signal: options?.signal },
     )
     return data
   },
 
-  async getConflicts(proposalId: string): Promise<ConflictRowDto[]> {
+  async getConflicts(proposalId: string, options?: RequestOptions): Promise<ConflictRowDto[]> {
     const { data } = await http.get<ConflictRowDto[]>(
       `/automation/proposals/${encodedId(proposalId)}/conflicts`,
+      { signal: options?.signal },
     )
     return data
   },
 
-  async getHistory(proposalId: string): Promise<CardHistoryRowDto[]> {
+  async getHistory(proposalId: string, options?: RequestOptions): Promise<CardHistoryRowDto[]> {
     const { data } = await http.get<CardHistoryRowDto[]>(
       `/automation/proposals/${encodedId(proposalId)}/history`,
+      { signal: options?.signal },
     )
     return data
   },
 
-  async getSimilarPast(proposalId: string): Promise<SimilarPastResultDto> {
+  async getSimilarPast(proposalId: string, options?: RequestOptions): Promise<SimilarPastResultDto> {
     const { data } = await http.get<SimilarPastResultDto>(
       `/automation/proposals/${encodedId(proposalId)}/similar-past`,
+      { signal: options?.signal },
     )
     return data
   },
--- a/frontend/taskdeck-web/src/api/proposalDeepReviewApi.ts
+++ b/frontend/taskdeck-web/src/api/proposalDeepReviewApi.ts
@@ -1,0 +1,111 @@
+import http from './http'
+
+export interface ProvenanceRowDto {
+  icon: string
+  key: string
+  value: string
+  weight: string
+}
+
+export interface ConfidenceComponentDto {
+  key: string
+  value: number
+}
+
+export interface ConfidenceBreakdownDto {
+  overall: number
+  components: ConfidenceComponentDto[]
+  note: string | null
+  threshold: number
+  meetsThreshold: boolean
+}
+
+export interface SideEffectRowDto {
+  key: string
+  value: string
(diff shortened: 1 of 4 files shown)


Chris0Jeky:
## Adversarial Review — PR #1062 (Deep Review API wiring)

No existing comments found. First review.

---

### MEDIUM-1: Shared mutable empty-state arrays risk cross-instance corruption

**File:** `src/composables/usePaperReviewSelectors.ts` lines 75-78

The module-level constants `EMPTY_PROVENANCE`, `EMPTY_CONFLICTS`, `EMPTY_HISTORY`, `EMPTY_SIMILAR` are plain arrays assigned directly to reactive `ref()` values. If any downstream consumer (or a future refactor) accidentally mutates these arrays (e.g. `.push()`, `.splice()`), the corruption leaks to every composable instance because they all reference the same object.

**Fix:** Freeze the arrays with `Object.freeze()` or use factory functions that return fresh arrays. The same concern applies to `EMPTY_SIDE_EFFECTS` and `EMPTY_CONFIDENCE`.

---

### MEDIUM-2: No request cancellation on proposal change (wasted bandwidth / race edge case)

**File:** `src/composables/usePaperReviewSelectors.ts` lines 200-207

When the user rapidly switches proposals, old requests are not cancelled — they run to completion (6 HTTP round trips per switch). The generation counter correctly prevents stale data from being written, but:
- Wasted bandwidth on slow connections
- If a token expires mid-flight, the stale 401 response from a no-longer-relevant request could trigger the global auth redirect in `http.ts`

**Fix:** Use an `AbortController` that is aborted when the watcher re-fires. Pass the signal to each `http.get()` call. The existing retry interceptor already handles `AbortSignal` correctly.

---

### MEDIUM-3: `ConfidenceBreakdownDto.overall` value not clamped to [0, 1]

**File:** `src/composables/usePaperReviewSelectors.ts` line 130 (`mapConfidence`)

The `overall` value from the backend is passed through without validation. The `PaperConfidenceDial` component and bar rendering assume a 0-1 range. If the backend ever returns a value outside this range (e.g. due to a bug or a different scoring scale), the dial would render incorrectly (negative widths, >100% widths).

**Fix:** Clamp the value: `overall: Math.max(0, Math.min(1, dto.overall))`. Similarly clamp `dto.components[*].value`.

---

### LOW-1: No loading state exposed to consumers

**File:** `src/composables/usePaperReviewSelectors.ts`

The composable exposes no `loading` ref. When a user selects a proposal, all 6 sections show empty state until responses arrive (could be 1-3 seconds on slow connections). Consumers cannot distinguish "still loading" from "backend returned empty data."

**Fix:** Add a `loading: Ref<boolean>` that is set `true` before `Promise.allSettled` and `false` after (guarded by generation check). Return it in the `PaperReviewSelectors` interface. This can be a follow-up if shipping urgency dictates.

---

### LOW-2: Test double-mock in provenance weight test is confusing

**File:** `src/tests/composables/usePaperReviewSelectors.spec.ts` lines 97-105

The provenance mock is set, then `mockAllEndpointsEmpty()` overwrites it, then it's set again. The second setup (lines 102-104) is the one that actually takes effect. The first is dead code.

**Fix:** Remove lines 97-100 or move `mockAllEndpointsEmpty()` before the specific mock override.

---

### LOW-3: `EMPTY_SIDE_EFFECTS` has hardcoded user-facing text

**File:** `src/composables/usePaperReviewSelectors.ts` lines 79-87

The fallback `EMPTY_SIDE_EFFECTS` contains hardcoded strings like "6 hours · single keystroke" and "Undo restores the prior state. Nothing is lost." These are user-visible when the side-effects endpoint fails. If the product copy changes, this will be a maintenance trap.

**Impact:** Low — this was pre-existing from the demo stub era. Noting for awareness.

---

### Summary

| Severity | Count | Action |
|----------|-------|--------|
| MEDIUM | 3 | Fix in this PR |
| LOW | 3 | Fix LOW-1 and LOW-2 in this PR; LOW-3 is cosmetic/pre-existing |

All MEDIUM and actionable LOW findings will be fixed in follow-up commits.

Chris0Jeky:
## Fix Evidence — Adversarial Review Findings

All findings from the review have been addressed in 3 focused commits:

### Commit `b5563ea5` — MEDIUM-2 (API layer)
**Add AbortSignal support to proposalDeepReviewApi**
- Each endpoint method now accepts an optional `{ signal?: AbortSignal }` parameter
- Signal is forwarded to `http.get()` so Axios will abort the request when the controller fires
- Non-breaking: callers without options get `{ signal: undefined }` which Axios ignores

### Commit `32c78db8` — MEDIUM-1, MEDIUM-2, MEDIUM-3, LOW-1 (composable)
**Harden usePaperReviewSelectors: freeze, clamp, cancel, loading**
- **MEDIUM-1**: All empty-state constants are now `Object.freeze()`-d at module level. Any accidental `.push()` will throw at runtime.
- **MEDIUM-2**: Added `AbortController` lifecycle — previous requests are aborted when the watcher re-fires on proposal change. Eliminates wasted bandwidth and stale-request auth redirect risk.
- **MEDIUM-3**: Added `clamp01()` helper; `overall` and each `component.value` are clamped to `[0, 1]` in `mapConfidence()`. The dial/bar rendering is now safe against out-of-range backend values.
- **LOW-1**: New `loading: ComputedRef<boolean>` exposed in `PaperReviewSelectors`. Set `true` before `Promise.allSettled`, `false` after (guarded by generation). Consumers can now show loading skeletons.

### Commit `60483b02` — LOW-2 + test alignment
**Update tests for AbortSignal param and fix double-mock**
- API tests updated to assert `{ signal: undefined }` second argument
- Composable integration test uses `expect.objectContaining({ signal: expect.any(AbortSignal) })`
- Removed dead duplicate mock setup in the provenance weight test

### Verification
- `npm run typecheck` — clean (0 errors)
- `npx vitest --run` on both test files — 18/18 passing
- `npm run lint` — 0 errors (6 pre-existing warnings in unrelated files)

### Remaining (LOW-3: hardcoded fallback copy)
Noted but not fixed — this is pre-existing from the demo stub era and affects only the error-fallback path. If the product copy changes, a follow-up can extract these to a constants file or i18n.

Chris0Jeky:
## Adversarial Review — Round 2

**Verdict: CLEAN after 2 rounds of adversarial review. No issues at confidence >= 80.**

Traced all concern areas:
- **Race conditions**: Dual-guard (generation counter + AbortController) eliminates all race paths. Exhaustive sequence analysis confirms correctness.
- **Memory leaks**: AbortController reassignment with old one aborted — no orphan references.
- **DTO mapping null handling**: All paths covered; null → undefined for optional fields, TypeError on backend contract violations gets caught by allSettled.
- **XSS**: Data flows to Vue template interpolation (auto-escaped). No `v-html` usage.
- **Type safety**: All DTO→internal mappings handle case normalization with safe defaults. Extra fields silently ignored.
- **allSettled error propagation**: Abort errors caught as rejections → empty fallback. Correct.
- **Generation counter overflow**: Safe up to 2^53 (285K years at 1/ms). Not a concern.
- **isLoading edge case**: Traced null-proposal + stale-settle sequence — final state always correct.

This PR is ready for merge.

PR #1073, opened 2026-05-16
Add egress disclosure API endpoint (RFAI-08)

Chris0Jeky:
## Summary

- **Fix DI bug**: `EgressRegistry` singleton was resolved with empty `IEnumerable<EgressEntry>` due to .NET DI greedy constructor selection. Changed to explicit instance registration.
- **Add GitHub seed entry**: `api.github.com` registered as known egress destination for `GitHubConnectorProvider`.
- **New endpoint**: `GET /api/privacy/egress` returns all registered outbound data destinations for GP-10 compliance. Requires auth, returns host/payloadCategory/toolOrAgent/dataClassification per destination.
- **9 integration tests**: Auth check, response shape, OpenAI/Gemini/Sentry/GitHub presence, required fields, count consistency, no-secrets verification.

## Test plan

- [x] All 9 `EgressDisclosureApiTests` pass
- [x] Architecture boundary test passes (controller extends `AuthenticatedControllerBase`)
- [x] Full backend suite green (6,482 tests, 0 failures)
- [ ] Adversarial review round 1
- [ ] Adversarial review round 2

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs
@@ -1,0 +1,47 @@
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Taskdeck.Application.Interfaces;
+using Taskdeck.Application.Services;
+
+namespace Taskdeck.Api.Controllers;
+
+[ApiController]
+[Authorize]
+[Route("api/privacy/egress")]
+[Produces("application/json")]
+public class EgressDisclosureController : AuthenticatedControllerBase
+{
+    private readonly IEgressRegistry _egressRegistry;
+
+    public EgressDisclosureController(IEgressRegistry egressRegistry, IUserContext userContext)
+        : base(userContext)
+    {
+        _egressRegistry = egressRegistry;
+    }
+
+    [HttpGet]
+    [ProducesResponseType(typeof(EgressDisclosureResponse), StatusCodes.Status200OK)]
+    [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+    public IActionResult GetDisclosure()
+    {
+        var entries = _egressRegistry.GetAllEntries();
+
+        var destinations = entries.Select(e => new EgressDestinationDto(
+            e.Host,
+            e.PayloadCategory,
+            e.ToolOrAgentName,
+            e.Classification.ToString())).ToList();
+
+        return Ok(new EgressDisclosureResponse(destinations, destinations.Count));
+    }
+}
+
+public sealed record EgressDisclosureResponse(
+    IReadOnlyList<EgressDestinationDto> Destinations,
+    int TotalCount);
+
+public sealed record EgressDestinationDto(
+    string Host,
+    string PayloadCategory,
+    string ToolOrAgent,
+    string DataClassification);
--- a/backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs
@@ -1,6 +1,5 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
-using Taskdeck.Application.Interfaces;
 using Taskdeck.Application.Services;
 
 namespace Taskdeck.Api.Controllers;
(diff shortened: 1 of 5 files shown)


Chris0Jeky:
## Adversarial Review — Round 1

### Finding 1 — IMPORTANT: Fragile no-secrets test uses naive substring matching

**File**: `backend/tests/Taskdeck.Api.Tests/EgressDisclosureApiTests.cs:155-158`

The `GetDisclosure_NoSecretOrCredentialDataExposed` test checks the **entire** JSON response body for substrings like `"key"`, `"secret"`, `"password"`, `"token"`. This will false-positive if any future seed entry has a legitimate `payloadCategory` containing these words (e.g., `"API key rotation events"`, `"OAuth token exchange metadata"`).

**Risk**: Test becomes a maintenance landmine. Developers adding new egress entries with natural language descriptions will hit cryptic assertion failures.

**Fix**: Use case-insensitive matching and check only for patterns that look like actual credentials (e.g., base64 strings, `sk-*` prefixes) rather than English words. Or scope the check to verify that no `appsettings.json` / environment variable values appear in the response.

### Finding 2 — MEDIUM: Unused IUserContext dependency adds coupling

**File**: `backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs:16-17`

The controller inherits `AuthenticatedControllerBase` and accepts `IUserContext` in its constructor purely to satisfy the base class. The endpoint never uses `TryGetCurrentUserId()` or any user-specific logic — all authenticated users see identical data. This adds unnecessary constructor complexity and test setup burden.

**Risk**: Low runtime risk, but violates Interface Segregation — the controller is coupled to identity infrastructure it doesn't need.

**Fix**: Add `EgressDisclosureController` to the `AllowedControllerBaseTypes` set in `ApiControllerBoundaryTests` (alongside `AuthController`, `HealthController`, `TelemetryController`), and revert to direct `ControllerBase` inheritance. The `[Authorize]` attribute is sufficient for access control.

### Finding 3 — LOW: Public parameterized constructor remains a DI trap

**File**: `backend/src/Taskdeck.Application/Services/EgressRegistry.cs:28`

The `EgressRegistry(IEnumerable<EgressEntry> entries)` constructor is public. The DI bug was fixed by using explicit instance registration, but if a future developer changes the registration back to `AddSingleton<IEgressRegistry, EgressRegistry>()`, the greedy constructor bug silently returns. The parameterized constructor should be `internal` to prevent accidental DI resolution.

**Risk**: Low (regression risk, not runtime). A future refactor could reintroduce the empty-registry bug.

**Fix**: Change constructor visibility to `internal` and add `[InternalsVisibleTo("Taskdeck.Application.Tests")]` if needed for unit tests.

---

**Summary**: 1 IMPORTANT, 1 MEDIUM, 1 LOW. No CRITICAL findings.

Chris0Jeky:
## R1 Fix Evidence

All 3 findings fixed in commit `af036118`:

| Finding | Severity | Fix |
|---------|----------|-----|
| Fragile no-secrets test | IMPORTANT | Replaced naive substring checks with credential pattern matching (sk-*, ghp_*, base64, Bearer). No false-positives on natural language descriptions. |
| Unused IUserContext dependency | MEDIUM | Reverted to `ControllerBase` inheritance; added `EgressDisclosureController` to `AllowedControllerBaseTypes` in arch test. |
| Public parameterized constructor DI trap | LOW | Changed `EgressRegistry(IEnumerable<EgressEntry>)` visibility to `internal`. InternalsVisibleTo already configured for test project. |

**Verification**: `EgressDisclosureApiTests` (9/9), `EgressRegistryTests` (33/33), `ApiControllerBoundaryTests` (2/2) — all green.

Chris0Jeky:
## Adversarial Review — Round 2

Reviewed final state of all 4 changed files after R1 fixes.

### Finding 4 — LOW: Redundant base64 regex on host field

**File**: `backend/tests/Taskdeck.Api.Tests/EgressDisclosureApiTests.cs:165`

The assertion `host.Should().NotMatchRegex(@"[A-Za-z0-9+/]{40,}={0,2}")` can never fail in practice. The `EgressRegistry.ValidateEntry()` already enforces `Uri.CheckHostName() == UriHostNameType.Dns` on all hosts — a valid DNS hostname contains dots (which break the regex) and individual labels are max 63 chars. The assertion is harmless but provides no additional coverage.

**Action**: Won't fix — the test documents intent (no credentials in host field) even if the validation layer makes it unreachable. Acceptable as defense-in-depth.

---

### No new IMPORTANT/MEDIUM/CRITICAL findings in R2.

The R1 fixes are solid:
- Controller correctly extends `ControllerBase` with `[Authorize]` — no user-specific logic needed
- Credential pattern matching in the no-secrets test is sound (sk-*, ghp_*, Bearer, base64)
- Internal constructor prevents DI greedy-constructor regression
- All 6 seed entries have valid DNS hostnames and meaningful classifications
- `Classification.ToString()` produces expected enum names (`MetadataOnly`, `UserContent`)
- Serialization uses default camelCase (matches test property assertions)
- No rate-limiting concern: endpoint returns static data, requires auth, covered by global middleware

**Verdict**: PR is ready for merge after R2. 0 new findings requiring fixes.

Issue #1110, opened 2026-05-29
BE: Add Architecture.Tests guards for controller auth + standardize board-auth placement


Chris0Jeky:
## Context
The 2026-05-29 identity/authz audit found identity is uniformly claims-first, but two structural inconsistencies remain that are easy to regress without a guard.

## Problems
1. **No automated guard** that every non-anonymous controller derives from `AuthenticatedControllerBase` and that every controller action carries `[Authorize]` or `[AllowAnonymous]`. `AuthController` (intentional mixed-auth) relies on per-method `[Authorize]` — a future action added without an attribute would be silently anonymous.
2. **Inconsistent board-authorization placement:** most board controllers authorize in the controller via `EnsureBoardPermissionAsync`, but `MetricsController`, `ForecastController`, `ExportController`, `KnowledgeController`, `NotificationsController` push it into the service layer. Both are safe, but the split makes it easy for a future board endpoint to assume the base helper already ran.

## Fix
- Add `Architecture.Tests` rules:
  - Every concrete controller (except an allowlist of anonymous ones: Health, Auth) derives from `AuthenticatedControllerBase`.
  - Every public action method has `[Authorize]`, `[AllowAnonymous]`, or inherits class-level `[Authorize]`.
- Document the chosen board-authorization convention (recommend controller-level `EnsureBoardPermissionAsync` for board-route endpoints) in a short note.

## Acceptance
- New Architecture.Tests cover both rules and pass.
- Convention documented.

_Seeded from the 2026-05-29 identity/authz consistency audit. Depends conceptually on the AuthenticatedControllerBase migration issue._

PR #1111, opened 2026-05-29
BE: Migrate Telemetry + EgressDisclosure controllers onto AuthenticatedControllerBase

### LINKED ISSUE #1109: BE: Two controllers bypass AuthenticatedControllerBase identity seam
## Context
A controller identity/authz audit (2026-05-29) confirmed identity resolution is uniformly claims-first across all 42 API controllers (zero bypass risks). Two outliers don't inherit the shared identity seam.

## Problem
`TelemetryController` and `EgressDisclosureController` extend plain `ControllerBase` instead of `AuthenticatedControllerBase` (which every other authenticated controller uses). Both are correctly `[Authorize]`d and neither touches user-scoped data today, so **there is no current leak** — but they sidestep the shared `TryGetCurrentUserId`/`IUserContext` seam, so any future endpoint needing a userId risks re-introducing ad-hoc identity resolution (the SEC-20 class of bug).

## Fix
- Make `TelemetryController` and `EgressDisclosureController` extend `AuthenticatedControllerBase`.
- No behavior change expected; existing tests should stay green.

## Acceptance
- Both controllers derive from `AuthenticatedControllerBase`.
- Backend build + tests green.

_Seeded from the 2026-05-29 identity/authz consistency audit._

Chris0Jeky:
Resolves #1109.

Two controllers extended plain `ControllerBase` while every other authenticated controller uses the shared `AuthenticatedControllerBase` identity seam. Both are `[Authorize]`d and touch no user-scoped data today, so there's no leak — but they sidestepped the shared `TryGetCurrentUserId`/`IUserContext` helper, risking ad-hoc identity resolution if a future endpoint needs a userId.

## Change
- `TelemetryController` and `EgressDisclosureController` now extend `AuthenticatedControllerBase`, injecting `IUserContext` and passing it to `base(...)`. No behavior change.

## Verification
- `dotnet build backend/src/Taskdeck.Api -c Release` → 0 errors/warnings.
- `TelemetryApiTests`, `TelemetryConfigurationTests`, `EgressDisclosureApiTests` — 32/32 pass (auth behavior unchanged).

Part of the 2026-05-29 identity/authz consistency hardening; #1110 adds Architecture.Tests guards to prevent regressions.

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/EgressDisclosureController.cs
@@ -1,5 +1,6 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Taskdeck.Application.Interfaces;
 using Taskdeck.Application.Services;
 
 namespace Taskdeck.Api.Controllers;
@@ -8,11 +9,12 @@ namespace Taskdeck.Api.Controllers;
 [Authorize]
 [Route("api/privacy/egress")]
 [Produces("application/json")]
-public class EgressDisclosureController : ControllerBase
+public class EgressDisclosureController : AuthenticatedControllerBase
 {
     private readonly IEgressRegistry _egressRegistry;
 
-    public EgressDisclosureController(IEgressRegistry egressRegistry)
+    public EgressDisclosureController(IEgressRegistry egressRegistry, IUserContext userContext)
+        : base(userContext)
     {
         _egressRegistry = egressRegistry;
     }
--- a/backend/src/Taskdeck.Api/Controllers/TelemetryController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/TelemetryController.cs
@@ -1,5 +1,6 @@
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Taskdeck.Application.Interfaces;
 using Taskdeck.Application.Services;
 
 namespace Taskdeck.Api.Controllers;
@@ -11,7 +12,7 @@ namespace Taskdeck.Api.Controllers;
 [ApiController]
 [Route("api/telemetry")]
 [Authorize]
-public class TelemetryController : ControllerBase
+public class TelemetryController : AuthenticatedControllerBase
 {
     private readonly ITelemetryEventService _telemetryEventService;
     private readonly SentrySettings _sentrySettings;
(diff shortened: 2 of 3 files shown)


Chris0Jeky:
## Adversarial Review 1 of 2 (self) — PR #1111

**Scope:** the 2-controller base-class migration.

**Risk surface examined:**
- **DI resolution:** `IUserContext` is the same dependency injected into all 40 other authenticated controllers, registered (scoped) in the API composition root — adding it to these two constructors resolves identically. Build is clean; the 32 Telemetry/EgressDisclosure tests (which run through the real DI container via the web factory) pass.
- **Behavior:** `AuthenticatedControllerBase` only adds protected helpers (`TryGetCurrentUserId`, `EnsureBoardPermissionAsync`); it changes no routing, filters, or response behavior. `[Authorize]`/`[AllowAnonymous]` attributes are untouched, so `GetConfig`'s anonymous access and the egress endpoint's auth are preserved.
- **Direct construction:** no unit test constructs these controllers with `new` (the 32 tests are integration-style), so the added constructor param breaks nothing.

**Verdict:** No blocking issues. Independent Review 2 in progress.

Chris0Jeky:
## Adversarial Review 2 of 2 (independent) + Fix Evidence

Independent reviewer raised one **HIGH** and one **MEDIUM**:

**HIGH — stale architecture-test allowlist (fixed).** `ApiControllerBoundaryTests.AllowedControllerBaseTypes` still listed `TelemetryController` and `EgressDisclosureController`. Since those skip the `ProtectedControllers_ShouldDeclareAuthorizeAttribute` guard, a future removal of `[Authorize]` from either would go undetected. Now that both extend `AuthenticatedControllerBase` with class-level `[Authorize]`, they no longer need the exception.
- **Fix:** removed both from the allowlist (kept `AuthController` = per-method auth, `HealthController` = anonymous, with an explanatory comment). The two controllers are now actively validated by the guard.
- **Verify:** `ApiControllerBoundaryTests` — 2/2 pass (both controllers now checked and passing).

**MEDIUM — `IUserContext` injected but unused (intentional, kept).** This is the deliberate goal of #1109/#1110: route all authenticated controllers through the single `AuthenticatedControllerBase`/`IUserContext` identity seam so no controller does ad-hoc identity resolution. The injection is the seam, not dead weight — and the now-active architecture guard enforces the convention. DI lifetime confirmed compatible (`IUserContext` scoped, matches scoped controllers).

All other checklist items (DI lifetime, base-ctor match, `[AllowAnonymous]` precedence on GetConfig, no direct `new` construction, namespace) were reviewed CLEAN. Both findings addressed.

Issue #760, opened 2026-04-04
TECH-DEBT: EF Core 9.x used on net8.0 target framework


Chris0Jeky:
## Problem

The project targets `net8.0` but uses `Microsoft.EntityFrameworkCore 9.0.14`. EF Core 9 is designed for .NET 9 and is outside the .NET 8 support matrix. While it works today, this creates risk of subtle runtime issues and means we're outside the tested/supported configuration.

## Options

1. **Upgrade TFM to net9.0** — aligns with EF Core version, gives access to .NET 9 features
2. **Pin EF Core to 8.x** — keeps the supported configuration, may require reverting any 9.x-specific API usage

## Additional Dependency Issues

While auditing, also found:
- `Taskdeck.Infrastructure.csproj` references `Microsoft.AspNetCore.Http` version **2.3.9** (from .NET Core 2.x era). On .NET 8 this type is part of the shared framework — the NuGet package reference should be removed.
- `OpenTelemetry.Instrumentation.Http` is at 1.15.0 while all other OTel packages are at 1.15.1 — minor version drift.

## Acceptance Criteria

- [ ] Decide: upgrade to net9.0 or pin EF Core to 8.x
- [ ] Remove stale `Microsoft.AspNetCore.Http` 2.3.9 NuGet reference from Infrastructure
- [ ] Align OTel package versions
- [ ] All tests pass after changes

PR #1112, opened 2026-05-29
chore: ignore EF Core major bumps in dependabot (stop recurring 8/9 split)

Chris0Jeky:
## Why
Dependabot's `dotnet-minor-patch` group has now twice (#1102, and again #1106) proposed `Microsoft.EntityFrameworkCore` **9.0.16** while leaving `.Sqlite`/`.Design` on **8.0.27**. That version desync reintroduces the ambiguous `ExecuteDeleteAsync` compile break each cycle, requiring a manual pin every time.

The project intentionally stays on the EF Core **8.x** line (per #760/#767). Until a deliberate whole-stack EF migration, dependabot should not propose EF Core majors.

## Change
Adds an `ignore` rule to the NuGet update config for `version-update:semver-major` on `Microsoft.EntityFrameworkCore`, `.Sqlite`, and `.Design`. Minor/patch bumps within 8.x still flow normally.

## Verification
- `dependabot.yml` parses as valid YAML.
- Indentation matches sibling keys (`groups`, `commit-message`) under the `/backend` nuget entry.

Durable follow-up to the per-PR EF pins on #1102 and #1106.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,6 +26,19 @@ updates:
           - minor
           - patch
     # Let major versions come as individual PRs for explicit review
+    ignore:
+      # The project intentionally pins Entity Framework Core to the 8.x line
+      # (see #760/#767). Dependabot repeatedly proposes EF Core 9.x for the main
+      # package while the Sqlite/Design providers stay on 8.x, which desyncs the
+      # providers and reintroduces the ambiguous ExecuteDeleteAsync break
+      # (fixed in #1102 and again in #1106). Block EF Core major bumps until the
+      # project deliberately migrates the whole EF stack together.
+      - dependency-name: "Microsoft.EntityFrameworkCore"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "Microsoft.EntityFrameworkCore.Sqlite"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "Microsoft.EntityFrameworkCore.Design"
+        update-types: ["version-update:semver-major"]
     commit-message:
       prefix: "deps(nuget)"
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,14 +25,17 @@ updates:
         update-types:
           - minor
           - patch
-    # Let major versions come as individual PRs for explicit review
+    # Let other major versions come as individual PRs for explicit review.
     ignore:
-      # The project intentionally pins Entity Framework Core to the 8.x line
-      # (see #760/#767). Dependabot repeatedly proposes EF Core 9.x for the main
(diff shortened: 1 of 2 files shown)


Chris0Jeky:
## Adversarial Review 1 of 2 (self) — PR #1112

**Scope:** dependabot `ignore` rule addition (config-only).

**Verified:**
- **Schema correctness:** `ignore` is a valid top-level key on a `updates[]` entry; each item uses the documented `dependency-name` + `update-types: ["version-update:semver-major"]` form. `dependency-name` values are exact package IDs (no wildcard needed since the three EF packages are listed explicitly).
- **Scope of the block:** only `semver-major` is ignored — **8.x minor/patch security/bugfix bumps still flow** for all three EF packages. It does not touch `EntityFrameworkCore.Tools` (already on a separate 10.x line) or any non-EF package.
- **Indentation:** `ignore:` is a sibling of `groups:`/`commit-message:` under the `/backend` nuget entry (4-space). YAML parses valid (`yaml.safe_load` OK).
- **No over-block:** the rule is on the `/backend` nuget ecosystem only; npm and actions configs untouched.

**Verdict:** No blocking issues. CI workflow-lint will independently validate the dependabot schema. Second review in progress.

Chris0Jeky:
## Adversarial Review 2 of 2 (independent) + Fix Evidence

Independent reviewer confirmed schema/enum/indentation/scope/grouping/package-IDs/YAML all **correct**, and raised one **MEDIUM** + minors:

**M1 — comment overstated the goal (fixed).** `Microsoft.EntityFrameworkCore.Tools` is intentionally on its own 10.x line (design-time-only, `PrivateAssets`) and isn't part of the runtime compile break, so it's correctly NOT in the ignore list — but the comment claimed to keep "the whole EF stack" on 8.x.
- **Fix:** reworded the comment to scope it to the three *runtime* EF Core packages and explicitly note Tools is tracked separately on 10.x.

**L2 — adjacent comment contradiction (fixed).** Changed `# Let major versions...` → `# Let other major versions...` so it no longer reads as contradicting the EF ignore block directly below it.

**L3 — discoverability (fixed).** Added a "Pinned / version-capped dependencies" table to `docs/ops/DEPENDENCY_UPDATE_POLICY.md` (the policy doc referenced in the dependabot.yml header) recording the EF 8.x cap, its reason (#1102/#1106), and the removal condition.

**L1 — `Npgsql.EntityFrameworkCore.PostgreSQL` (test-only, 9.x):** intentionally not covered — it's an independent vendor provider whose majors don't track Microsoft's EF Core, so capping it here would be wrong. Noted for the record.

**Verification:** `yaml.safe_load` valid. All findings addressed.

PR #1146, opened 2026-05-31
Proposal create: return 409 for duplicate idempotency key (fix #1125 500s)

### LINKED ISSUE #1125: Harden /api/automation/proposals against malformed input (HTTP 500 → 400/422)
## Context
Two backend tests currently document confirmed HTTP 500 bugs on the product's core automation endpoint as *expected* behaviour, via `[Fact(Skip=...)]`.

## Evidence
- `backend/tests/Taskdeck.Api.Tests/ProposalOperationsAdversarialTests.cs:199` (`KnownBug_XssInActionType_Causes500`) and `:217` (`KnownBug_DeepNestedParameters_Causes500`) assert `StatusCode == 500` for `<script>` in `actionType` and nested `parameters` JSON.
- Root cause: `AutomationProposalService.CreateProposalAsync` (`backend/src/Taskdeck.Application/Services/AutomationProposalService.cs:53-100`) wraps the create path in `try { ... } catch (DomainException)` only. The domain op constructor (`AutomationProposalOperation.cs:22-50`) throws `DomainException` for empties (→400) but does not validate `actionType` against an allowlist or bound `parameters` JSON; any non-`DomainException` (or downstream `SaveChanges`/serialization fault) escapes as 500.

## Acceptance criteria
- [ ] Validate operations defensively (actionType/targetType allowlist or normalization; `parameters` parsed + bounded depth/size) returning `Result.Failure` → 400/422 for malformed input; convert unexpected exceptions to a safe 4xx with logging.
- [ ] Verify reproduction first (un-skip → run → observe), then flip both tests to assert `StatusCode < 500` (ideally 400/422) and remove `Skip`.
- [ ] Add an adversarial-input regression case.

_Source: deep audit proposal-endpoint-robustness (HIGH); coordinator-verified the narrow catch._

Chris0Jeky:
## Summary
Addresses #1125 (the duplicate-idempotency-key 500) and corrects its premise. The two `[Fact(Skip)]` "known 500 bug" tests were **misdiagnosed**.

### What I found (empirically verified)
- The `DeepNestedParameters` case does **not** reproduce — nested parameter JSON is handled fine (`<500`).
- The `MalformedActionType` case **passes when run alone** but **deterministically 500s when run after** another proposal-creating test (3/3). Root cause: both skipped tests reused `idempotencyKey: "key1"`, and `AutomationProposalOperation.IdempotencyKey` has a **global unique index** (`AutomationProposalOperationConfiguration.cs:54-55`). On the shared test DB the second insert hit a `DbUpdateException` (a non-`DomainException`) that escaped `CreateProposalAsync`'s `catch (DomainException)` → unhandled → 500. So it was an **idempotency-key collision surfacing as a 500**, not an XSS/`actionType` validation gap.

### Fix
- **Product** (`UnitOfWork.cs`): translate an `AutomationProposalOperations.IdempotencyKey` unique violation into `DomainException(ErrorCodes.Conflict)` in `SaveChangesAsync`, mirroring the existing `IsProposalRevisionUniqueViolation` handler. The create endpoint now returns **409 Conflict** (via `ResultExtensions` `Conflict → 409`) instead of 500 — which is the correct semantics for a duplicate idempotency key.
- **Tests**: un-skip both tests with **unique** keys (proving unusual `actionType` and deeply nested params don't 500), and add `DuplicateIdempotencyKey_ShouldReturnConflictNot500` asserting the new 409.

### Verification
- `ProposalOperationsAdversarialTests` full class: **38/38 pass** (was 36 + 2 skipped).
- Regression: `~Proposal | ~Capture` in Api.Tests: **300/300 pass**.
- The `UnitOfWork` catch only matches the specific IdempotencyKey unique-violation message, so other DB errors still surface normally (no silent-failure widening).

### Scope note on #1125 (updated 2026-05-31 after pass-2 review)
This PR fixes the **500 root cause** but does **not** implement #1125's acceptance criterion #1 (create-time `actionType`/`targetType` allowlist + bounded `parameters` JSON → 400/422). So this PR is intentionally **not** auto-closing #1125 — **Refs #1125**, which stays open for the create-time defensive-validation hardening, landing in a separate focused PR. Idempotency-key contract follow-ups (per-owner scoping + replay-vs-409 ADR) are tracked in #1149.

### DIFF
--- a/backend/src/Taskdeck.Infrastructure/Repositories/UnitOfWork.cs
+++ b/backend/src/Taskdeck.Infrastructure/Repositories/UnitOfWork.cs
@@ -350,12 +350,14 @@ public class UnitOfWork : IUnitOfWork
         if (exception.InnerException is null)
             return false;
 
-        return exception.InnerException.Message.Contains(
-            "AutomationProposalOperations.IdempotencyKey",
-            StringComparison.OrdinalIgnoreCase)
-            || exception.InnerException.Message.Contains(
-                "IX_AutomationProposalOperations_IdempotencyKey",
-                StringComparison.OrdinalIgnoreCase);
+        var message = exception.InnerException.Message;
+
+        // Match UNIQUE violations only. The bare column name also appears in NOT NULL
+        // violation messages ("NOT NULL constraint failed: AutomationProposalOperations.IdempotencyKey"),
+        // so require the UNIQUE qualifier (SQLite) or the unique index name (other providers).
+        return (message.Contains("UNIQUE constraint failed", StringComparison.OrdinalIgnoreCase)
+                && message.Contains("AutomationProposalOperations.IdempotencyKey", StringComparison.OrdinalIgnoreCase))
+            || message.Contains("IX_AutomationProposalOperations_IdempotencyKey", StringComparison.OrdinalIgnoreCase);
     }
 
     private bool TryResolveDuplicateDailySnapshotConflicts(DbUpdateException exception)
--- a/backend/src/Taskdeck.Infrastructure/Repositories/UnitOfWork.cs
+++ b/backend/src/Taskdeck.Infrastructure/Repositories/UnitOfWork.cs
@@ -151,6 +151,13 @@ public class UnitOfWork : IUnitOfWork
                     "Proposal revision was created by another session. Refresh and retry your edit.",
                     ex);
             }
+            catch (DbUpdateException ex) when (IsOperationIdempotencyKeyUniqueViolation(ex))
+            {
+                throw new DomainException(
+                    ErrorCodes.Conflict,
+                    "An automation operation with this idempotency key already exists.",
+                    ex);
+            }
             catch (DbUpdateException ex) when (!resolvedRecoverableUniqueConflict && TryResolveRecoverableUniqueConflicts(ex))
             {
                 resolvedRecoverableUniqueConflict = true;
@@ -335,6 +342,19 @@ public class UnitOfWork : IUnitOfWork
             StringComparison.OrdinalIgnoreCase)
             || exception.InnerException.Message.Contains(
                 "IX_ProposalRevisions_ProposalId_RevisionNumber",
+                StringComparison.OrdinalIgnoreCase);
+    }
+
+    private static bool IsOperationIdempotencyKeyUniqueViolation(DbUpdateException exception)
+    {
+        if (exception.InnerException is null)
+            return false;
+
+        return exception.InnerException.Message.Contains(
+            "AutomationProposalOperations.IdempotencyKey",
+            StringComparison.OrdinalIgnoreCase)
+            || exception.InnerException.Message.Contains(
+                "IX_AutomationProposalOperations_IdempotencyKey",
                 StringComparison.OrdinalIgnoreCase);
     }
 
--- a/backend/tests/Taskdeck.Api.Tests/ProposalOperationsAdversarialTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/ProposalOperationsAdversarialTests.cs
@@ -191,44 +191,73 @@ public class ProposalOperationsAdversarialTests : IClassFixture<TestWebApplicati
             $"Proposal creation returned 500 for expiryMinutes={expiryMinutes}");
     }
 
-    // ─────────────────────── Known 500 bugs (skip-annotated) ───────────────────────
-    // These tests document real API bugs discovered by adversarial testing.
-    // When operations contain unexpected actionType values or nested parameter JSON,
-    // the proposal creation endpoint returns 500 instead of 4xx.
+    // ─────────────────────── Operation input robustness ───────────────────────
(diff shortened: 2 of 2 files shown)


Chris0Jeky:
## Self-review (adversarial)

- **Root cause certainty**: verified by isolation (passes alone) vs ordering (500s after a prior create), and by the unique index in `AutomationProposalOperationConfiguration.cs:54-55`. Not a guess.
- **No silent-failure widening**: `IsOperationIdempotencyKeyUniqueViolation` matches only the specific column/index strings; any other `DbUpdateException` still propagates and surfaces as 500. The catch sits before the generic recoverable-conflict handler, consistent with ordering of the existing handlers.
- **Cross-provider**: matches both the SQLite column form (`AutomationProposalOperations.IdempotencyKey`) and the index name (`IX_..._IdempotencyKey`) used by Postgres-style messages.
- **409 is correct semantics** for a duplicate idempotency key (vs 500). `ResultExtensions` maps `Conflict → 409`.
- **Tests**: unique keys remove the cross-test global-uniqueness coupling; the new test proves the 409 path. 38/38 class + 300/300 proposal+capture regression.

No findings. Not merging — for review.

Chris0Jeky:
## Review fix evidence

**Gemini [MEDIUM] `UnitOfWork.cs` — NOT NULL violations could be misclassified as 409:** valid bug. The bare column name `AutomationProposalOperations.IdempotencyKey` also appears in `"NOT NULL constraint failed: ..."` messages, so the original check could have mapped a NOT NULL violation to 409. **Fixed** by requiring the `"UNIQUE constraint failed"` qualifier (SQLite) or the unique index name (Postgres-style) — column-name match alone is no longer sufficient.

Verification: `ProposalOperationsAdversarialTests` 38/38 still pass (the duplicate-key test still returns 409, since the real UNIQUE message contains the qualifier). Copilot's review errored out and posted no findings.

Chris0Jeky:
## Pass-2 adversarial review

Two independent diverse-lens reviewers (correctness/regression + semantics/security/scope), each adversarially verified against the actual EF model, migration, and consumers. **Verdict: the code is correct and safe to merge** — with one tracking correction (see #1125 note below).

**Prior bot-fix — verified present at HEAD ✅**
Gemini [MEDIUM]: "NOT NULL violations could be misclassified as 409." HEAD's `IsOperationIdempotencyKeyUniqueViolation` requires `("UNIQUE constraint failed" AND "AutomationProposalOperations.IdempotencyKey")` OR the index name. The SQLite NOT NULL message lacks the "UNIQUE constraint failed" qualifier → correctly excluded → still surfaces as 500 (correct, since `IdempotencyKey` is required). Fix is genuinely present, not just claimed.

**Correctness / regression ✅**
- **Catch ordering**: the new `when (IsOperationIdempotencyKeyUniqueViolation(ex))` sits in the same structural slot the pre-existing `IsProposalRevisionUniqueViolation` 409 handler already occupied; SQLite raises exactly one constraint failure per failed `SaveChanges`, and the idempotency message can't also contain another resolver's table/column substring — so no recoverable conflict is shadowed.
- **Postgres branch is live**: migration `...AddAutomationArchiveChatOpsEntities.cs:209` creates `IX_AutomationProposalOperations_IdempotencyKey`; EF convention naming produces exactly that name.
- **Retry loop**: the catch always rethrows `DomainException`, exiting the loop cleanly; unmatched `DbUpdateException`s still propagate (no silent-failure widening).
- **Tests sound**: `DuplicateIdempotencyKey_ShouldReturnConflictNot500` proves first-create succeeds (`<400`, 201) then second-create returns 409; the two un-skipped tests use fresh GUID keys (removing the cross-test global-uniqueness coupling that was the real root cause) and run against a real authenticated path.
- `ErrorCodes.Conflict → 409` mapping confirmed in `ResultExtensions`. Error message is a static string (no key interpolation → no info leak).

### Findings (all LOW; none are code defects)

**1. Scope/tracking — #1125 should not auto-close (handling below).** This PR fixes the duplicate-idempotency-key 500 and un-skips the tests, but #1125's acceptance criterion #1 (create-time `actionType`/`targetType` allowlist + bounded `parameters` JSON → 400/422) is **not** implemented — an arbitrary `actionType`/unbounded `parameters` is still persisted verbatim (mitigated, not eliminated, by apply-time rejection in `OperationHandlerRegistry` under review-first). Per maintainer decision: **this PR will be de-linked from auto-closing #1125** (body edited `Closes #1125` → `Refs #1125`), and the create-time defensive-validation hardening will land as a **separate focused PR**, closing #1125 only when actually met.

**2 & 3. Idempotency contract (out of scope) — seeded as #1149.** (a) The operation `IdempotencyKey` unique index is process-global (cross-tenant) — a hand-crafted simple key can collide across users (production paths all use GUID/hash keys, so LOW). (b) The manual create endpoint has no true replay-returns-original contract (409 is a correct stopgap for a unique-constraint violation, but the contract deserves an ADR). Both pre-existing, neither a regression. Tracked in #1149.

All required checks green; mergeable CLEAN.

Issue #1128, opened 2026-05-30
Harden the test loop for agents (DOM cleanup, non-watch default, single verify command, network guard, parallelism)


Chris0Jeky:
## Context
The suite is large but fragile and hostile to agents — a core reason the project feels "hard to test / hard to let LLMs test".

## Evidence
- `src/tests/setup.ts:22-25` only resets Pinia (no unmount/DOM cleanup); `enableAutoUnmount` used in zero files; `InboxView.spec.ts` mounts 67× with 0 unmounts → cross-test DOM pollution. Root cause of the **MetricsView `role=alert` flake** (`MetricsView.coverage.spec.ts:306`, times out via a fixed double-flush `waitForUi` racing an `onMounted` board auto-load). Confirmed: a fresh `npx vitest --run` produced 1 failure (3661/3662) while PR #1122 evidence claimed 3662.
- `package.json` `"test": "vitest"` → interactive watch (agents hang forever).
- No root "verify everything" command; backend forced serial via `-m:1` over 6,689 tests (real cause = cross-process `FirstRunBootstrapper` race, not per-class DBs).
- `useAnalyticsScript.spec.ts` injects a real external `<script src=plausible.example.com>` under happy-dom (no network guard) → console spam, egress smell.

## Acceptance criteria
- [ ] `enableAutoUnmount(afterEach)` + global `document.body/head` cleanup in `setup.ts`.
- [ ] MetricsView test made deterministic (stub no-boards or `vi.waitFor` on `[role=alert]`); log in failure ledger.
- [ ] `"test": "vitest run"` + `"test:watch": "vitest"`.
- [ ] `scripts/verify.{ps1,sh}` chaining backend + frontend typecheck/unit + fast e2e smoke with one pass/fail; documented in CLAUDE.md/AGENTS.md.
- [ ] Isolate cross-process-stateful tests into a no-parallel xUnit collection; drop blanket `-m:1`.
- [ ] Disable happy-dom JS file loading in `vitest.config.ts`; use a non-routable analytics test host.

_Source: deep audit test-loop-hardening (HIGH); coordinator confirmed the live flake._

Chris0Jeky:
**Re-scope from the 2026-07-02 whole-project analysis** (tracker #1278): keep only the two ACs that pay for themselves before archive — agents do nearly all implementation, so per-PR loop cost is the multiplier on everything remaining.

[...]

PR #1163, opened 2026-06-04
Fix flaky WIP-limit smoke test (await the rejection before asserting the toast)

Chris0Jeky:
## What & why
`column WIP limit should reject additional cards` (`frontend/taskdeck-web/tests/e2e/smoke.spec.ts:290`) is intermittently flaky — it failed on PR #1158's CI (151 passed / 1 failed) on a change (a gitleaks workflow comment) that cannot possibly affect it.

**Root cause:** the test's `addCard(secondCard, { expectVisible: false })` helper did **not** await the create `POST` (it only set up a `waitForResponse` for the success case). So when the test then asserted the transient `text=has reached its WIP limit` toast, it was racing the still-in-flight 409 — under worker CPU contention the toast could auto-dismiss, or not yet render, before the 8s `toBeVisible` window, and the assertion failed with "element(s) not found".

## Fix
- `addCard` now awaits the create response in **both** branches: a 2xx when the card is expected, or the non-2xx rejection (e.g. 409 WIP-limit) when it is not. Awaiting the rejection is what lets a caller assert the rejection toast deterministically.
- The WIP test asserts the (transient) toast **immediately after** that awaited response — before its auto-dismiss — then asserts the card count is 0.
- Only the WIP test uses `expectVisible: false`; the success-path matcher is unchanged, so all other `addCard` callers are unaffected.

## Verification
- ESLint on the spec: clean. Only one `expectVisible: false` caller (the WIP test).
- E2E can't be exercised meaningfully on this dev box (needs the full app + many runs to characterise flakiness); the change is strictly more correct (it adds a wait that previously didn't exist) and is validated by the E2E Smoke CI job. Reasoning is laid out above for adversarial review.

Relates to #1128 (harden the test loop for agents).

### DIFF
--- a/frontend/taskdeck-web/tests/e2e/smoke.spec.ts
+++ b/frontend/taskdeck-web/tests/e2e/smoke.spec.ts
@@ -103,20 +103,28 @@ async function addCard(
   const addCardInput = column.getByPlaceholder('Enter card title...')
   await expect(addCardInput).toBeVisible()
   await addCardInput.fill(cardTitle)
-  // Wait for the create POST to settle in BOTH cases: a success (2xx) when the
-  // card is expected, or the rejection (non-2xx, e.g. 409 WIP-limit) when it is
-  // not. Awaiting the rejection is what lets callers assert the transient
-  // rejection toast without racing the still-in-flight request (the toast can
-  // auto-dismiss before a non-awaited request even returns).
-  const createCardResponse = page.waitForResponse((response) =>
-    response.request().method() === 'POST'
-    && /\/api\/boards\/[a-f0-9-]+\/cards$/i.test(response.url())
-    && (expectVisible ? response.ok() : !response.ok()))
+  // Wait for the create POST to settle (any status) before asserting anything.
+  // Awaiting the response is what lets callers assert the transient rejection
+  // toast without racing the still-in-flight request (the toast can auto-dismiss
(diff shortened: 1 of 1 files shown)


Chris0Jeky:
## Review round — all findings addressed (Gemini + 2 independent adversarial passes)

Both **Gemini Code Assist** (MEDIUM) and the **independent adversarial review** (LOW, verified) converged on the *same* concern, now fixed in `47a43b3e`.

| Finding | Severity | Fix | Commit |
|---|---|---|---|
| Filtering response status inside `page.waitForResponse` (`!response.ok()` / `response.ok()`) means an unexpected API outcome (e.g. WIP enforcement regresses to 2xx) never matches the predicate, so the test hangs to the 45s test-level timeout as a generic "waiting for response" instead of a precise failure | Gemini MEDIUM / adversarial LOW | `addCard` now waits for **any** create POST (15s bound), then asserts the expected ok-ness explicitly: `created.ok()` truthy when the card is expected, falsy (rejection) otherwise. A WIP-enforcement regression now fails fast with a clear message. | `47a43b3e` |

### Verification
- ESLint on `smoke.spec.ts`: clean.
- The success-path matcher no longer over-filters; all existing `addCard` callers (expectVisible default) now additionally assert the create succeeded.
- E2E behaviour is validated by the E2E Smoke CI job (the change strictly adds/bounds a wait and explicit assertions).

Relates to #1128.

Issue #1164, opened 2026-06-04
Serialize cross-process Database.Migrate() (single-owner / advisory lock)


Chris0Jeky:
## Context
Follow-up to #1130. That PR enabled SQLite **WAL + busy_timeout** (the high-value fix for everyday `SQLITE_BUSY`), but #1130's second acceptance criterion — *serialize/skip migration application across processes* — is intentionally left open there because it is a distinct, higher-risk concern that needs cross-process testing.

Each host still calls `Database.Migrate()` on startup with no cross-process coordination:
- `backend/src/Taskdeck.Api/Program.cs` (two call sites)
- `backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs`
- `backend/src/Taskdeck.Cli/Program.cs`

`busy_timeout` now makes a racing migrator **wait** for the lock rather than fail immediately, substantially mitigating the race. But two processes booting simultaneously can still both decide migrations are pending and contend on `__EFMigrationsHistory`.

## Action
Serialize migration application — e.g. a portable file-based advisory lock (a sidecar lock file next to the `Data Source`, opened with exclusive share) wrapping `Database.Migrate()`, or a single-owner model where only the API applies migrations and CLI/MCP skip when already migrated. (.NET named `Mutex` is **not** cross-process on Unix, so avoid it.)

## Acceptance criteria
- [ ] A shared helper serializes `Database.Migrate()` so concurrent startups apply migrations exactly once; applied at all four call sites.
- [ ] CLI/MCP do not race the API to apply migrations.
- [ ] A test exercises two concurrent migrators against one fresh file DB (both succeed, schema applied once).

_Source: #1130 (2026-06-04), second acceptance criterion split out for a focused, well-tested change._

Issue #1130, opened 2026-05-30
SQLite local concurrency: enable WAL + busy_timeout and fix the per-process Migrate() race


Chris0Jeky:
## Context
The product is local-first and explicitly agent-driven, but three processes (API web, MCP HTTP/stdio, CLI) open the **same** SQLite file and each call `Database.Migrate()` on startup with **no WAL mode, no busy_timeout, no PRAGMA** configured anywhere. Running the UI and an MCP agent (and/or the CLI) together will hit `SQLITE_BUSY` "database is locked" — the single most likely thing to break the maintainer's stated agent+UI workflow.

## Evidence
- `backend/src/Taskdeck.Infrastructure/DependencyInjection.cs:34-40` configures `UseSqlite` with only a command timeout; grep for `journal_mode/WAL/busy_timeout/PRAGMA` across `backend/src` returns only comments/Designer false-positives.
- `Database.Migrate()` called in `Api/Program.cs:112` & `:189`, `PipelineConfiguration.cs:34`, `Cli/Program.cs:43`.
- `.gitignore:12-13` references `*.db-wal`/`*.db-shm` — WAL was *intended* but never enabled in code.

## Acceptance criteria
- [ ] Set `journal_mode=WAL` and `busy_timeout` (e.g. 5000ms) on connection open (connection-string + interceptor or `SqliteConnection` PRAGMA).
- [ ] Serialize/skip migration application across processes (startup advisory lock, or migrate only from a single owner process; CLI/MCP should not race the API).
- [ ] Add a concurrency regression test (two contexts writing) and document in `EF_MIGRATION_WORKFLOW.md`.

_Source: deep-audit completeness critic (HIGH; critic ranked this above the perf findings for a local-first app)._

Chris0Jeky:
Closing as substantially delivered. WAL + busy_timeout and WAL-safe dev-sandbox export/import merged in PR #1165 (commit da764b92): SqlitePragmaConnectionInterceptor + Database:BusyTimeoutMilliseconds, checkpoint-before-backup, strict side-file deletion (ACs 1 + 3). The remaining AC2 — cross-process Database.Migrate() serialization — is carried by #1164; the export/import redesign via the SQLite backup API is carried by #1166. Tracked in OUTSTANDING_TASKS.md §C.

PR #1165, opened 2026-06-04
Enable SQLite WAL + busy_timeout for local concurrency (#1130)

Chris0Jeky:
## What & why
Taskdeck is local-first and the **API UI, the MCP server, and the CLI can all open the same SQLite file**, but the `DbContext` was configured with only a command timeout — **no WAL, no `busy_timeout`, no PRAGMAs**. Running the UI and an MCP agent (or the CLI) together is the single most likely thing to break the maintainer's agent+UI workflow with `SQLITE_BUSY` ("database is locked"). The audit ranked this HIGH.

## Change (issue #1130, ACs 1 + 3)
- **`SqlitePragmaConnectionInterceptor`** (a `DbConnectionInterceptor`) runs `PRAGMA journal_mode=WAL` + `PRAGMA busy_timeout=<configured>` on **every** connection open, registered via `AddInterceptors` in `AddInfrastructure`.
  - WAL lets many readers run concurrently with one writer (instead of readers/writers blocking).
  - `busy_timeout` makes a contended writer wait for the lock instead of failing immediately.
  - WAL is persisted in the DB header (idempotent on re-open); `busy_timeout` is per-connection so it's set every open. On `:memory:` WAL is a harmless no-op, so the interceptor is safe for all hosts and tests.
- **New `Database:BusyTimeoutMilliseconds`** setting (default `5000`, range `0–60000`), validated like the existing `CommandTimeoutSeconds`.
- **Tests** (`SqlitePragmaInterceptorTests`): asserts `journal_mode=wal` + `busy_timeout=5000` on connection open, and that two `DbContext`s on the same file write **concurrently** without `SQLITE_BUSY`.
- **Docs**: `CONFIGURATION_REFERENCE` (both Database tables) + `EF_MIGRATION_WORKFLOW` (WAL/busy_timeout note + the still-open per-process Migrate race).

## Scope
- AC #2 (serialize cross-process `Database.Migrate()`) is intentionally split to **#1164** — it's a distinct, higher-risk change needing cross-process testing, and `busy_timeout` already mitigates the race. So this PR **Relates to #1130** (does not close it).

## Verification
- New tests pass; `MigrationBootstrapTests` + `ConnectorEncryptionKeyFailFastTests` (`:memory:` DI) + the new tests: 14 pass — the interceptor coexists with migrations and in-memory DBs.
- Infrastructure builds clean.

Each PR gets 2 independent adversarial reviews before merge; all findings of every severity will be fixed.


Chris0Jeky:
## Review round — all findings addressed (Gemini + Codex + 2 independent adversarial passes)

This PR has been through **Gemini Code Assist**, **Codex**, and **2 independent adversarial review passes** (each finding adversarially verified). **Every finding at every severity is fixed.**

### Automated bot findings

| Finding | Severity | Fix | Commit |
|---|---|---|---|
| Interceptor ran PRAGMAs unconditionally on any connection → would break on a non-SQLite provider or a read-only DB (`SQLITE_READONLY` prevents connection open) | Gemini **HIGH** | Guard to `SqliteConnection` only; treat PRAGMAs as best-effort — a `SqliteException` is swallowed so it can't block connection open | `9fc6b57b` |
| `journal_mode=WAL` ran *before* `busy_timeout`, so a WAL switch contended by another process's lock could fail immediately with `SQLITE_BUSY` before the wait applied | Codex **P2** | Set `busy_timeout` **first**, then `journal_mode=WAL` (busy_timeout is applied even if the WAL switch is what fails) | `9fc6b57b` |
| Temp-file cleanup caught only `IOException`; Windows can throw `UnauthorizedAccessException` on a locked file, failing the test run | Gemini **MEDIUM** | Broaden cleanup catches to `Exception` (best-effort) in both test files | `9fc6b57b` |

### Independent adversarial review (2 lenses → verified)

| Finding | Severity | Fix | Commit |
|---|---|---|---|
| Enabling WAL silently regressed the dev-sandbox file-copy DB export/import: export could miss uncheckpointed `-wal` data; import left the previous DB's `-wal`/`-shm`, which SQLite replays → corruption | **MEDIUM** | `IUnitOfWork.CheckpointWalAsync` (`PRAGMA wal_checkpoint(TRUNCATE)`) called before export; import deletes pre-existing `-wal`/`-shm`; tests for both | `3cf977b1` |
| The concurrent-write test passed even without the interceptor (a fast single insert never contends) — it guarded nothing | **LOW** | Replaced with an honest coexistence smoke test + a deterministic guard (held `BEGIN IMMEDIATE` lock + `busy_timeout=0` → fails fast with `SQLITE_BUSY`, proving busy_timeout is load-bearing) | `4bb05902` |
| Integration tests bypass the interceptor (factory overrides the DbContext), so the `AddInfrastructure` DI wiring was unasserted | **LOW** | Added a DI-wiring test: `AddInfrastructure` + file Data Source → resolve `DbContext` → assert `journal_mode=wal` + `busy_timeout` | `4bb05902` |

### Verification
- Full backend solution builds clean.
- `SqlitePragmaInterceptorTests` (PRAGMA + coexistence + busy_timeout contention) + `ConnectorEncryptionKeyFailFastTests` (incl. new DI-wiring test) + a fake-using worker test: pass.
- `DatabaseExportImportRoundTripTests` (incl. checkpoint-on-export + stale-WAL-deletion-on-import): 23 pass.
- Added the no-op `CheckpointWalAsync` to the 8 hand-written `IUnitOfWork` test fakes.

### Scope
Relates to #1130 (WAL + busy_timeout, ACs 1 + 3). AC #2 (cross-process `Database.Migrate()` serialization) is tracked in **#1164**.

Chris0Jeky:
**Addendum — Codex re-review (P2) addressed in `b485766d`:** `wal_checkpoint(TRUNCATE)` can report `busy` (frames not fully folded) under concurrency, and the prior `ExecuteSqlRawAsync` discarded that result. `CheckpointWalAsync` now inspects the result row, retries up to 3×, and throws if it still can't fully checkpoint — so the dev-sandbox export fails with a clear 'stop other processes and retry' message instead of silently returning an incomplete snapshot. Added a DI-based test exercising the real checkpoint on a WAL database (Api.Tests: 7 pass).

Chris0Jeky:
**Addendum — Codex re-review rounds addressed:**
- `b485766d`: `CheckpointWalAsync` inspects the `wal_checkpoint` result, retries, and fails loudly if it can't fully fold the WAL.
- `37ef42b0`: import deletes the previous DB's `-wal`/`-shm` **strictly** (a failure fails the import + restores backup, not a silent success); checkpoint opens **through EF** so the interceptor installs `busy_timeout` first.
- `b0373e64`: import **checkpoints the existing DB before its rollback backup**, so a failed import can't restore a main file missing committed WAL data.

The dev-sandbox file-copy export/import is now WAL-safe for realistic cases. It remains inherently fragile under WAL/concurrency (pooled-connection handles vs `File.Replace`, atomic side-file handling); a proper redesign using SQLite's online backup API / `VACUUM INTO` is tracked in **#1166** (dev-sandbox-only, not a live-data risk). The core of this PR — WAL + `busy_timeout` for the live UI+MCP+CLI concurrency — is unaffected by that follow-up.

PR #1171, opened 2026-06-05
perf: bound notification paging + AsSplitQuery card reads (Part of #1133)

Chris0Jeky:
Part of #1133 (this PR delivers the two query-perf fixes from the issue; the SignalR incremental-refetch and FTS card-search ACs are explicitly out of scope here).

## Fix 1 — NotificationRepository unbounded fetch (HIGH)
`GetByUserIdAsync` did `ToListAsync` then `OrderByDescending(CreatedAt).Skip(offset).Take(limit)` **in memory** — it materialized every matching row regardless of `limit`/`offset`, so paging never reduced DB load.

- Pushes `ORDER BY CreatedAt DESC` + `LIMIT`/`OFFSET` into SQL.
- SQLite path uses raw `FromSqlRaw` (fully parameterized) because SQLite cannot translate `DateTimeOffset` ORDER BY from LINQ (documented in ADR-0023 and `DatabaseProviderCompatibilityTests.DateTimeOffset_OrderBy_RequiresClientSideForSqlite`). This mirrors the existing `AgentRunRepository` SQLite pattern.
- Non-SQLite providers keep strongly-typed LINQ (`OrderByDescending(...).Skip(...).Take(...)`).
- Negative `limit`/`offset` are clamped to 0 so a negative SQLite `LIMIT` (which means "unbounded") cannot silently re-introduce the very bug being fixed.

**Audit of the rest of the file:** `GetUnreadByUserIdAsync` and `GetByUserAndDeduplicationKeyAsync` do **not** have the anti-pattern — they filter in SQL and materialize (no in-memory sort/page), so they were left unchanged.

## Fix 2 — CardRepository cartesian Include (MEDIUM)
Added `.AsSplitQuery()` to every read method that chains `.Include(CardLabels).ThenInclude(Label)` across multiple cards (cartesian fan-out): `GetByBoardIdAsync` (the AC's named target), `GetByBoardIdsAsync`, `GetByColumnIdAsync`, and `SearchAsync`. Placement matches the established `BoardRepository.GetByIdWithDetailsAsync` pattern (added in #1147).

Not changed (and why):
- `GetAgendaByBoardIdsAsync` — has **no** `Include`, so no collection fan-out; `AsSplitQuery()` would be a no-op.
- `SearchAcrossBoardsAsync` / `GetByDueDateRangeAsync` — only `Include` to-one reference navigations (`Board`, `Column`); split query does not apply.
- `GetByIdWithLabelsAsync` — single root (`FirstOrDefaultAsync`), one collection; no cartesian multiplication, so split provides no benefit.

## Tests
- **NotificationRepository:** new `GetByUserIdAsync_WithPaging_ReturnsExactNewestFirstSliceBoundedToLimit` — seeds 5 rows (> page size) with explicit timestamps, requests `limit:2, offset:1`, and asserts the result is exactly the expected newest-first slice `[idx3, idx2]` and bounded to `limit` rows. (No SQL-row-count harness exists in this repo; per the task this is a correctness+bound assertion.)
- **CardRepository:** new `GetByBoardIdAsync_WithMultipleLabels_LoadsAllLabelsWithoutDuplicatingCards` — seeds 3 cards (2 labels / 1 label / none), asserts the split query returns each card exactly once with all labels loaded, ordered by position.

These run against real file-based SQLite via `TestWebApplicationFactory`, so the SQLite raw-SQL paging path and the split-query path are the ones actually exercised.

## Verification
- `dotnet build backend/Taskdeck.sln -c Release` → **0 errors** (4 pre-existing CS1998 warnings in unrelated `*.Tests` files).
- `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release --filter "FullyQualifiedName~NotificationRepository|FullyQualifiedName~CardRepository"` → **Passed: 24, Failed: 0, Skipped: 0**.

### DIFF
--- a/backend/src/Taskdeck.Infrastructure/Repositories/CardRepository.cs
+++ b/backend/src/Taskdeck.Infrastructure/Repositories/CardRepository.cs
@@ -17,6 +17,9 @@ public class CardRepository : Repository<Card>, ICardRepository
             .Where(c => c.BoardId == boardId)
             .Include(c => c.CardLabels)
                 .ThenInclude(cl => cl.Label)
+            // Split the Cards->CardLabels->Label collection fan-out into separate queries
+            // to avoid a cartesian row explosion on the hottest board read path.
+            .AsSplitQuery()
             .OrderBy(c => c.ColumnId)
                 .ThenBy(c => c.Position)
             .ToListAsync(cancellationToken);
@@ -36,6 +39,8 @@ public class CardRepository : Repository<Card>, ICardRepository
             .Where(card => materializedBoardIds.Contains(card.BoardId))
             .Include(card => card.CardLabels)
                 .ThenInclude(cardLabel => cardLabel.Label)
+            // Split the collection fan-out to avoid a cartesian product across boards.
+            .AsSplitQuery()
             .OrderBy(card => card.BoardId)
                 .ThenBy(card => card.ColumnId)
             .ThenBy(card => card.Position)
@@ -69,6 +74,8 @@ public class CardRepository : Repository<Card>, ICardRepository
             .Where(c => c.ColumnId == columnId)
             .Include(c => c.CardLabels)
                 .ThenInclude(cl => cl.Label)
+            // Split the collection fan-out to avoid a cartesian product over the column's cards.
+            .AsSplitQuery()
             .OrderBy(c => c.Position)
             .ToListAsync(cancellationToken);
     }
@@ -102,6 +109,8 @@ public class CardRepository : Repository<Card>, ICardRepository
         }
 
         return await query
+            // Split the Cards->CardLabels->Label collection fan-out to avoid a cartesian product.
+            .AsSplitQuery()
             .OrderBy(c => c.ColumnId)
                 .ThenBy(c => c.Position)
             .ToListAsync(cancellationToken);
--- a/backend/src/Taskdeck.Infrastructure/Repositories/NotificationRepository.cs
+++ b/backend/src/Taskdeck.Infrastructure/Repositories/NotificationRepository.cs
@@ -45,7 +45,9 @@ public class NotificationRepository : Repository<Notification>, INotificationRep
                 parameters.Add(boardId.Value);
             }
 
-            sql.Append(" ORDER BY CreatedAt DESC");
+            // Id is a deterministic tiebreaker so offset paging stays consistent when
+            // two notifications share the same CreatedAt (mirrors BoardRepository.SearchIdsAsync).
+            sql.Append(" ORDER BY CreatedAt DESC, Id");
             sql.Append(" LIMIT {").Append(parameters.Count).Append('}');
             parameters.Add(boundedLimit);
             sql.Append(" OFFSET {").Append(parameters.Count).Append('}');
@@ -73,6 +75,7 @@ public class NotificationRepository : Repository<Notification>, INotificationRep
 
         return await query
             .OrderByDescending(n => n.CreatedAt)
+            .ThenBy(n => n.Id)
             .Skip(boundedOffset)
             .Take(boundedLimit)
             .ToListAsync(cancellationToken);
--- a/backend/src/Taskdeck.Infrastructure/Repositories/NotificationRepository.cs
+++ b/backend/src/Taskdeck.Infrastructure/Repositories/NotificationRepository.cs
@@ -1,3 +1,4 @@
+using System.Text;
 using Microsoft.EntityFrameworkCore;
 using Taskdeck.Application.Interfaces;
 using Taskdeck.Domain.Entities;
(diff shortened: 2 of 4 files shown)


Chris0Jeky:
## Adversarial Code Review (self-review)

### CRITICAL
- None.

### HIGH
- None. Verified: all dynamic values in the raw SQL path (`userId`, `boardId`, `limit`, `offset`) are parameterized via `FromSqlRaw` positional placeholders — no SQL injection. The only inlined fragment is the constant `IsRead = 0`. Guid/bool parameter binding is confirmed by the pre-existing `WithBoardFilter` and `WithUnreadOnly` tests, which now execute through the new raw-SQL path and pass.

### MEDIUM
- **Pagination is not deterministic when `CreatedAt` ties.** Both the SQLite raw path (`ORDER BY CreatedAt DESC`) and the LINQ path (`OrderByDescending(CreatedAt)`) order only by `CreatedAt`. `CreatedAt` is `DateTimeOffset.UtcNow`, so two notifications created in the same tick (plausible for batch-emitted notifications) can be skipped or duplicated across offset pages — which undermines the "paging is correct" goal of this PR. The codebase already solves this in `BoardRepository.SearchIdsAsync` with `.ThenBy(b => b.Id)` "so offset pagination produces consistent pages." Fix: add `Id` as a deterministic tiebreaker on both paths. (Pre-existing flaw, but in-scope since this PR owns the paging correctness.)

### LOW
- `StringBuilder.Append(int)` for placeholder indices is culture-sensitive in principle, but the values are small non-negative integers that format identically across cultures and there is no warnings-as-errors policy — no change needed.
- `SELECT *` in the raw query is safe: EF maps FromSql results by column name, and this matches the established `AgentRunRepository` pattern.

### Bot Comments Addressed
- None present (no human, bot, or review-thread comments on the PR).

### Audit notes (no change required)
- `GetUnreadByUserIdAsync` / `GetByUserAndDeduplicationKeyAsync` do not have the materialize-then-sort/page anti-pattern (filter in SQL, no in-memory sort/page) — intentionally left unchanged.
- `GetAgendaByBoardIdsAsync` has no `Include`, and `SearchAcrossBoardsAsync`/`GetByDueDateRangeAsync` only include to-one reference navigations, so `AsSplitQuery()` does not apply there.

### Summary
1 MEDIUM (pagination tiebreaker), 0 CRITICAL/HIGH. Not merge-blocking in the strict sense, but the MEDIUM will be fixed now per the fix-everything policy.

Chris0Jeky:

| Pagination not deterministic when `CreatedAt` ties | MEDIUM | `96df48b9` | New tie-paging regression test + 25/25 affected tests pass |
| SQL injection / parameter binding in raw path | HIGH (cleared) | n/a | Confirmed parameterized; existing board/unread filter tests pass via the raw path |
| `Append(int)` culture, `SELECT *` mapping | LOW (cleared) | n/a | No-op values; matches `AgentRunRepository` precedent — no change needed |

### Fix detail
Added `Id` as a deterministic secondary sort to both the SQLite raw-SQL path (`ORDER BY CreatedAt DESC, Id`) and the LINQ path (`.ThenBy(n => n.Id)`), matching `BoardRepository.SearchIdsAsync`. Added `GetByUserIdAsync_WithTiedCreatedAt_PagesDeterministicallyWithoutGapsOrDuplicates`, which seeds tied timestamps and asserts the pages cover every row exactly once (no gaps/duplicates) and that repeated queries are stable.

### Verification
- `dotnet test ... --filter "FullyQualifiedName~NotificationRepository|FullyQualifiedName~CardRepository"` -> **Passed: 25, Failed: 0, Skipped: 0**.
- `dotnet build backend/Taskdeck.sln -c Release` -> 0 errors.

All findings addressed (no remaining MEDIUM/LOW). CI status: PENDING (re-running after the fix push; Gitleaks + Dependency Review already green, no failures). Not merging — leaving for coordinator review.

Chris0Jeky:
## Adversarial-review fix evidence (test-quality gaps)

Commit `2d5d46f` (`backend/tests/Taskdeck.Api.Tests/NotificationRepositoryIntegrationTests.cs`, +193). Production code unchanged — both prior reviews confirmed it correct (parameterized FromSqlRaw, no injection).

### MEDIUM — paging regression test now actually detects a revert to in-memory paging
The existing `GetByUserIdAsync_WithPaging_ReturnsExactNewestFirstSliceBoundedToLimit` could not distinguish the SQL-paged impl from the old materialize-then-`Skip/Take` anti-pattern — both return the same rows.

Fix: added `GetByUserIdAsync_WithPaging_PushesLimitAndOffsetIntoSql`. It builds a `TaskdeckDbContext` over a temp SQLite file with a test-only `CapturingCommandInterceptor : DbCommandInterceptor` (overrides `ReaderExecuting`/`ReaderExecutingAsync` to record `command.CommandText`), runs `GetByUserIdAsync(limit: 2, offset: 1)`, then asserts the captured SELECT against `Notifications` contains **LIMIT** AND **OFFSET** (case-insensitive). A revert to materialize-then-page emits no LIMIT/OFFSET on the SQLite path, so this test fails.

**Approach chosen: interceptor (not the ToQueryString fallback).** The interceptor captures the SQL the *production repository method* actually executes, so it proves the real code path. The `ToQueryString()` fallback would require reconstructing the FromSqlRaw query in the test, which only proves "a query I wrote in the test has LIMIT/OFFSET" — it would not exercise the repository. I mirrored the existing self-contained `SqlitePragmaInterceptorTests` harness (temp `Data Source` file + `DbContextOptionsBuilder.AddInterceptors(...)` + `Migrate` + best-effort temp-file cleanup with `SqliteConnection.ClearAllPools()`).

### LOW — added missing raw-SQL edge-case tests (factory fixture, mirrors existing style)
- `GetByUserIdAsync_WithNegativeLimit_ReturnsEmpty` — `limit: -1` returns empty (proves the clamp; a negative LIMIT means "unbounded" in SQLite).
- `GetByUserIdAsync_WithZeroLimit_ReturnsEmpty` — `limit: 0` returns empty.
- `GetByUserIdAsync_WithUnreadOnlyAndBoardFilter_ReturnsOnlyMatchingUnreadOnBoard` — seeds read/unread × on/off the board and asserts only the single unread notification on the target board is returned.

(Did not add an offset-beyond-count test — `GetByUserIdAsync_WithOffsetBeyondTotal_ShouldReturnEmpty` already covers it.)

### Verification
- `dotnet build backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release` → **0 errors, 0 warnings**.
- `dotnet test ... --no-build --filter "FullyQualifiedName~NotificationRepository|FullyQualifiedName~CardRepository"` → **Passed: 29, Failed: 0, Skipped: 0** (includes the 4 new tests + unchanged CardRepository `AsSplitQuery` coverage).

The AccountDeletionService LOW from review is out of scope here and is tracked separately in #1172.

PR #1170, opened 2026-06-05
Required CI security gate: gitleaks + dependency + SAST + bundle-size (#1132 AC1/AC5)

Chris0Jeky:
## Summary
Closes the CI-gate acceptance criteria of #1132 (AC1 + AC5). The required PR merge gate
(`ci-required.yml`) ran **no** secret/SAST/dependency scan and the bundle-size budget ran only
behind the non-blocking `performance` label — so a PR with a hardcoded secret, a high-severity
vulnerable dependency, or a bundle blow-up could merge unblocked.

### AC1 — secret / dependency / SAST scans now block merge
Promoted three reusable workflows from the label-gated extended/nightly lanes into the required lane:

| Job | Reusable workflow | Enforcement |
|-----|-------------------|-------------|
| `secret-scan` | `reusable-gitleaks` | `scan-mode: pr`, `fail-on-findings: true` |
| `dependency-security` | `reusable-dependency-security-signals` | `frontend-audit-level: high`, `enforce-findings: true` |
| `sast-scan` | `reusable-sast-scanning` (Semgrep) | `enforce-findings: true` |

gitleaks runs in **PR mode** (scans only this PR's commits), so it does not trip on the pre-existing
test fixtures already in history — it catches *newly introduced* secrets.

### AC5 — bundle-size budget in the required lane
Added a `Check frontend bundle size` step to the required `frontend-unit` job. It **reuses the
existing `vite build`** (no extra install/build → the "cheap" check the issue asks for) and is
gated to Linux to avoid a redundant run on the Windows matrix leg. Error thresholds match
`scripts/ci/check-bundle-size.mjs` defaults (entry 150 / single 250 / total-js 1200 KB).

## Validation
- Both workflow files parse as YAML.
- Bundle gate run locally against the current `dist`: **994.24 KB total < 1200 KB error threshold → "All bundle size checks passed."**
- The security scans run for the first time as **blocking** on this PR's own CI (pull_request uses
  the PR's workflow definitions). **If they surface any pre-existing finding I will address it in
  this PR** (fix the real issue, or add a documented suppression + tracking issue) per the AC's
  "validate they don't false-positive-block; tune as needed."

## Note for the maintainer
`ci-required.yml` has no aggregation gate job — each job is an independent check. For these three to
*enforce* merge-blocking, the new check names (**Secret Scan / Dependency Security / SAST Scan**, and
the bundle step inside **Frontend Unit**) may need adding to the branch-protection required-checks
[...]

PR #1169, opened 2026-06-05
Security config fail-fast: single ≥32 JWT floor + CORS fail-closed in prod (#1132)

Chris0Jeky:
## Summary
Part of #1132 — closes two of its acceptance criteria (the surgical config-hardening ones).
The remaining ACs ship as separate, focused PRs (global `FallbackPolicy`; required CI
security/SAST/dependency scans + bundle gate).

### AC3 — single ≥32 JWT secret floor, fail fast
`JwtSettingsValidator` floored at **16** while `AuthenticationRegistration` **silently no-op'd**
JwtBearer registration below **32**. The gap: a 16–31 char secret booted the app with
authentication effectively *off* (no scheme registered → `[Authorize]` faults at request time).

- New single source of truth: `JwtSettings.MinSecretKeyLength = 32`.
- `JwtSettingsValidator` consumes it (16 → 32; `ValidateOnStart()` already wired → fail-fast at boot).
- `AuthenticationRegistration` now **throws** `InvalidOperationException` (clear message) instead of
  returning a no-op `ServiceCollection` on a missing / short / incomplete JWT config.

### AC2 — CORS fails closed in Production
In non-dev, empty `Cors:AllowedOrigins` fell back to localhost while the `AllowFrontend` policy
calls `.AllowCredentials()` → a deploy that forgets origins authorized **credentialed cross-origin
requests from localhost**. The prod branch now returns an **empty origin set** (deny-all
cross-origin) and emits a loud bootstrap warning. Same-origin (the desktop single-exe serving the
SPA from `wwwroot`) never hits CORS, so the local-first posture is unaffected.

## Why no-op→throw / fallback→deny is safe for existing boots
`AuthenticationRegistration` already required ≥32 to register a scheme (it just no-op'd below that),
so every booting harness already uses a ≥32 key — only the *silent-failure* path changes. The
sub-32 strings in the suite are unit-only (`IsPlaceholder`, DataAnnotations `ExpirationMinutes`),
never booting a web host. CORS fallback only changes the prod-with-no-origins case.

## Tests
- `OptionsValidationTests`: 31-char fails / 32-char passes boundary (tied to the constant).
- `AuthenticationRegistrationTests` (new): throws on empty/short/missing-issuer; registers auth when valid.
- `CorsApiTests`: updated the test that locked in the *insecure* fallback; added fail-closed
  (default + arbitrary origin denied) and configured-prod-origin-allowed cases.

```
[...]

PR #1176, opened 2026-06-05
Global authorization FallbackPolicy (require auth by default) + AllowAnonymous audit (#1132 AC4)

### LINKED ISSUE #1132: Security gate + config hardening: required secret/SAST/dependency scans, CORS fail-closed, JWT floor, FallbackPolicy, bundle gate
## Context
Crypto/auth primitives are mature, but the required PR gate and a few config edges are weak.

## Evidence
- `ci-required.yml` jobs:block (85-143) runs **no** secret/SAST/dependency scan — `reusable-gitleaks.yml`, `reusable-dependency-security-signals.yml`, `reusable-sast-scanning.yml` are wired only into label-gated `ci-extended`/nightly/release (named only in comment lines 21/30/43/48). A PR with a hardcoded secret or vulnerable package is not blocked.
- `CorsRegistration.cs:40-49`: in non-dev, empty `Cors:AllowedOrigins` falls back to localhost origins while `AllowFrontend` calls `.AllowCredentials()` (line 19) → a prod deploy that forgets origins authorizes credentialed cross-origin from localhost.
- JWT floor inconsistency: `JwtSettingsValidator.cs:18`=16, `AuthenticationRegistration.cs:38-44` **silently no-ops** JwtBearer registration at <32, `AuthenticationService.cs:441-445` throws at <32 → a 16-31 char secret can boot the app with auth effectively off.
- No runtime `FallbackPolicy` (`Program.cs:325-326` registers only AdminOnly); protection relies on the `ApiControllerBoundaryTests` regex scan (TopDirectoryOnly, misses subfolders/minimal-API/SignalR).
- Bundle-size gate (994/1000 KB, 5.76 KB headroom) runs only behind the `performance` label in non-blocking `ci-extended`.

## Acceptance criteria
- [ ] gitleaks (PR mode, fail-on-detect) + dependency-security + SAST added as **required** jobs in `ci-required.yml`. (Validate they don't false-positive-block; tune as needed.)
- [ ] Prod CORS fails closed (empty origins + startup warning, or throw) when no origins configured.
- [ ] Single ≥32 JWT floor: raise `JwtSettingsValidator.MinSecretKeyLength` to 32 (fail fast) and make `AuthenticationRegistration` throw, not no-op.
- [ ] Global `AuthorizationOptions.FallbackPolicy = RequireAuthenticatedUser`; explicit `[AllowAnonymous]` on login/register/health/OAuth.
- [ ] Move the cheap frontend bundle-size check into the required lane (or auto-label `performance` on `frontend/**`).

_Source: deep audit ci-security-gate-hardening (HIGH)._

Chris0Jeky:
## Summary
Closes #1132 AC4 — adds a global `AuthorizationOptions.FallbackPolicy = RequireAuthenticatedUser`
so any endpoint that lacks explicit authorization metadata requires an authenticated user
(defense-in-depth beyond the per-controller `[Authorize]` the architecture test already enforces).

### Endpoint audit (the AC's "explicit `[AllowAnonymous]`")
| Endpoint | Handling |
|----------|----------|
| login / register / health / OAuth controllers | already `[AllowAnonymous]` (unchanged) |
| `BoardsHub` (`MapHub`) | already `[Authorize]` → fallback doesn't apply |
| SPA fallback (`MapFallbackToFile`) | **`.AllowAnonymous()`** so the app shell loads to reach the login page |
| `/mcp` (`MapMcp`) | **`ApiKeyMiddleware` now sets an authenticated principal** for valid keys |
| Swagger | dev-only middleware, runs before authorization → unaffected |
| standalone `--mcp` HTTP host | has no `UseAuthorization` → unaffected |

### Why a principal for MCP (not `AllowAnonymous`)
`ApiKeyMiddleware` already 401s missing/invalid/revoked keys before routing but only set
`HttpContext.Items[UserId]`, not an authenticated `HttpContext.User`. The MCP SDK endpoint does
**not** honor an `.AllowAnonymous()` endpoint convention (verified: the SPA `MapFallbackToFile`
`.AllowAnonymous()` works, the MCP one didn't), so a valid-key request hit the fallback → 401.
A valid API key *is* authentication, so the middleware now sets a minimal `ClaimsPrincipal`
(NameIdentifier = userId, authType `ApiKey`). `TokenValidationMiddleware` is a no-op for it (its
JWT-invalidation check only fires when an `iat` claim is present), and it is set before
`UseAuthentication`, which finds no JWT on `/mcp` and does not overwrite it.

## Tests
New `FallbackPolicyTests`: anonymous allowlisted endpoint reachable; protected endpoint 401 when
anonymous; SPA fallback not blocked when anonymous; `/mcp` without a key 401 from ApiKeyMiddleware.

```
Passed!  Failed: 0, Passed: 276  (FallbackPolicy + Mcp + ApiKey + Auth)
```
Earlier broad run (Auth + Cors + Mcp + Health + Boards + Authorization + CrossUser) also green after
the MCP-principal fix. Full Api.Tests suite runs in CI.

## Review
Two independent adversarial reviews will run; all findings (every severity) addressed before merge.

### DIFF
--- a/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
@@ -124,6 +124,10 @@ public static class PipelineConfiguration
         // Authenticated via ApiKeyMiddleware (Bearer tdsk_... tokens).
         // Rate limiting applied per API key user identity.
         var mcpEndpoint = app.MapMcp();
+        // MCP is authenticated by ApiKeyMiddleware (above): it 401s missing/invalid/revoked keys
+        // before routing and sets an authenticated principal for valid keys, which satisfies the
+        // global FallbackPolicy (#1132 AC4). The MCP SDK endpoint does not honor an
+        // .AllowAnonymous() convention, so the principal — not endpoint metadata — is the opt-in.
         if (rateLimitingSettings.Enabled)
         {
             mcpEndpoint.RequireRateLimiting(RateLimiting.RateLimitingPolicyNames.McpPerApiKey);
@@ -131,8 +135,9 @@ public static class PipelineConfiguration
 
         // SPA fallback: any route not matched by a controller or hub endpoint returns index.html,
         // enabling Vue Router's client-side navigation. API (/api/*) and hub (/hubs/*) routes
-        // are matched above and never reach this fallback.
-        app.MapFallbackToFile("index.html");
+        // are matched above and never reach this fallback. AllowAnonymous so the app shell loads for
+        // unauthenticated users (e.g. to reach the login page) under the global FallbackPolicy (#1132 AC4).
+        app.MapFallbackToFile("index.html").AllowAnonymous();
 
         return app;
     }
--- a/backend/src/Taskdeck.Api/Middleware/ApiKeyMiddleware.cs
+++ b/backend/src/Taskdeck.Api/Middleware/ApiKeyMiddleware.cs
@@ -1,3 +1,4 @@
+using System.Security.Claims;
 using System.Security.Cryptography;
 using System.Text;
 using Microsoft.EntityFrameworkCore;
@@ -107,6 +108,15 @@ public sealed class ApiKeyMiddleware
 
         // Set the authenticated user ID for HttpUserContextProvider
         context.Items[HttpUserContextProvider.UserIdItemKey] = apiKey.UserId;
+
+        // Establish an authenticated principal so the global authorization FallbackPolicy
+        // (RequireAuthenticatedUser, #1132 AC4) is satisfied for valid-key MCP requests — a valid
+        // API key IS authentication. It carries only the user id (no JWT 'iat', so
+        // TokenValidationMiddleware's JWT-invalidation check is a no-op for it). Set before
+        // UseAuthentication runs; JwtBearer finds no token on /mcp and does not overwrite it.
+        context.User = new ClaimsPrincipal(new ClaimsIdentity(
+            new[] { new Claim(ClaimTypes.NameIdentifier, apiKey.UserId.ToString()) },
+            authenticationType: "ApiKey"));
 
         // Update last-used timestamp before continuing the pipeline.
         // This is non-critical so failures are swallowed.
--- a/backend/src/Taskdeck.Api/Middleware/ApiKeyMiddleware.cs
+++ b/backend/src/Taskdeck.Api/Middleware/ApiKeyMiddleware.cs
@@ -27,6 +27,12 @@ public sealed class ApiKeyMiddleware
     /// <summary>The path prefix that triggers API key authentication.</summary>
     private const string McpPathPrefix = "/mcp";
 
+    /// <summary>
+    /// Authentication type stamped on the ClaimsIdentity for a valid API key so downstream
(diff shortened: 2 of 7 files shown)


Chris0Jeky:
## Adversarial review (2 independent reviewers) + fix evidence

Two independent adversarial reviews (`auth-bypass-regression` + `correctness-completeness` lenses),
each finding verified by an independent skeptic. **6 findings confirmed, 0 dismissed.** Both reviewers
confirmed the core change is sound (no auth bypass; the fallback opt-outs and the ApiKey principal are
correct). All findings addressed:

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| 1 | MEDIUM | Tests were tautological — all 4 passed even if `SetFallbackPolicy` were removed (`/api/boards` 401s from `[Authorize]`, `/mcp` from ApiKeyMiddleware) | Added `FallbackPolicy_RequiringAuthenticatedUser_IsRegistered`: asserts `IAuthorizationPolicyProvider.GetFallbackPolicyAsync()` returns a policy with `DenyAnonymousAuthorizationRequirement` — **fails if `SetFallbackPolicy` is removed** |
| 2 | MEDIUM | No ADR/STATUS for the security-posture change | Added **ADR-0036** (default-deny FallbackPolicy) + INDEX entry. STATUS reconciliation lands in the post-#1132 docs-sweep (all 3 #1132 PRs together) per the normal cadence |
| 3 | LOW | Redundant per-`/mcp` DB load in TokenValidationMiddleware (the new principal made it stop short-circuiting) | TokenValidationMiddleware now short-circuits on `AuthenticationType == ApiKeyMiddleware.AuthenticationType` (new shared const) — API-key principals carry no JWT `iat` and are already validated by ApiKeyMiddleware |
| 4 | LOW | Misleading "JwtBearer finds no token on /mcp" comment | Reworded: JwtBearer **reads** the `Bearer tdsk_...` header and **fails** JWT validation; the ApiKey principal survives because a failed authenticate returns a null Principal that AuthenticationMiddleware does not assign |
| 5 | LOW | (duplicate of #3) | — |
| 6 | LOW | Dead `using Taskdeck.Api.Tests.Support` | Removed |

## Verification
```
Build: 0 errors.
Passed!  Failed: 0, Passed: 114  (FallbackPolicy + Mcp + ApiKey + TokenValidation)
```
The new policy-registration test and the TokenValidation short-circuit are included and green.

This PR completes **#1132 AC4**, the last of #1132's acceptance criteria (AC1/AC5 → #1170; AC2/AC3 → #1169).

Chris0Jeky:
@gemini-code-assist Thanks — this redundant query was independently caught by the adversarial review and is **already eliminated**, but via a different mechanism than omitting the claim: `TokenValidationMiddleware` now short-circuits when the principal's `AuthenticationType` is `ApiKeyMiddleware.AuthenticationType` (`ApiKey`), so it never runs the JWT-invalidation DB load for API-key principals (commit 900af6be). The `ClaimTypes.NameIdentifier` claim is **intentionally retained** so the principal is a complete authenticated identity carrying the user id (the alternative — an authenticated identity with no subject — is semantically odd and would break any future code reading `context.User` for the MCP user id). Net: no redundant query, and the principal stays well-formed. A regression test (`FallbackPolicy_RequiringAuthenticatedUser_IsRegistered`) also locks in the policy.

Issue #1131, opened 2026-05-30
CLI hardening: unusable on a fresh machine + bypasses claims-first authorization


Chris0Jeky:
## Context
The CLI fresh-machine bootstrap is delivered. The remaining acceptance is the explicit CLI authorization/trust model; the maintainer decision described in the 2026-08-23 realignment remains open.

## Evidence
- **Historical, fixed by PR #1177:** `Infrastructure/DependencyInjection.cs:112-120` throws `InvalidOperationException` when `Connectors:EncryptionKey` is empty; the CLI calls `AddInfrastructure` but never runs `FirstRunBootstrapper` (grep `FirstRun|EncryptionKey|EnsureConnector` in `Taskdeck.Cli` = nothing) and ships no `appsettings.json`. So `taskdeck boards list` throws at startup unless `TASKDECK_CONNECTORS__ENCRYPTIONKEY` is exported first.
- **Authorization bypass:** `CardsCommandHandler.cs:59-67` calls `_cardService.CreateCardAsync(CreateCardDto)` which takes no userId/actor and performs no `BoardAccess` check (`CardService.cs:35-44`). Board/card/column CLI commands operate on any board id regardless of ownership; only `ApiKeysCommandHandler` resolves an actor.
- **Historical migration evidence superseded:** current `Program.cs` calls `SerializedMigrator.Migrate`, with cross-process serialization and pre-migration snapshot safeguards; it does not call the original uncoordinated migration path.

## Acceptance criteria
- [x] CLI bootstraps the connector key (or registers `ICredentialEncryptionService` lazily so non-connector commands don't require it); add a CLI smoke test running `boards list` against a clean env.
- [ ] Route CLI board/card/column mutations through the same authorization path as controllers (resolve CLI actor, check board access), OR explicitly document + gate the CLI as an unrestricted local-admin tool.

_Source: deep-audit completeness critic (HIGH + MEDIUM)._

2026-09-07 reconciliation: current `CliFirstRunBootstrapper` runs before infrastructure registration. The existing `BoardsList_FreshMachine_BootstrapsKeyAndReturnsEmptyArray` and persisted-key reuse regressions are part of the CLI suite, which passed 243/243 during the full solution run at head `8e7fd25d6`. PR #1177 records original delivery as `54fb4770b`. This confirms AC1 without choosing AC2's local-admin-versus-board-authorization decision.

Chris0Jeky:
[...]

PR #1177, opened 2026-06-05
CLI hardening (#1131 AC1): provision connector encryption key on fresh machine

Chris0Jeky:
## Summary
Part of #1131 (AC1 only). On a clean machine, `taskdeck boards list` crashed at startup: `Cli/Program.cs` calls `AddInfrastructure`, which fail-fasts with `InvalidOperationException` when `Connectors:EncryptionKey` is empty. The CLI ships no `appsettings.json` and never provisioned the key, so it was unusable unless the operator exported `TASKDECK_CONNECTORS__ENCRYPTIONKEY` by hand.

This makes the CLI self-sufficient on first run by provisioning the connector encryption key **before** `AddInfrastructure`.

> AC2 (claims-first authorization routing for board/card/column mutations) is intentionally **deferred** and not in this PR.

## Approach: provision (not lazy)
New CLI-local helper `Taskdeck.Cli/CliFirstRunBootstrapper.cs`, invoked from `Program.cs` after the connection string is resolved and before `AddInfrastructure`:

1. If `Connectors:EncryptionKey` is already present (env var, appsettings, user secrets) it is a **no-op** — explicit/operator config always wins.
2. Otherwise it reads/generates a key and persists it to a CLI-local `appsettings.local.json`, then loads it into configuration so `AddInfrastructure` succeeds.

Chose **provision** over lazy `ICredentialEncryptionService` registration because the key must be **stable across runs** — a regenerated key each invocation would make previously-encrypted connector credentials undecryptable. Persisting mirrors the API's `FirstRunBootstrapper` write pattern. The shared `AddInfrastructure` is **not** modified (the API intentionally fail-fasts on a missing key in Production).

### Key persistence location
Next to the **resolved SQLite data directory** (parsed from the connection string), not next to the executable. This is user-writable (the install dir may be read-only), co-locates the key with the encrypted data it protects, and is the same dir as `taskdeck.db`. Falls back to the current working directory if the data source can't be resolved to a path.

### Safety
- 256-bit key via `RandomNumberGenerator.GetBytes(32)` -> base64.
- Only provisions when missing; never overwrites a configured key.
- Merge-preserves any other keys already in `appsettings.local.json` (JSON read-modify-write).
- Atomic write (temp file + `File.Move(overwrite)` with retry) so a partial/corrupt file is never observed.
- Cross-process named mutex with re-check inside the lock, so two CLI invocations racing on a fresh machine converge on **one** shared key instead of each generating a different one.
- Never writes to stdout (CLI keeps stdout clean JSON); diagnostics go to stderr only, and only on the persist-failure fallback.

## Tests
Both end-to-end and focused tests (the harness change made the end-to-end path practical):
- **End-to-end CLI smoke tests** (`CliTestHarness` extended with `provisionEncryptionKey: false` + per-instance isolated data dir): run `boards list --json` against a clean env (no key, no pre-existing `appsettings.local.json`) and assert exit 0, empty JSON array, and that `appsettings.local.json` now contains a valid base64 256-bit key. A second test asserts the key is **stable across two runs**.
- **Unit tests** of `CliFirstRunBootstrapper`: provisions when missing; no-op when configured; reuses the persisted key on a second (fresh-process) call; preserves unrelated keys in the file; `GenerateKey` yields a unique base64 256-bit value.
- **Integration test**: `AddInfrastructure` throws on a clean config (reproduces the bug), then succeeds after the bootstrap, and the resolved `ICredentialEncryptionService` does a real encrypt/decrypt round-trip.

## Verification
- `dotnet build backend/Taskdeck.sln -c Release` -> **0 errors** (11 warnings, all pre-existing in unrelated `Api.Tests`/`Application.Tests`).
- `dotnet test backend/tests/Taskdeck.Cli.Tests/...` -> **90 passed, 0 failed** (8 new). New class alone: **8 passed**.

## Files changed
- `backend/src/Taskdeck.Cli/CliFirstRunBootstrapper.cs` (new)
- `backend/src/Taskdeck.Cli/Program.cs` (wire-in before `AddInfrastructure`)
- `backend/tests/Taskdeck.Cli.Tests/CliTestHarness.cs` (fresh-machine mode + isolated data dir)
- `backend/tests/Taskdeck.Cli.Tests/CliFirstRunBootstrapTests.cs` (new)

### DIFF
--- a/backend/src/Taskdeck.Cli/CliFirstRunBootstrapper.cs
+++ b/backend/src/Taskdeck.Cli/CliFirstRunBootstrapper.cs
@@ -12,18 +12,30 @@ namespace Taskdeck.Cli;
 /// The CLI ships no <c>appsettings.json</c> and never runs the API's
 /// <c>FirstRunBootstrapper</c>, yet <c>AddInfrastructure</c> fail-fasts when
 /// <c>Connectors:EncryptionKey</c> is missing. On a clean machine that made
-/// <c>taskdeck boards list</c> crash at startup unless the operator exported
-/// <c>TASKDECK_CONNECTORS__ENCRYPTIONKEY</c> by hand.
+/// <c>taskdeck boards list</c> crash at startup unless the operator exported a
+/// key by hand (<c>Connectors__EncryptionKey</c>, or
+/// <c>TASKDECK_CONNECTORS__ENCRYPTIONKEY</c> now that the CLI host registers the
+/// <c>TASKDECK_</c> prefix).
 ///
 /// This helper provisions a cryptographically-random 256-bit connector key on
-/// first run, persists it to a CLI-local <c>appsettings.local.json</c> (next to
-/// the resolved SQLite data directory, so the key is co-located with the
-/// encrypted data it protects and lives in a user-writable location), and loads
-/// it into configuration so <c>AddInfrastructure</c> succeeds.
+/// first run, persists it to a CLI-local <c>appsettings.local.json</c> next to
+/// the resolved SQLite data directory (a user-writable location alongside the
+/// data it protects), and loads it into configuration so
+/// <c>AddInfrastructure</c> succeeds.
 ///
 /// It intentionally does NOT modify the shared <c>AddInfrastructure</c>: the API
-/// deliberately fail-fasts on a missing key in Production. This is a CLI-local
-/// convenience that mirrors the API's <c>FirstRunBootstrapper</c> write pattern.
+/// deliberately fail-fasts on a missing key in Production. NOTE the key location
+/// differs from the API's <c>FirstRunBootstrapper</c>, which persists ITS key
+/// next to the executable (<c>AppContext.BaseDirectory</c>) -- this CLI bootstrap
+/// writes next to the data directory. Because the two locations differ, a
+/// deployment that points BOTH the API and CLI (or several hosts) at one shared
+/// database would auto-generate a DIFFERENT key per host and be unable to decrypt
+/// the other's connector credentials. For any shared-database or multi-host
+/// deployment, set one explicit <c>Connectors__EncryptionKey</c> (env var or
+/// appsettings) on every host instead of relying on this per-host
+/// auto-generation; the bootstrap emits a stderr warning when it auto-generates a
+/// key against a non-default data directory (<c>TASKDECK_CONNECTION_STRING</c>
+/// set) to surface this.
 ///
 /// Must run BEFORE the DI container is built and must never write to stdout
 /// (the CLI keeps stdout clean JSON for callers); diagnostics go to stderr only.
@@ -38,8 +50,14 @@ internal static class CliFirstRunBootstrapper
     /// <summary>
     /// Ensures a connector encryption key is available in configuration,
     /// generating and persisting one on first run when none is configured.
-    /// No-op when a key is already supplied (environment variable, appsettings,
-    /// or a previously generated <c>appsettings.local.json</c>).
+    /// Returns early (no file I/O) when a key is already present in the live
+    /// <see cref="IConfiguration"/> -- i.e. an environment variable
+    /// (<c>Connectors__EncryptionKey</c>, or <c>TASKDECK_CONNECTORS__ENCRYPTIONKEY</c>
+    /// once the CLI host registers the <c>TASKDECK_</c> prefix) or an appsettings
+    /// entry. The previously generated <c>appsettings.local.json</c> is NOT
+    /// registered as a configuration source, so it does not trigger this early
+    /// return; instead <see cref="EnsureKeyOnDisk"/> re-reads that file on every
+    /// run and reuses the persisted key (idempotent).
     /// </summary>
     public static void EnsureConnectorEncryptionKey(IConfiguration configuration)
     {
@@ -88,9 +106,16 @@ internal static class CliFirstRunBootstrapper
             var fullPath = Path.GetFullPath(dataSource);
             return Path.GetDirectoryName(fullPath) ?? Directory.GetCurrentDirectory();
         }
-        catch (ArgumentException)
+        catch (Exception ex) when (
+            ex is ArgumentException
+                or PathTooLongException
+                or NotSupportedException
+                or IOException
+                or System.Security.SecurityException)
         {
-            // Non-file data sources (e.g. ":memory:") have no real directory.
+            // Non-file data sources (e.g. ":memory:") or otherwise unresolvable paths
+            // (too long, unsupported format, restricted) must not crash startup --
+            // fall back to the current working directory.
             return Directory.GetCurrentDirectory();
         }
     }
@@ -123,12 +148,17 @@ internal static class CliFirstRunBootstrapper
     internal static string EnsureKeyOnDisk(string localConfigPath)
     {
         var mutexName = BuildMutexName(localConfigPath);
-        using var mutex = new Mutex(initiallyOwned: false, name: mutexName);
+        Mutex? mutex = null;
         var acquired = false;
         try
         {
             try
             {
+                // The named (Global\ on Windows) mutex ctor AND WaitOne can throw on
+                // locked-down or multi-user hosts -- e.g. when another user already
+                // owns the name. That must NOT crash the CLI at startup (the exact
+                // failure this bootstrap exists to prevent).
+                mutex = new Mutex(initiallyOwned: false, name: mutexName);
                 acquired = mutex.WaitOne(TimeSpan.FromSeconds(10));
             }
             catch (AbandonedMutexException)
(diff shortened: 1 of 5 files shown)


Chris0Jeky:
## Adversarial Code Review

Self-review of the bootstrap fail-safe. Focus: the helper's entire purpose is to *prevent* a startup crash, so any unhandled exception inside it is a real defect.

### CRITICAL
- None.

### HIGH
- None. (Authorization routing / AC2 is intentionally out of scope; the encryption-key fail-fast is the only startup blocker addressed here.)

### MEDIUM
- **M1 — Wrong-type key in `appsettings.local.json` crashes the bootstrapper.** `EnsureKeyOnDisk` extracts the existing key via `JsonNode...GetValue<string>()`. If a hand-edited or tool-written file has `Connectors:EncryptionKey` as a non-string JSON value (e.g. `{"Connectors":{"EncryptionKey":123}}`), `GetValue<string>()` throws `InvalidOperationException` *outside* the parse try/catch — crashing the very startup path this fix exists to protect.
- **M2 — Unreadable-but-existing config file crashes the bootstrapper, and naive recovery risks clobbering a valid key.** `ReadConfigObject` does not guard `File.ReadAllText` against `IOException`/`UnauthorizedAccessException` (file transiently locked by an editor/AV/backup while we hold our own mutex — an external lock). That propagates and crashes the CLI. Simply "start fresh" would be worse: it could overwrite a still-valid on-disk key, making previously-encrypted connector credentials undecryptable. Fix: when an existing file can't be read, preserve it and use a transient in-memory key for this run (the next run picks up the real key).

### LOW
- **L1 — Misleading comment.** The class/method docs list "user secrets" as a source that wins over the generated file. The CLI csproj has no `UserSecretsId`, so user secrets are never loaded. Tidy the wording so future readers aren't misled.
- **L2 — Test gap.** No coverage for malformed / wrong-type `appsettings.local.json`. Add tests now that the read path is hardened (corrupt JSON and wrong-type value should both regenerate a valid key without throwing).

### Bot Comments Addressed
- None outstanding. CI is still pending; `Secret Scan / Gitleaks` already passed (confirms the only base64 literal — the all-zeros test placeholder, pre-existing in `CliTestHarness`/now in tests — is not flagged as a real secret).

### Summary
0 CRITICAL, 0 HIGH, 2 MEDIUM, 2 LOW. Not architecturally merge-blocking, but the two MEDIUMs are robustness holes in the fail-safe itself and will be fixed in this PR (plus the LOWs). Fixes to follow in the next commits with a mapping comment.

Chris0Jeky:

All findings from the review above are addressed in this PR.

| M1 — wrong-type key value crashes bootstrapper | MEDIUM | `7544849` | Type-safe extraction via `JsonValue.TryGetValue<string>`; new `[Theory]` case `{"EncryptionKey":12345}` passes |
| M2 — unreadable existing config crashes / risks clobbering valid key | MEDIUM | `7544849` | `ReadExisting` now catches `IOException`/`UnauthorizedAccessException` and sets `PreserveFile` so an existing-but-unreadable file is never overwritten (transient key used for that run) |
| L1 — misleading "user secrets" comment | LOW | `7544849` | Reworded class/method docs (CLI has no `UserSecretsId`) |
| L2 — no malformed-config test coverage | LOW | `2a986f1` | New `[Theory]` covers corrupt JSON, non-object root, wrong-type value, and empty-string value — all regenerate a valid 256-bit key without throwing |

### Verification
- `dotnet build backend/Taskdeck.sln -c Release` -> **0 errors**.
- `dotnet test backend/tests/Taskdeck.Cli.Tests/...` -> **94 passed, 0 failed** (was 90; +4 from the malformed-config theory).

CI status: PENDING (will re-run on the pushed fix commits). No bot or human comments were outstanding.

Chris0Jeky:
## Fix evidence — adversarial + bot review findings addressed

All findings from both adversarial reviewers and the Gemini/Codex bots are fixed and verified. Pushed commits:

| Commit | Scope |
| --- | --- |
| `7f5f0de4` | `CliFirstRunBootstrapper.cs` — mutex hardening, 0600 key file, path-catch, docstrings + shared-DB warning |
| `1948a520` | `Program.cs` — register `TASKDECK_` env prefix |
| `a550fb0b` | `DependencyInjection.cs` — correct fail-fast message |
| `78b7a7af` | `CliTestHarness.cs` — allow extra env vars in subprocess runs |
| `34cb38dc` | `CliFirstRunBootstrapTests.cs` — new end-to-end env-override no-op test |

### MEDIUM 1 — Unguarded `Mutex` ctor / silent persist on lock timeout (`7f5f0de4`)
- The `new Mutex(Global\…)` ctor **and** `WaitOne` are now both inside `try`. Added a `catch` for `UnauthorizedAccessException | WaitHandleCannotBeOpenedException | IOException` (kept `AbandonedMutexException` → owned). On ctor/wait failure the CLI no longer crashes before `AddInfrastructure`; it warns on stderr and degrades.
- On **lock timeout / not-acquired** (`acquired == false`, incl. ctor failure), the bootstrap no longer persists under no lock. It uses a transient in-memory key for that run only (no write) and warns on stderr, so two unsynchronized writers can never race to persist diverging keys. Switched to nullable `Mutex?` with manual dispose + best-effort `ReleaseMutex` in `finally`.
- Maps to **Gemini HIGH "Named Mutex Permission Issue"** and **Codex P2 "Stop when the bootstrap mutex times out."**

### MEDIUM 2 — Documented `TASKDECK_CONNECTORS__ENCRYPTIONKEY` silently ignored (`1948a520`, `a550fb0b`, `78b7a7af`, `34cb38dc`)
- `Program.cs` now calls `builder.Configuration.AddEnvironmentVariables("TASKDECK_")`, so `TASKDECK_CONNECTORS__ENCRYPTIONKEY` maps to `Connectors:EncryptionKey` (canonical `Connectors__EncryptionKey` still works via the default no-prefix provider).
- `DependencyInjection.cs` fail-fast message corrected from `TASKDECK_CONNECTORS__ENCRYPTIONKEY` → `Connectors__EncryptionKey` (correct for both API and CLI hosts).
- New end-to-end test `BoardsList_WithOperatorKeyViaTaskdeckEnvVar_HonorsOverrideAndDoesNotPersist`: on a clean machine with only `TASKDECK_CONNECTORS__ENCRYPTIONKEY` set, the bootstrap no-ops and never writes `appsettings.local.json` (the override is honored, not overwritten). This test fails without the `Program.cs` wiring, so it is load-bearing.
- Maps to **Codex P2 "Honor the documented TASKDECK key override."**

### MEDIUM 3 — Key file world-readable on Unix (`7f5f0de4`)
- `PersistKey` now calls `File.SetUnixFileMode(tempPath, UserRead | UserWrite)` (0600) before `File.Move`, on non-Windows only. The mode is applied to the temp file before the rename, so the final `appsettings.local.json` is never world-readable. No-op on Windows (NTFS ACL inheritance).
- Maps to **Codex P2 "Keep the CLI-generated key file private."**

### LOW — broaden `ResolveDataDirectory` catch (`7f5f0de4`)
- Catch widened from `ArgumentException` to `ArgumentException | PathTooLongException | NotSupportedException | IOException | SecurityException`, so a bad data-source path falls back to the working directory instead of crashing.
- Maps to **Gemini MEDIUM "Uncaught Exceptions in ResolveDataDirectory."**

### LOW — docstring accuracy + shared-DB footgun (`7f5f0de4`)
- Corrected the class docstring: removed the misleading "mirrors the API write pattern" claim and documented that the API persists ITS key next to the executable (`AppContext.BaseDirectory`) while the CLI writes next to the data dir — so a shared-DB / multi-host deployment auto-generates diverging keys; operators should set one explicit shared `Connectors__EncryptionKey`.
- Corrected the `EnsureConnectorEncryptionKey` docstring: the early no-op fires only for keys in live `IConfiguration` (env/appsettings); `appsettings.local.json` is not a registered config source — it is re-read each run inside `EnsureKeyOnDisk` (idempotent).
- Added `WarnIfSharedDbAutoGenerated`: emits a one-line **stderr** warning when the bootstrap auto-generates a key while pointed at a non-default data dir (`TASKDECK_CONNECTION_STRING` set), nudging operators toward an explicit shared key. stdout stays clean JSON.

### Already addressed in earlier branch commits (bot reviewed `51ee6565`, pre-dating these)
- **Gemini HIGH "Potential JSON Parse Crash"** — `ReadExisting` already uses `JsonValue.TryGetValue<string>` (no throw on non-string key) since `75448499`.
- **Gemini HIGH "Uncaught Exceptions in ReadConfigObject"** — `ReadExisting` already catches `IOException | UnauthorizedAccessException` on `File.ReadAllText` and `JsonException | InvalidOperationException` on parse since `75448499`, with a `PreserveFile` guard that avoids clobbering an unreadable-but-present key.

### Verification
- `dotnet build backend/Taskdeck.sln -c Release` → **0 warnings, 0 errors**.
- `dotnet test backend/tests/Taskdeck.Cli.Tests/Taskdeck.Cli.Tests.csproj -c Release` → **95 passed, 0 failed, 0 skipped** (includes the new env-override test).

Out of scope (not touched, per coordinator): API `FirstRunBootstrapper` key-location consistency, and re-architecting the CLI key-file path/config-source wiring.

PR #1180, opened 2026-06-06
Serialize cross-process Database.Migrate() via file advisory lock (#1164)

### LINKED ISSUE #1164: Serialize cross-process Database.Migrate() (single-owner / advisory lock)

Chris0Jeky:
## Summary

Closes #1164.

Each host (API, MCP-http, MCP-stdio, CLI) calls `Database.Migrate()` on startup with **no cross-process coordination**. #1130 enabled WAL + `busy_timeout` (a racing writer now *waits* instead of failing immediately), but two processes booting simultaneously can still both decide migrations are pending and contend on `__EFMigrationsHistory` (UNIQUE-key collision). This PR adds the remaining acceptance criterion: **serialize migration application across processes**.

## Lock design

New `SerializedMigrator.Migrate(DbContext, ILogger?)` in `Taskdeck.Infrastructure.Persistence` wraps `Database.Migrate()` with a **portable, cross-process file-based advisory lock**:

- Resolves the SQLite **Data Source** file path from `context.Database.GetDbConnection().DataSource`, falling back to parsing `GetConnectionString()` via `SqliteConnectionStringBuilder` (the live connection reports an empty `DataSource` once open, so the connection-string parse is the reliable source).
- Opens a sidecar lock file `"<dbpath>.migrate.lock"` via `new FileStream(lockPath, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None)` in a **retry-with-backoff loop** (catch `IOException`, 100ms sleep, bounded **30s** total). The winner applies migrations; every other process blocks on the lock, then acquires it and sees migrations already applied → no-op (`Migrate()` is idempotent). The `FileStream` is disposed in a `finally`; the lock file is intentionally left behind (cleanup is best-effort/optional).
- A .NET named `Mutex` is **deliberately not used** — it is not cross-process on Unix (per the issue).

**No single-owner model.** The file lock already satisfies "CLI/MCP do not race the API": all hosts coordinate on the same lock file, so only one applies at a time and the rest no-op. This keeps every host self-sufficient on a fresh DB (no host has to be running first), which a single-owner/skip model would not.

### Fallbacks (never block startup)
- **Non-SQLite provider** (future PostgreSQL per ADR-0023, or the EF Core InMemory provider): skip the lock, migrate directly (`IsSqlite()` guard).
- **In-memory / non-file** (`:memory:`, `file::memory:`, `Mode=Memory`, empty data source): skip the lock — there is no shared file to coordinate on (tests + ephemeral DBs must not stall).
- **Lock can't be acquired in 30s, or lock dir unwritable** (`UnauthorizedAccessException`/`DirectoryNotFoundException`): log a warning and proceed — `Migrate()` is idempotent and `busy_timeout` still serializes the actual write.

## Call sites updated (all 4)
- `backend/src/Taskdeck.Cli/Program.cs` — CLI (no logger: CLI stdout must stay clean JSON)
- `backend/src/Taskdeck.Api/Program.cs` — MCP-http mode and MCP-stdio mode (stdio logs to stderr, so the logger never corrupts the JSON-RPC stream)
- `backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs` — main API web host

## Tests
`backend/tests/Taskdeck.Api.Tests/SerializedMigratorTests.cs` (mirrors the #1130 `SqlitePragmaInterceptorTests` / `MigrationBootstrapTests` real-file-SQLite style):
- `Two_concurrent_migrators_apply_schema_exactly_once` — two DbContexts on the SAME fresh file DB, each calling the helper from a `Task.Run`, released together by a `Barrier`; asserts both complete without exception, the `Boards` table exists, and `__EFMigrationsHistory` contains each migration exactly once (no duplicate-key error).
- `Single_migrator_on_file_db_creates_sidecar_lock_file` — documents the `<dbpath>.migrate.lock` path resolution.
- `In_memory_database_migrates_without_creating_a_lock_file` — `:memory:` migrates with no sidecar lock file.

## Verification
- `dotnet build backend/Taskdeck.sln -c Release` → **0 errors**.
- `dotnet test … --filter "SerializedMigrator|MigrationBootstrap|SqlitePragmaInterceptor"` → **13 passed, 0 failed** (3 new).
- CLI smoke: `boards list` against a temp DB → exit 0, migrations applied via `SerializedMigrator`.

## Residual risks
- **Windows vs POSIX file-lock semantics**: `FileShare.None` is enforced by the OS on Windows and by `flock()` (advisory) + an in-process table on .NET/Unix; both serialize same-process threads and cross-process opens, which is what the concurrency test and production need.
- **Lock-file cleanup**: the sidecar `.migrate.lock` is intentionally left on disk (harmless; reused on next boot). A stale lock cannot wedge startup because the holder always releases via `finally`, and acquisition is bounded at 30s with a proceed-anyway fallback.
- **Unwritable lock dir**: degrades gracefully to the pre-#1164 behavior (WAL + busy_timeout), logged as a warning.

### DIFF
--- a/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
@@ -28,10 +28,12 @@ public static class PipelineConfiguration
                 "Configure ForwardedHeaders:KnownProxies or ForwardedHeaders:KnownNetworks.");
         }
 
+        // Apply EF Core migrations serialized across processes via a cross-process file lock
+        // so concurrent API/MCP/CLI startups apply the schema exactly once (#1164).
         using (var scope = app.Services.CreateScope())
         {
             var dbContext = scope.ServiceProvider.GetRequiredService<TaskdeckDbContext>();
-            dbContext.Database.Migrate();
+            SerializedMigrator.Migrate(dbContext, app.Logger);
         }
 
         if (app.Environment.IsDevelopment())
--- a/backend/src/Taskdeck.Api/Program.cs
+++ b/backend/src/Taskdeck.Api/Program.cs
@@ -105,11 +105,13 @@ if (args.Contains("--mcp"))
 
         var mcpHttpApp = mcpHttpBuilder.Build();
 
-        // Apply EF Core migrations before starting.
+        // Apply EF Core migrations before starting, serialized across processes via a
+        // cross-process file lock so the MCP HTTP host does not race the API/CLI (#1164).
         using (var scope = mcpHttpApp.Services.CreateScope())
         {
             var dbContext = scope.ServiceProvider.GetRequiredService<Taskdeck.Infrastructure.Persistence.TaskdeckDbContext>();
-            dbContext.Database.Migrate();
+            var migrationLogger = scope.ServiceProvider.GetRequiredService<ILogger<Program>>();
+            Taskdeck.Infrastructure.Persistence.SerializedMigrator.Migrate(dbContext, migrationLogger);
         }
 
         // Correlation ID propagation: honours client X-Request-Id header.
@@ -182,11 +184,14 @@ if (args.Contains("--mcp"))
         })
         .Build();
 
-    // Apply EF Core migrations before starting the MCP host (mirrors web mode behaviour).
+    // Apply EF Core migrations before starting the MCP host (mirrors web mode behaviour),
+    // serialized across processes via a cross-process file lock (#1164). In stdio mode logs
+    // go to stderr, so the logger never corrupts the stdout JSON-RPC stream.
     using (var scope = mcpHost.Services.CreateScope())
     {
         var dbContext = scope.ServiceProvider.GetRequiredService<Taskdeck.Infrastructure.Persistence.TaskdeckDbContext>();
-        dbContext.Database.Migrate();
+        var migrationLogger = scope.ServiceProvider.GetRequiredService<ILogger<Program>>();
+        Taskdeck.Infrastructure.Persistence.SerializedMigrator.Migrate(dbContext, migrationLogger);
     }
 
     await mcpHost.RunAsync();
--- a/backend/src/Taskdeck.Cli/Program.cs
+++ b/backend/src/Taskdeck.Cli/Program.cs
@@ -53,7 +53,10 @@ using var host = builder.Build();
 using (var startupScope = host.Services.CreateScope())
 {
     var dbContext = startupScope.ServiceProvider.GetRequiredService<TaskdeckDbContext>();
-    dbContext.Database.Migrate();
+    // Serialize migrations across processes (API/MCP/CLI) via a file lock (#1164).
+    // Pass no logger: CLI stdout must stay clean JSON (the helper never writes to stdout,
+    // but logging is suppressed here regardless).
+    SerializedMigrator.Migrate(dbContext);
 }
 
 var dispatcher = new CommandDispatcher(host.Services);
--- a/backend/src/Taskdeck.Infrastructure/Persistence/SerializedMigrator.cs
+++ b/backend/src/Taskdeck.Infrastructure/Persistence/SerializedMigrator.cs
@@ -1,0 +1,266 @@
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+
+namespace Taskdeck.Infrastructure.Persistence;
+
+/// <summary>
+/// Applies EF Core migrations under a portable, cross-process advisory lock so that
+/// multiple Taskdeck hosts (API, MCP-http, MCP-stdio, CLI) booting against the same
+/// SQLite file do not race to apply migrations on <c>__EFMigrationsHistory</c>.
+/// <para>
+/// #1130 enabled <c>WAL + busy_timeout</c> (a racing writer now waits instead of failing
+/// immediately with <c>SQLITE_BUSY</c>), which substantially mitigates the race. #1164 is
+/// the remaining acceptance criterion: actually <em>serialize</em> migration application so
+/// concurrent startups apply the schema exactly once.
+/// </para>
+/// <para>
+/// <b>Mechanism.</b> A sidecar lock file <c>"&lt;dbpath&gt;.migrate.lock"</c> is opened with
+/// <see cref="FileShare.None"/>; only one process can hold it at a time. The winner applies
+/// migrations; every other process blocks on the lock and, once it acquires it, sees the
+/// migrations already applied and no-ops (<see cref="RelationalDatabaseFacadeExtensions.Migrate"/>
+/// is idempotent). A .NET named <see cref="System.Threading.Mutex"/> is deliberately avoided:
+/// it is not cross-process on Unix.
+/// </para>
+/// <para>
+/// <b>Fallbacks (never block startup).</b> Non-relational/in-memory databases
+/// (<c>:memory:</c>, <c>Mode=Memory</c>, the EF Core InMemory provider) and non-SQLite
+/// providers skip the lock entirely — there is no shared file to coordinate on, and tests /
+/// ephemeral databases must not stall. If the lock cannot be acquired within the timeout, or
+/// the lock directory is unwritable, a warning is logged and migration proceeds anyway:
+/// <c>Migrate()</c> remains idempotent and <c>busy_timeout</c> still makes a racing writer wait.
+/// </para>
+/// </summary>
+public static class SerializedMigrator
+{
+    /// <summary>Default upper bound on how long to wait for the migration lock.</summary>
+    public static readonly TimeSpan DefaultLockTimeout = TimeSpan.FromSeconds(30);
+
+    private static readonly TimeSpan RetryDelay = TimeSpan.FromMilliseconds(100);
+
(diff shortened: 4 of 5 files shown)


Chris0Jeky:
## Adversarial Code Review

Self-review of the diff against logic, concurrency, resource-safety, clean-architecture, and test-coverage criteria.

### CRITICAL
- None.

### HIGH
- None. The fail-open behavior (lock not acquirable within timeout → log warning + proceed) is the prescribed design in #1164 ("`Migrate()` is itself idempotent under busy_timeout"), not a fail-open-where-fail-closed bug. Lock release is guaranteed via `finally`, so a crash mid-migration cannot wedge subsequent startups.

### MEDIUM
- **M1 — Test deadlock window.** `Two_concurrent_migrators_apply_schema_exactly_once` constructs each `DbContext` *inside* the task, after which both call `Barrier.SignalAndWait()`. If context construction threw before a task reached the barrier, the sibling would block on `SignalAndWait()` until the xUnit timeout (hang instead of fail). Low probability, but a real robustness gap in a concurrency test.
- **M2 — Fail-open path untested.** The "could not acquire the lock within the timeout → warn + proceed" branch is a core safety behavior of the helper and had no test. Behavior changes must ship with tests (DoD).

### LOW
- **L1 — Untested public overload.** `Migrate(DbContext, TimeSpan, ILogger?)` was only exercised transitively via the default overload. The M2 fix exercises it directly with an explicit short timeout.

### Bot Comments Addressed
- None present (no bot or human comments on the PR at review time).

### Summary
0 CRITICAL, 0 HIGH, 2 MEDIUM, 1 LOW. Nothing merge-blocking in the production logic; fixes harden the new test (remove the deadlock window) and close the coverage gap on the fail-open path. Fixes incoming in follow-up commits.

Chris0Jeky:

| M1 — Concurrency test could deadlock if a migrator task threw before reaching the `Barrier` | MEDIUM | `551dd444` | Contexts now constructed up front; only `SignalAndWait()` + `Migrate()` run inside the tasks, so no throw can strand the sibling. `Two_concurrent_migrators_apply_schema_exactly_once` passes (591 ms). |
| M2 — Fail-open path (lock not acquirable within timeout → warn + proceed) had no test | MEDIUM | `551dd444` | Added `Migrate_proceeds_with_a_warning_when_lock_cannot_be_acquired_within_timeout`: holds the sidecar lock with `FileShare.None`, calls the helper with a 300 ms timeout, asserts no throw + schema applied + a `Warning` was logged (via `InMemoryLogger`). Passes (1 s). |
| L1 — Public `Migrate(DbContext, TimeSpan, ILogger?)` overload only exercised transitively | LOW | `551dd444` | The M2 test now calls that overload directly with an explicit timeout. |

### Verification
- `dotnet build backend/tests/Taskdeck.Api.Tests` -c Release → **0 errors**.
- `dotnet test … --filter "FullyQualifiedName~SerializedMigrator"` → **4 passed, 0 failed** (was 3; +1 fail-open test).

All findings addressed (0 CRITICAL / 0 HIGH / 2 MEDIUM / 1 LOW — all fixed; none deferred or dropped). No bot or human comments were outstanding.

CI status: **PENDING** (re-triggered by the fix push; 18 checks queued at comment time). Earlier on the first push, Backend Architecture, SAST, Secret Scan, Dependency Security, and Docs Governance had already passed.

Chris0Jeky:
## Adversarial Review — Round 2 Fixes Applied

Second independent adversarial pass (1 MEDIUM + 5 LOW). All findings fixed; the three open bot inline comments are also addressed. No findings deferred or dropped.

Commits: `28fc3606` (src), `15df5915` (tests).

| Finding | Severity | Fix Commit | How / Verified |
|---------|----------|-----------|----------------|
| Fail-open path can throw on concurrent `Migrate()` — unlocked fallback called `Database.Migrate()` unguarded; a racing loser could throw `SqliteException` ("table already exists") or a duplicate `__EFMigrationsHistory` PK insert and crash startup. `busy_timeout` only serializes `SQLITE_BUSY`, not logical DDL conflicts. | MEDIUM | `28fc3606` | New `MigrateWithoutLock` wraps the lock-free `Migrate()` in `catch (Exception ex) when (NoMigrationsPending(context))`: if `GetPendingMigrations()` is now empty (a concurrent winner applied everything) → log a Warning + swallow as a no-op; if non-empty (or the pending set can't be read) → the exception filter declines and the original failure propagates with its stack trace intact. The **locked** path keeps `Migrate()` unguarded (no concurrent conflict possible). Misleading type/`AcquireLock` comments corrected: the catch+re-check — not `busy_timeout` — delivers the no-throw fallback. |
| Lock-wait deadline used wall-clock `DateTime.UtcNow` (clock step skews the wait). | LOW | `28fc3606` | `AcquireLock` now uses a monotonic `System.Diagnostics.Stopwatch` (`StartNew()`, loop while `Elapsed < timeout`). |
| Path-aliasing not documented — lock keyed on `Path.GetFullPath` (lexical only; no symlink/junction canonicalization). | LOW | `28fc3606` | Added an inline LIMITATION note: differently-spelled paths to the same physical file key to different lock files and are not serialized; residual race bounded by WAL + busy_timeout + the unlocked-Migrate re-check. |
| Concurrency test wording overstated "cross-process" (it uses two in-process `Task.Run` threads). | LOW | `15df5915` | Softened class/test docs to "in-process proxy for cross-process" (noting `FileShare.None` is OS-enforced). Also looped the concurrent migrate over **4 fresh databases** to shrink the probabilistic window where the threads never actually contend. |
| Single-migrator test pinned an optional behavior (asserted the `.migrate.lock` file still exists, but cleanup is documented as best-effort). | LOW | `15df5915` | Replaced the brittle file-exists assertion with asserting the helper's `"acquired migration lock"` **Debug** log entry (via `InMemoryLogger`) plus schema-applied (`Boards` table). |
| Fail-open happy-path had no no-warning coverage (a regression that silently always failed open would pass unnoticed). | LOW | `15df5915` | Same test now also asserts a healthy single file-DB migrate emits **NO** Warning/Error, so a silent always-degrade regression is caught. |

### Bot comments addressed
| Bot | Severity | Status |
|-----|----------|--------|
| `gemini-code-assist` — SQLite URI data sources (`Data Source=file:taskdeck.db?cache=shared`) passed raw to `Path.GetFullPath` throw on Windows / produce a literal `?` lock file on Unix | HIGH | **Fixed** in `28fc3606`: `Uri.TryCreate(..., UriKind.Absolute)` + `uri.IsFile` extracts `LocalPath` before `Path.GetFullPath`. Relative/non-URI data sources are unaffected (`TryCreate` returns false). |
| `gemini-code-assist` — `Barrier.SignalAndWait()` with no timeout can hang CI if a sibling dies pre-barrier | MEDIUM | **Fixed** in `15df5915`: bounded `SignalAndWait(TimeSpan.FromSeconds(10))`; a missed barrier throws `TimeoutException` (fail fast, never hang). |
| `chatgpt-codex-connector` (P2) — fail-open path bypasses the lock and re-introduces the `__EFMigrationsHistory`/writer race | P2 | **Addressed** by the MEDIUM fix above: the unlocked path no longer crashes on a concurrent conflict — it re-checks the pending set and only swallows when the schema is already fully applied, otherwise surfaces the real failure. |

### Verification
- `dotnet build backend/Taskdeck.sln -c Release` → **0 errors** (11 pre-existing warnings in unrelated test files).
- `dotnet test backend/tests/Taskdeck.Api.Tests/...csproj -c Release --no-build --filter "FullyQualifiedName~SerializedMigrator"` → **4 passed, 0 failed, 0 skipped** (~5 s).

### Residual / not done
- The unlocked-Migrate **swallow** branch and the unwritable-dir / non-SQLite `ResolveLockPath` branches are not unit-tested: deterministically forcing a concurrent DDL conflict on the unlocked path (or an unwritable temp dir) is environment-dependent and brittle. The logic is small and exception-filter-gated. The existing timeout test still exercises `MigrateWithoutLock` end-to-end (its success path).

Issue #1181, opened 2026-06-06
Packaged binary: GET / returns 401 instead of the SPA shell (bare root not served under global auth fallback)


Chris0Jeky:
## Summary
The published self-contained desktop binary serves the SPA for **every path except the bare root `/`**, which returns **HTTP 401** (`{"errorCode":"Unauthorized"}`) instead of `index.html`. `/` is the URL the first-run flow logs and the browser auto-opens to, so a user navigating to `http://localhost:5000` gets a JSON 401 error on first launch.

Discovered while implementing the #1123 AC2 post-publish smoke test (release-desktop.yml). The AC2 smoke asserts `GET /` returns the SPA, so this currently blocks a green release run (the gate working as intended).

## Reproduction (local, win-x64 self-contained single-file publish)
Run the published exe (cwd = publish dir) with throwaway secrets, then:
```
GET /index.html      -> 200   (static files serve the shell)
GET /board           -> 200   (MapFallbackToFile SPA fallback, AllowAnonymous)
GET /login           -> 200   (SPA fallback)
GET /                -> 401   (bare root)
GET /health/ready    -> 200
```
Reproduces identically in both `ASPNETCORE_ENVIRONMENT=Production` (default) and `Development`, so it is environment-independent. The packaged SPA-serving path has never shipped, so this was never exercised before.

## Root cause
`backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs` uses `UseDefaultFiles()` + `UseStaticFiles()` + `app.MapFallbackToFile("index.html").AllowAnonymous()`. The SPA fallback's catch-all route (`{*path:nonfile}`) does not match the **empty** root path, so `/` matches no endpoint and is rejected by the global `SetFallbackPolicy(RequireAuthenticatedUser)` (#1132 AC4) -> 401. Non-empty client routes (`/board`) do match the fallback and are served. `UseDefaultFiles` is not rewriting `/` -> `/index.html` in this configuration.

Existing coverage gap: `backend/tests/Taskdeck.Api.Tests/FallbackPolicyTests.cs` only exercises a non-empty route (`/some/client/route`); the bare root `/` is untested.

## Recommended fix
Ensure the bare root serves the SPA shell anonymously, e.g. add alongside the existing fallback:
```csharp
// Bare root must serve the SPA shell; the {*path:nonfile} fallback doesn't match the empty path.
app.MapFallbackToFile("/", "index.html").AllowAnonymous();
```
[...]

PR #1145, opened 2026-05-31
Release build: align local scripts with CI flags, drop stale PKG-01 warning

Chris0Jeky:
## Summary
First slice of #1123 (v0.1.0 release validation). Makes the local release-build scripts produce the same artifact CI does, and removes a stale instruction that would block shipping.

- **Flag alignment**: `scripts/build-release.{sh,ps1}` used `-p:PublishTrimmed=true` and omitted `IncludeNativeLibrariesForSelfExtract`, diverging from the validated CI workflow (`.github/workflows/release-desktop.yml:124-126` → `PublishTrimmed=false`, `IncludeNativeLibrariesForSelfExtract=true`). Trimming silently breaks EF Core migrations / DI / System.Text.Json / SignalR, and the SQLite native lib (`e_sqlite3`) needs native-lib extraction in single-file self-contained mode — so the local scripts likely produced a **broken exe**. Now matched to CI; dropped the irrelevant `TrimmerRootAssembly`.
- **Stale warning removed**: both scripts carried a `# Do not ship ... until PKG-01 (#533) is merged` block claiming the binary won't serve the SPA. PKG-01 **is** merged — `PipelineConfiguration.cs:80` (`UseStaticFiles`) and `:135` (`MapFallbackToFile("index.html")`) wire SPA serving. A maintainer obeying the stale comment would needlessly block the release. Replaced with an accurate note.

## Verification
- `bash -n scripts/build-release.sh` → OK.
- PowerShell tokenizer parse of `build-release.ps1` → OK.
- No residual `PublishTrimmed=true` / `TrimmerRootAssembly` / "do not ship" refs; new flags present in both scripts.
- CI flags cross-checked against `release-desktop.yml`.

## Remaining in #1123 (follow-ups)
Post-publish launch/smoke test in `release-desktop.yml` (boot exe, poll `/health/ready`, GET `/`), and the actual `v0.1.0` tag + clean-VM smoke. Left for a separate change.

Not merging — for review. Refs #1123.

Issue #1123, opened 2026-05-30
Ship & validate the v0.1.0 "First Light" release (fix build scripts, add smoke test, push tag)


Chris0Jeky:
## Context
All packaging machinery exists (PKG-01..04 closed, `release-desktop.yml`, `scripts/build-release.*`) but **no release has ever shipped** — verified: `git tag` count = 0, `gh release list` empty, the release workflow has never run. Shipping a downloadable v0.1.0 to get the first 5-10 users is the #1 goal in `docs/strategy/00_MASTER_STRATEGY.md:45-61`.

## Evidence
- `git tag` = 0; no GitHub releases; `release-desktop.yml` has no runs.
- `scripts/build-release.sh:142-149` / `.ps1:142-149` use `--self-contained true -p:PublishSingleFile=true -p:PublishTrimmed=true` and **omit** `IncludeNativeLibrariesForSelfExtract`, diverging from CI `release-desktop.yml:119-127` (`PublishTrimmed=false`, `IncludeNativeLibrariesForSelfExtract=true`). The script's own comment (`build-release.sh:139-141`) warns trimming breaks EF migrations/SignalR/reflection; EF Sqlite `e_sqlite3` native lib needs extraction in single-file mode; `Program.cs:112` runs `Database.Migrate()` at startup. Likely produces a broken exe.
- `scripts/build-release.ps1:43-46` contains a now-FALSE blocking instruction ("Do not ship until PKG-01 merged") though SPA serving is wired (`PipelineConfiguration.cs:80` UseStaticFiles, `:135` MapFallbackToFile).
- `release-desktop.yml:136-143` only checks `wwwroot/index.html` exists — no launch/smoke test of the published binary.

## Acceptance criteria
- [ ] Local build scripts match validated CI flags (`PublishTrimmed=false`, `IncludeNativeLibrariesForSelfExtract=true`); stale PKG-01 warning blocks removed.
- [ ] Release workflow adds a post-publish smoke step per runner-native RID: launch exe with temp data dir + dummy secrets, poll `/health/ready`, GET `/` for index.html, shut down; fail release on error.
- [ ] v0.1.0 binary smoke-tested end-to-end on a clean VM (first-run auto-secrets, browser launch, capture→review→apply).
- [ ] `v0.1.0` tag pushed, workflow green, GitHub Release with downloads + checksums published.

[...]

PR #1182, opened 2026-06-06
Add post-publish smoke test to release-desktop workflow (#1123 AC2)

Chris0Jeky:
## What
Implements **#1123 AC2** only: a **post-publish smoke step** in `release-desktop.yml`'s `build-backend` matrix job, inserted **after "Verify frontend assets present"** and **before** the version/archive steps. It launches the just-published self-contained exe, polls `/health/ready`, asserts `GET /` serves the SPA `index.html`, and **fails the release job** on any error. The process is always killed via a `trap ... EXIT`.

Part of #1123. AC1 shipped in #1145; **AC3 (clean-VM) and AC4 (tag push) remain and need maintainer action.**

## How the exe is launched
- `shell: bash` (one script for Windows + Linux + macOS; bash/curl/openssl are on all three runners).
- Exe name resolved per OS: `Taskdeck.Api.exe` on Windows, `Taskdeck.Api` elsewhere.
- Launched with **cwd = the publish dir** so ASP.NET Core resolves the content root there and serves `wwwroot/index.html` + `appsettings.json` (content root defaults to the CWD, not the exe folder).
- Env: `ASPNETCORE_URLS=http://127.0.0.1:5099`; `ConnectionStrings__DefaultConnection` -> a SQLite file in a `mktemp -d` temp dir (native path via `cygpath` on Windows); `FirstRun__AutoOpenBrowser=false`.
- **Secrets first-run needs:** the exe defaults to `ASPNETCORE_ENVIRONMENT=Production`. First-run auto-generates the **JWT secret**, but in Production the **connector encryption key is NOT auto-generated** and `ValidateProductionSecrets` hard-fails without it. So the step passes throwaway `Jwt__SecretKey` and `Connectors__EncryptionKey` (`openssl rand -base64 48` / `-base64 32`; the connector key must base64-decode to exactly 32 bytes per `AesCredentialEncryptionService`). Supplying the JWT secret also stops first-run from writing `appsettings.local.json` into the publish dir, and the DB lives in a temp dir, so **nothing pollutes the archive** built next.

## Assertions
- `GET /health/ready` must return success (bounded ~60s; `curl -f` so a 503 NotReady keeps polling). This gates on EF migrations applied + DB reachable + workers heartbeating.
- `GET /` must return **HTTP 200** with a SPA marker (`<div id="app"` or `<!doctype html`). On failure it prints actionable diagnostics (`/index.html` and `/board` status codes + body + app-log tail).

## Runner-native RID gating (why osx-x64 is skipped)
The matrix builds win-x64 (windows-latest), linux-x64 (ubuntu-latest), osx-x64 + osx-arm64 (both macos-latest). **macos-latest runners are arm64**, so the **osx-x64** binary is cross-arch and may not launch. The step is gated with `if: matrix.rid != 'osx-x64'` (commented inline). win-x64, linux-x64, and **osx-arm64** are exercised; osx-x64 is intentionally skipped.

## Cleanup
`trap cleanup EXIT` kills the launched PID (and `taskkill //F //IM Taskdeck.Api.exe` on Windows for the MSYS->Win32 PID-mapping gap) and removes the temp dir, so the process never leaks and runs in a finally manner regardless of success/failure.

## Verification
- `python -c "import yaml; yaml.safe_load(...)"` -> **YAML OK**.
- **Locally smoke-tested** the script against a real `dotnet publish` win-x64 self-contained single-file exe (the runner-native RID on this box): exe boots in Production with the throwaway secrets, `/health/ready` -> 200 in ~1s, publish dir stays clean (no new files), PID cleanup works.

## :warning: Finding surfaced by this smoke (tracked as #1181)
Local testing revealed the packaged binary returns **`GET /` -> 401** (bare root not served), while `/index.html`, `/board`, `/login` all return 200. Root cause: the `MapFallbackToFile("index.html")` catch-all (`{*path:nonfile}`) doesn't match the empty root path, so `/` hits the global `RequireAuthenticatedUser` fallback policy. Environment-independent (Prod + Dev). This is a real first-run UX bug (`/` is the auto-opened/logged URL) in the SPA-serving/first-run domain (AC3-adjacent), **out of AC2 scope** — filed as **#1181** with a recommended fix.

**Decision point for review:** the smoke asserts `GET /` strictly per the AC2 spec, so a release run stays red until #1181 is fixed (the gate doing its job — a first release whose root URL 401s shouldn't ship). If you'd rather unblock shipping immediately and track `/` separately, I can relax the assertion to "SPA shell served" (assert `/index.html` + a client route, warn-and-track on bare `/`).

Could **not** verify: macOS/Linux runtime behavior (no local runners) — gated/traced by logic only. Residual risks: port 5099 conflict (low), first-run startup time vs. the 60s budget (observed ~1s locally), and the #1181 root-path 401.

### DIFF
--- a/.github/workflows/release-desktop.yml
+++ b/.github/workflows/release-desktop.yml
@@ -142,6 +142,154 @@ jobs:
           fi
           echo "Frontend assets verified ($(find publish/${{ matrix.rid }}/wwwroot -type f | wc -l) files)"
 
+      # -----------------------------------------------------------------------
+      # Post-publish smoke test (#1123 AC2)
+      #
+      # Proves the just-published self-contained executable actually boots and
+      # serves the product before we archive and ship it:
+      #   1. GET /health/ready returns success  -> EF migrations applied, DB
+      #      reachable, workers heartbeating (PipelineConfiguration runs the
+      #      SerializedMigrator at startup; HealthController.ReadyCheck gates on
+      #      DB + queue + workers).
+      #   2. GET /                returns the SPA index.html -> same-origin
+      #      wwwroot serving works (UseDefaultFiles + UseStaticFiles).
+      # The process is always killed (trap ... EXIT) so it never leaks, and any
+      # failure (no boot, never ready, / not HTML) fails the release job.
+      #
+      # Runner-native RIDs only: the GitHub macos-latest runner is arm64, so the
+      # osx-x64 binary is cross-arch and may not launch there. It is SKIPPED via
+      # the `if:` gate below; osx-arm64 is the macOS RID exercised instead.
+      # win-x64 (windows-latest) and linux-x64 (ubuntu-latest) are native.
+      #
+      # shell: bash so one script covers Windows + Linux + macOS (bash ships on
+      # the Windows runner). curl + openssl are available on all three runners.
+      #
+      # Secrets: a downloaded desktop exe runs with ASPNETCORE_ENVIRONMENT unset,
+      # which defaults to Production. First-run auto-generates the JWT secret, but
+      # in Production the connector encryption key is NOT auto-generated and
+      # FirstRunBootstrapper.ValidateProductionSecrets hard-fails without it. We
+      # therefore pass throwaway Jwt__SecretKey and Connectors__EncryptionKey env
+      # vars (generated with openssl; the connector key must base64-decode to
+      # exactly 32 bytes per AesCredentialEncryptionService). Supplying the JWT
+      # secret also stops first-run from writing appsettings.local.json into the
+      # publish dir, and the DB is pointed at a temp dir, so nothing pollutes the
+      # archive built in the next step.
+      - name: Smoke test published executable
+        if: matrix.rid != 'osx-x64'  # cross-arch on the arm64 macOS runner — not launchable there
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          PORT=5099
+          URL="http://127.0.0.1:${PORT}"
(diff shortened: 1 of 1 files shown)


Chris0Jeky:

### Review A (correctness + security focus)
**0 findings.** Traced the smoke test logic against the full backend startup pipeline:
- Secret handling: `openssl rand` produces correct key sizes for JWT (≥32 bytes) and connector encryption (exactly 32 bytes after base64 decode)
- Production bootstrapper: env vars satisfy `ValidateProductionSecrets`, no `appsettings.local.json` pollution
- Health endpoint: `[AllowAnonymous]` on `ReadyCheck` confirmed
- Process lifecycle: `exec` + `trap cleanup EXIT` correct
- Cross-arch skip: `osx-x64` on arm64 runners correctly gated

### Review B (workflow + shell safety focus)
**0 findings.** Verified:
- YAML validity: pure 148-line insertion, parses cleanly
- Shell safety: `cygpath -w` for Win32 paths, `taskkill //F //IM` MSYS escape, `kill -0` skip on Windows
- SPA serving: `UseStaticFiles` before `UseAuthorization` confirmed
- DB isolation: temp dir, not publish dir
- Cleanup: `trap ... EXIT` ensures kill + temp-dir removal on all paths
- Archive integrity: no state leaks into the release artifact

### Existing comments
- Gemini Code Assist: quota limit warning (not a review finding, no action needed)

### Verdict
**Clean — ready to merge.** The smoke test correctly validates that the published executable boots, runs migrations, passes production secret validation, and serves the SPA.

PR #1183, opened 2026-06-06
Fix bare root '/' returning 401 under global FallbackPolicy (#1181)

### LINKED ISSUE #1181: Packaged binary: GET / returns 401 instead of the SPA shell (bare root not served under global auth fallback)

Chris0Jeky:
## Summary
- **Bug**: `GET /` returned HTTP 401 instead of the SPA shell because the `{*path:nonfile}` catch-all in `MapFallbackToFile` does not match the empty root path, causing it to fall through to the global `RequireAuthenticatedUser` fallback policy (#1176 AC4).
- **Fix**: Added an explicit `MapGet("/")` endpoint with `.AllowAnonymous()` that serves `index.html` from wwwroot (or returns 404 if absent), matching the `no-cache` policy applied to `index.html` by the static files middleware.
- **Test**: Added `BareRoot_NotBlockedByFallbackPolicy_WhenAnonymous` regression test in `FallbackPolicyTests`.

Closes #1181

## Test plan
- [x] `FallbackPolicyTests` — all 6 tests pass (5 existing + 1 new)
- [ ] Verify `GET /` returns 200 with SPA shell on a published self-contained binary
- [ ] Verify `GET /board`, `GET /login` still work (SPA fallback unchanged)
- [ ] Verify `GET /api/boards` still returns 401 without auth (FallbackPolicy intact)

### DIFF
--- a/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
@@ -122,17 +122,21 @@ public static class PipelineConfiguration
         app.MapControllers();
         app.MapHub<BoardsHub>("/hubs/boards");
 
-        // Bare root "/" serves the SPA shell. The {*path:nonfile} catch-all in MapFallbackToFile
-        // does not match the empty path, so "/" needs its own endpoint (#1181).
-        app.MapGet("/", async (HttpContext ctx, IWebHostEnvironment env) =>
+        // Bare root "/": in production, UseDefaultFiles + UseStaticFiles (above) rewrites "/" to
+        // "/index.html" and serves it before routing. When wwwroot is absent (dev/test without a
+        // frontend build), that middleware no-ops and the request reaches routing. The {*path:nonfile}
+        // catch-all in MapFallbackToFile does not match the empty path, so this explicit endpoint
+        // prevents "/" from hitting the global RequireAuthenticatedUser FallbackPolicy (#1181).
+        var webRootProvider = app.Environment.WebRootFileProvider;
+        app.MapGet("/", async (HttpContext ctx) =>
         {
-            var file = env.WebRootFileProvider.GetFileInfo("index.html");
+            var file = webRootProvider.GetFileInfo("index.html");
             if (!file.Exists)
             {
                 ctx.Response.StatusCode = 404;
                 return;
             }
-            ctx.Response.ContentType = "text/html";
+            ctx.Response.ContentType = "text/html; charset=utf-8";
             ctx.Response.Headers["Cache-Control"] = "no-cache";
             await ctx.Response.SendFileAsync(file);
         }).AllowAnonymous();
--- a/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
+++ b/backend/src/Taskdeck.Api/Extensions/PipelineConfiguration.cs
@@ -122,6 +122,21 @@ public static class PipelineConfiguration
         app.MapControllers();
         app.MapHub<BoardsHub>("/hubs/boards");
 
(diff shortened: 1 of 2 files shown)


Chris0Jeky:

### Review A findings
- **A-1 (LOW)**: Test comment said "MapFallbackToFile mapping" but implementation uses `MapGet` — **fixed** in `16278825`

### Review B findings
- **B-1 (MEDIUM)**: `MapGet("/")` is dead code in production (UseDefaultFiles handles it) — **fixed**: clarified comment that this is a dev/no-wwwroot fallback (`16278825`)
- **B-2 (MEDIUM)**: Missing `charset=utf-8` in Content-Type — **fixed**: now `text/html; charset=utf-8` (`16278825`)
- **B-3 (MEDIUM)**: Misleading test comment — **fixed** in `16278825`
- **B-4 (MEDIUM)**: Test assertion too weak — **addressed**: assertion stays as `NotBe(Unauthorized)` because the test host may or may not have wwwroot/index.html (returns 200 with frontend build, 404 without). Comment updated to document this.
- **B-5 (MEDIUM)**: Per-request DI resolution of IWebHostEnvironment — **fixed**: captured `WebRootFileProvider` in closure at startup (`16278825`)

### Existing comments
- No bot or human review comments found on this PR.

All findings addressed in commit `16278825`.

PR #1185, opened 2026-06-06
Fix N+1 notification deletion in AccountDeletionService (#1172)

### LINKED ISSUE #1172: AccountDeletionService deletes notifications via a single capped fetch (no paging loop)
Surfaced by the adversarial review of PR #1171 (out of scope there; tracked per review policy).

`AccountDeletionService` (backend/src/Taskdeck.Application/Services/AccountDeletionService.cs:104) fetches notifications with a single `GetByUserIdAsync(userId, limit: 100000)` call and iterates to delete. If a user ever held > 100000 notifications, deletion would be **incomplete** (a theoretical GDPR-completeness gap).

This is **pre-existing** — not introduced by #1171. The prior in-memory impl also capped at `Take(100000)`; the new SQL `LIMIT 100000` preserves identical behavior.

### Fix
Page through notifications in a delete loop (like `DataExportService.StreamNotificationsAsync`, which uses a correct `while(true)` offset loop with `StreamPageSize=500` and no hard cap) instead of relying on one large `LIMIT`.

### Severity
LOW — > 100000 notifications/user is implausible and the user-facing read path is properly bounded. Defensive correctness for account deletion.

Chris0Jeky:
## Summary

- **Add `DeleteByUserIdAsync`** to `INotificationRepository` — a batched bulk-delete method that replaces the unbounded fetch-then-loop-delete pattern
- **Implement batched SQL DELETE** in `NotificationRepository` using a loop with batch size 1000, following the `AuditLogRepository.DeleteOldEntriesAsync` pattern (supports both SQLite LIMIT and SQL Server CTE+TOP)
- **Replace the N+1 delete** in `AccountDeletionService` (lines 103-110) with a single call to the new repository method
- **Update unit tests** to mock and verify the new `DeleteByUserIdAsync` method

Closes #1172

## Test plan

- [x] `dotnet build backend/Taskdeck.sln -c Release` — 0 errors, 0 warnings
- [x] All 25 `AccountDeletionServiceTests` pass
- [x] Architecture tests pass (no layer violations)
- [ ] CI green on PR

### DIFF
--- a/backend/src/Taskdeck.Application/Interfaces/INotificationRepository.cs
+++ b/backend/src/Taskdeck.Application/Interfaces/INotificationRepository.cs
@@ -21,4 +21,11 @@ public interface INotificationRepository : IRepository<Notification>
         Guid userId,
         Guid? boardId = null,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Deletes all notifications for a user in batched SQL DELETEs to avoid
+    /// unbounded memory and N+1 single-row deletes.
+    /// </summary>
+    /// <returns>Total number of deleted rows.</returns>
+    Task<int> DeleteByUserIdAsync(Guid userId, CancellationToken cancellationToken = default);
 }
--- a/backend/src/Taskdeck.Application/Services/AccountDeletionService.cs
+++ b/backend/src/Taskdeck.Application/Services/AccountDeletionService.cs
@@ -100,14 +100,9 @@ public class AccountDeletionService : IAccountDeletionService
             var auditLogs = await _unitOfWork.AuditLogs.GetByUserAsync(userId, limit: 100000, cancellationToken: cancellationToken);
             var auditLogsAnonymized = auditLogs.Count();
 
-            // 2. Delete notifications (personal data)
-            var notifications = await _unitOfWork.Notifications.GetByUserIdAsync(userId, limit: 100000, cancellationToken: cancellationToken);
-            var notificationsDeleted = 0;
-            foreach (var notification in notifications)
-            {
-                await _unitOfWork.Notifications.DeleteAsync(notification, cancellationToken);
-                notificationsDeleted++;
-            }
+            // 2. Delete notifications (personal data) — batched SQL DELETE avoids
+            //    unbounded memory from fetching all rows and N+1 single-row deletes.
+            var notificationsDeleted = await _unitOfWork.Notifications.DeleteByUserIdAsync(userId, cancellationToken);
 
             // 3. Delete capture/inbox items (personal data)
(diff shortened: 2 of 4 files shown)


Chris0Jeky:
## Adversarial Code Review

### CRITICAL
- None

### HIGH
- **H1: EF Core change tracker stale after raw SQL DELETE.** The new `DeleteByUserIdAsync` uses `ExecuteSqlRawAsync` which bypasses the EF Core change tracker. If any Notification entities happen to be tracked in the same `DbContext` instance (e.g. from an earlier query in the same transaction), the change tracker will not know they were deleted. In the current `AccountDeletionService` flow this is safe because notifications are never queried via EF before the batch delete call. However, this is fragile — a future refactor that queries notifications before deleting them could hit a stale-tracker bug. **Recommendation:** Add a brief comment in the repository method documenting this caveat, consistent with the existing `DeleteOldEntriesAsync` pattern which also has this characteristic. Severity: HIGH because the current code is safe but the implicit invariant is undocumented. **Fix: add comment.**

### MEDIUM
- **M1: Missing `CancellationToken.None` fallback consistency.** The `AuditLogRepository.DeleteOldEntriesAsync` pattern this code follows is used outside transactions with a fresh scope. Here, `DeleteByUserIdAsync` runs inside a transaction in `AccountDeletionService`. If cancellation fires mid-batch, the partial delete is rolled back by the transaction — which is correct. But the loop exit via `cancellationToken.IsCancellationRequested` silently returns a partial count that may not reflect reality after rollback. This is acceptable because the caller never uses the count after a rollback (the catch block handles it), but a comment would clarify the intent. **Fix: add clarifying comment.**

### LOW
- **L1: Test coverage gap — no test for `DeleteByUserIdAsync` returning a large count.** The test only verifies count=3 and count=0. A test with a larger count (confirming the returned value flows through to `NotificationsDeleted`) would be slightly more robust, but the existing tests are sufficient for the contract. **No fix needed — the two existing test cases cover the interface contract.**

### Bot Comments Addressed
- None — no bot or human comments exist on this PR yet. CI checks are still pending.

### Summary
- CRITICAL: 0, HIGH: 1, MEDIUM: 1, LOW: 1
- The HIGH finding is about an undocumented invariant (change tracker bypass), not a current bug. The code is safe as-is but should be documented.
- Not merge-blocking — fixes are documentation/comments only.

Chris0Jeky:

| H1: EF Core change tracker bypass undocumented | HIGH | `f5f5fcce` | build clean, 25/25 tests pass |
| M1: Cancellation mid-batch safety undocumented | MEDIUM | `f5f5fcce` | build clean, 25/25 tests pass |
| L1: Test coverage for large count | LOW | No fix needed — existing tests cover the contract | N/A |

All findings addressed. CI status: PENDING (new push just landed).

PR #1188, opened 2026-06-06
Add ADR-0037: Idempotency-key contract for automation proposal operations (#1149)

### LINKED ISSUE #1149: Define & document AutomationProposalOperation idempotency-key contract (per-owner scoping + replay vs 409)
## Context

Surfaced by the pass-2 adversarial review of PR #1146 (which correctly maps a duplicate `AutomationProposalOperation.IdempotencyKey` UNIQUE violation from a 500 to a 409). Two related design observations are out of scope for that focused fix but should be decided and documented.

### 1. The operation idempotency-key unique index is process-global (cross-tenant)

`AutomationProposalOperationConfiguration.cs:54-55` declares `builder.HasIndex(apo => apo.IdempotencyKey).IsUnique()` — a single-column unique index with no `ProposalId` / `RequestedByUserId` scoping and no filter (migration `IX_AutomationProposalOperations_IdempotencyKey`, `unique: true`). The entity carries no user/tenant field, so the key namespace is global across all proposals and users.

Effect: two different authenticated users who both submit a human-friendly key like `"key1"` collide globally; the second now receives a 409 for a semantically-unrelated operation, which is also a minor cross-tenant existence oracle for that key string.

Real-world severity is **LOW** today: every production path uses collision-resistant keys — capture uses `SHA256(captureItemId:sequence:normalizedTitle)` plus a source-reference replay guard; chat/planner/MCP `Propose*` executors use `Guid.NewGuid()`; inbox-triage/digest use item/board-GUID-scoped keys. Only a hand-crafted simple key (as in the old adversarial tests) can organically collide.

**Decision needed:** scope the uniqueness to `(ProposalId, IdempotencyKey)` or `(RequestedByUserId, IdempotencyKey)` so idempotency is per-owner, OR document that the global guard is intentional (schema comment + ADR).

### 2. No true idempotent-replay contract on the manual create endpoint

The whole point of an idempotency key is usually that re-submitting the same key safely returns the **original** result (replay), not an error. `AutomationProposalService.CreateProposalAsync` does no pre-check on operation `IdempotencyKey`; a duplicate now returns 409 (better than the prior 500, and a correct semantic for a unique-constraint violation, but not replay).

Prior art in the repo: the **execute** endpoint honors a true replay contract via the `Idempotency-Key` HTTP header (`AutomationExecutorService.cs:79` treats already-applied proposals as success), and `CaptureTriageService` replays by source-reference. The operation-level key on the manual create endpoint has no such replay path.

**Decision needed:** either (a) document that operation `IdempotencyKey` is a uniqueness guard (not a replay key) and 409 is the deliberate response, or (b) implement true replay (look up the existing operation/proposal by key and return the original 200/201). This is a hard-to-reverse public-API contract decision (200 vs 201, which proposal to return on partial multi-op key collision, transactional handling) and warrants an **ADR** under `docs/decisions/`.

## Acceptance criteria
- [ ] Write an ADR capturing the chosen idempotency-key contract (scoping + replay-vs-409).
- [ ] If per-owner scoping is chosen: migration + config change + test; otherwise add a schema comment documenting the global guard rationale.
- [ ] If replay is chosen: implement look-up-and-return-original on the create path with tests; otherwise document the 409 contract for clients.

_Source: pass-2 adversarial review of #1146 (2026-05-31). Not merge-blocking for #1146._

Chris0Jeky:
## Summary
- Documents the existing idempotency-key contract as ADR-0037
- Covers: key format (free-form 1-100 chars), global uniqueness scope, 409 Conflict on duplicate, execution-level replay safety, caller-chosen generation strategies
- Adds entry to `docs/decisions/INDEX.md`

Closes #1149

## Test plan
- [x] Docs-only change — no code modified
- [x] ADR follows template from `docs/decisions/README.md`
- [x] INDEX.md updated with correct number and link

### DIFF
--- a/docs/decisions/ADR-0037-idempotency-key-contract.md
+++ b/docs/decisions/ADR-0037-idempotency-key-contract.md
@@ -1,0 +1,108 @@
+# ADR-0037: Idempotency-Key Contract for Automation Proposal Operations
+
+- **Status**: Accepted
+- **Date**: 2026-06-06
+- **Deciders**: Repository maintainers
+
+## Context
+
+Automation proposal operations carry an `IdempotencyKey` field (max 100 characters, non-empty, enforced at the domain level). The key serves two purposes: preventing duplicate operations from being persisted, and enabling safe replay of proposal execution.
+
+Prior to PR #1146, a duplicate key insertion produced an unhandled `DbUpdateException` surfacing as HTTP 500. The fix translated this into a 409 Conflict response, but the contract governing key format, scope, and replay semantics was never formally documented. Different callers generate keys using different patterns, and the uniqueness constraint is global (not scoped to a user or board), which has implications for key collision and cross-user visibility.
+
+## Decision
+
+### Key format
+
+The idempotency key is a free-form string of 1–100 characters. Callers choose a generation strategy appropriate to their domain:
+
+| Caller | Pattern | Example |
+|--------|---------|---------|
+| CaptureTriageService | `SHA256(captureItemId:sequence:normalizedTitle)` | Deterministic — same input always produces the same key |
+| InboxTriageAssistant | `"inbox-triage:{itemId:N}:{boardId:N}"` | Per-board, per-item scoped |
+| InboxTriageDigestAgent | `"digest:{runId:N}:{itemId:N}"` | Per-digest-run, per-item scoped |
+| MCP write tools / chat / tests | `Guid.NewGuid().ToString()` | One-shot, no replay intent |
+
+No single format is mandated. Callers that need replay idempotency (capture triage) use deterministic keys; callers that do not use random GUIDs.
+
+### Uniqueness scope
+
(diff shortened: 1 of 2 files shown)


Chris0Jeky:

### Review A findings (2 factual inaccuracies — both fixed)
- **A-1 (MEDIUM)**: Context said max-length enforced at domain level; actually enforced at infrastructure level via HasMaxLength — **fixed** in `26316302`
- **A-2 (MEDIUM)**: Alternatives section said "board/item context" in SHA256 hash; actually a per-capture-item GUID — **fixed** in `26316302`

### Review B findings
**0 findings.** Every factual claim cross-referenced against 13 source files. All key patterns, uniqueness scope, 409 behavior, execution idempotency, and header requirement verified as accurate.

### Verdict
All findings addressed. ADR is factually accurate, follows the template, and INDEX.md entry is correct.

PR #1186, opened 2026-06-06
Harden FirstRunBootstrapper key-file write: 0600 perms + guarded Mutex ctor

Chris0Jeky:
## Summary

Closes #1178

- **Guarded Mutex constructor**: The `new Mutex()` call and `WaitOne` in `PersistValue` are now wrapped in a try/catch for `UnauthorizedAccessException`, `WaitHandleCannotBeOpenedException`, and `IOException` -- matching the CLI's `CliFirstRunBootstrapper` pattern. On locked-down enterprise hosts, the API degrades gracefully to in-memory-only values instead of crashing at startup.
- **Unix file permissions (0600)**: After `File.WriteAllText` and before the atomic `File.Move`, the temp file mode is set to `UserRead | UserWrite` on non-Windows systems so the key file is never world-readable on POSIX.
- **`BuildMutexName` promoted to `internal`** for direct test access.
- **5 new tests**: `BuildMutexName` determinism/uniqueness/prefix, plus structural checks that the mutex guard and Unix permissions patterns are present in the source.

## Test plan

- [x] `dotnet build backend/Taskdeck.sln -c Release` -- 0 errors, 0 new warnings
- [x] `dotnet test` with `FullyQualifiedName~FirstRunBootstrapperTests` -- 16/16 pass (11 existing + 5 new)
- [ ] CI green on this PR

Issue #1140, opened 2026-05-30
Workspace hygiene & one-command dev-up: clean script, absolute SQLite path, .dockerignore


Chris0Jeky:
## Context
The working tree and onboarding flow trip up agents. (Note: the stray dirs below are all gitignored/untracked — local cruft, not committed — so the actionable parts are the SQLite root-drop, .dockerignore, and a dev-up script, not deleting tracked files.)

## Evidence
- Gitignored-but-on-disk clutter floods grep/find: `Microsoft/Windows/`, empty root `node_modules/`, root `taskdeck.db` (671KB, from `appsettings.json:128` relative `Data Source=taskdeck.db`), `.runtime-codex` 1.5GB, `artifacts/` 299MB, `.tmp` logs. `.dockerignore` excludes none of `Microsoft/`/`artifacts/`.
- No root bootstrap/seed orchestrator — README Quick Start is two manual terminals with no seed; seed only in `START_HERE.md:128-134` (requires backend already up); `demo:director` only spawns Playwright.

## Acceptance criteria
- [ ] Make the default SQLite path absolute under a data dir via env so `dotnet run` from root doesn't drop a DB in the repo root.
- [ ] `scripts/clean-workspace.{ps1,sh}` removing generated DBs/logs/stray dirs.
- [x] Add `Microsoft/`, `artifacts/`, `.analysis-logs/`, `.tmp/` to `.dockerignore`. (PR `#2624`, merge `20699b327`, 2026-09-05)
- [ ] `scripts/dev-up.{ps1,sh}` (or Taskfile) starts backend+frontend, waits for readiness, runs `demo:seed`; referenced from README + AGENTS.md.

_Source: deep audit workspace-hygiene-and-clean (MEDIUM); coordinator confirmed the clutter is gitignored/untracked._

Chris0Jeky:
## 2026-08-20 source quick-start regression

The documented one-command path now has a reproduced false-green failure on current main: an existing node_modules tree lacked newly declared vue-i18n, dev-up skipped dependency reconciliation, and the Windows PID file recorded Notepad as the frontend launcher while reporting "Stack is up".

The focused regression and acceptance criteria are tracked in #1875. This keeps #1140's remaining workspace-cleanup scope intact while giving the shipped source-start blocker a Priority I execution seam.

Chris0Jeky:
[...]

PR #1204, opened 2026-06-12
chore: gitignore SQLite migrate-lock sidecar files

Chris0Jeky:
## Summary
- Adds `*.migrate.lock` to `.gitignore` and `**/*.migrate.lock` to `.dockerignore`.

## Context
`SerializedMigrator` (shipped via #1186, closes the #1164 follow-up) creates a persistent 0-byte advisory-lock sidecar `<db>.migrate.lock` next to the SQLite database on every startup of the API/MCP/CLI, and intentionally never deletes it (only the FileStream is disposed). The existing `*.db` / `*.db-shm` / `*.db-wal` ignore patterns don't match the `.migrate.lock` suffix, so any local run permanently dirties `git status` (one is sitting untracked at `backend/src/Taskdeck.Api/taskdeck.db.migrate.lock` right now) and the file would leak into docker build contexts.

## Verification
- `git check-ignore backend/src/Taskdeck.Api/taskdeck.db.migrate.lock` now matches (verified locally on the branch).
- No code/build impact — ignore-file-only change.

Part of the workspace-hygiene track (#1140).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -20,6 +20,7 @@ backend/**/*.db-wal
 taskdeck.db
 taskdeck.db-shm
 taskdeck.db-wal
+**/*.migrate.lock
 api-tests.log
 
 docs/archive/
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ backend/.vs/
 *.db
 *.db-shm
 *.db-wal
+*.migrate.lock
 
 # .NET Build Results
 [Dd]ebug/


Chris0Jeky:
## Adversarial Code Review

### CRITICAL
- None

### HIGH
- None

### MEDIUM
- None

### LOW
- None

### Bot Comments Addressed
- None present at review time (will re-check before merge per policy).

### Summary
0 findings. Two-line ignore-pattern change, verified:
- `git check-ignore -v` matches both `backend/src/Taskdeck.Api/taskdeck.db.migrate.lock` (the sidecar currently dirtying `git status`) and a root-level `taskdeck.db.migrate.lock` (`.gitignore:14`).
- `git ls-files | grep -i 'migrate.lock'` → no tracked files are affected by the new patterns.
- `.dockerignore` uses `**/*.migrate.lock` (Docker's `**` matches zero or more directories, so root-level and nested sidecars are both excluded from build contexts), consistent with the existing `backend/**/*.db` style.
- Pattern matches `SerializedMigrator.LockFileSuffix` (`.migrate.lock`, SerializedMigrator.cs:46) — the only producer of these files.

Not merge-blocking.

Issue #1195, opened 2026-06-06
LlmQueueToProposalWorker fetches unbounded rows before applying batch cap


Chris0Jeky:
## Summary
The worker calls `GetByStatusAsync(Pending)` and `GetByStatusAsync(Processing)` without any limit, materializing all pending/processing items into memory. Under backlog accumulation, hundreds of items could be loaded before `BuildFairBatchItems` applies a cap.

## Location
`backend/src/Taskdeck.Api/Workers/LlmQueueToProposalWorker.cs`, lines 69-76

## Suggested Fix
Pass `MaxBatchSize * 2` as a limit to `GetByStatusAsync` so the DB query itself is bounded.

## Source
Found during codebase health analysis (2026-06-06).

Chris0Jeky:
## Caller audit — ⚠️ a blanket `limit = 100` default would introduce GDPR bugs

While scoping this, I audited every caller of the three unbounded `LlmQueueRepository` reads (`GetByStatusAsync`, `GetByUserAsync`, `GetByUserAndStatusAsync`). They have **mixed semantics** — some callers *require all rows*, so adding a `limit = 100` default (as first suggested) would silently truncate them:

**MUST stay unbounded (correctness/compliance):**
- `AccountDeletionService.cs:108` — `GetByUserAsync` to **delete all** user captures. Capping at 100 leaves >100 captures undeleted → **GDPR account-deletion bug**.
- `DataExportService.cs:46,261` — `GetByUserAsync` for **full data export**. Capping → incomplete export → **GDPR data-portability bug**.

**Display/UI reads — should be bounded (the actual target of this issue):**
- `CaptureService.cs:101` — list view (this is also #1193's slice).
- `HealthController.cs:83` — pending count for `/health` (better served by a `COUNT` query than fetch-then-count; capping makes the count misleading).

**Workers / agents — bounding changes throughput semantics, needs a deliberate per-call limit:**
- `LlmQueueToProposalWorker.cs:69,73` (drains pending/processing), `InboxTriageAssistant.cs:80`, `InboxTriageDigestAgent.cs:82`, `LlmQueueService.cs:82,88,118`, `OpsCliService.cs:195`.

### Recommended design (not a blanket default)
[...]

PR #1250, opened 2026-06-27
fix(queue): bound the two remaining unbounded LlmQueue Pending reads (#1237)

### LINKED ISSUE #1237: Bound the remaining unbounded LlmQueue Pending reads (ProcessNextRequestAsync, OpsCliService)
## Summary
Surfaced by the adversarial review of PR #1236 (#1195). Two non-worker call sites still materialize the full LlmQueue backlog with the same unbounded-then-filter pattern #1195 fixed for the worker:

1. **`LlmQueueService.ProcessNextRequestAsync`** (`LlmQueueService.cs:118`) — calls the unbounded `GetByStatusAsync(RequestStatus.Pending)`, then filters `!IsCaptureRequestType`, sorts ascending, and iterates to claim the first claimable item. Under backlog it loads every pending row to claim one.
2. **`OpsCliService` queue-pending listing** (`OpsCliService.cs:195`) — calls the unbounded `GetByStatusAsync(Pending)` then `.Take(50)` in memory.

## Why out of scope for #1236
#1195/#1236 is scoped to the `LlmQueueToProposalWorker`. These are the same *class* of unbounded read but different call sites; folding them in would have widened the PR.

## Suggested fix
PR #1236 added the bounded, type-aware primitives `ILlmQueueRepository.GetOldestPendingNonCaptureAsync(int limit, ...)` / `GetOldestProcessingCaptureAsync(int limit, ...)`. Reuse them:
- `ProcessNextRequestAsync` → `GetOldestPendingNonCaptureAsync(limit)` (it only needs the oldest claimable non-capture item; a small limit with a re-poll loop, or a dedicated "claim next" method, avoids materializing the backlog). Note the in-query type filter also removes the capture-rows-in-Pending starvation noted in the #1236 review.
- `OpsCliService` → pass an explicit bounded limit (e.g. 50) instead of `.Take(50)` after an unbounded read.

## Source
Adversarial review of PR #1236 — reviewer A LOW + reviewer B LOW (independent), 2026-06-20.

Chris0Jeky:
Closes #1237.

Surfaced by the #1236 (#1195) adversarial review: two non-worker call sites still materialized the full LlmQueue Pending backlog with the same unbounded-then-filter pattern #1195 fixed for the worker.

## Changes
- **`LlmQueueService.ProcessNextRequestAsync`** — claims from the bounded, type-aware `GetOldestPendingNonCaptureAsync(ClaimScanLimit=100)` (reuses the #1236 primitive) instead of loading the entire Pending backlog and filtering `!IsCaptureRequestType` + sorting in memory. It only needs the oldest claimable non-capture item; 100 leaves ample headroom for lost optimistic-claim races before falling through to "no pending requests" (the caller retries).
- **`OpsCliService` queue-pending listing** — uses a new bounded `GetByStatusForDisplayAsync(Pending, limit: 50)` instead of `GetByStatusAsync(...).Take(50)`.
- **New `ILlmQueueRepository.GetByStatusForDisplayAsync(status, limit)`** — mirrors `GetByStatusAsync` (newest-first) but pushes a SQL `LIMIT`; **display/diagnostics only**. The SQLite raw-SQL + LINQ paths are refactored into a shared `GetByStatusCoreAsync(status, int? limit)`. Documented that claim/process paths must use the type-aware `GetOldest*` primitives so a bound never starves non-capture work (#1195). `GetByStatusAsync` (unbounded) is unchanged — the health backlog gauge still needs the full depth.

## Tests
- New repository integration test: SQL `LIMIT` pushdown (command interceptor), newest-first, and that the display read **includes** capture types (unlike the work-drain reads) + `limit < 1` guard.
- `ProcessNextRequestAsync` service tests restructured: capture-skipping / oldest-first now live in the query, so the tests mock the non-capture primitive and assert the unbounded `GetByStatusAsync` is **never** called.
- The new method threaded through the 3 `ILlmQueueRepository` test fakes.

## Verification
- `dotnet build backend/Taskdeck.sln -c Release` → 0 errors.
- Targeted suite (`~LlmQueue|~OpsCli|~WorkerResilience`): 23 Application + 88 Api = **0 failures**.

The OpsCli queue-pending command was previously untested (thin formatting wrapper); the bounded read it now calls is covered by the new repository test.

### DIFF
--- a/backend/src/Taskdeck.Application/Interfaces/ILlmQueueRepository.cs
+++ b/backend/src/Taskdeck.Application/Interfaces/ILlmQueueRepository.cs
@@ -24,10 +24,11 @@ public interface ILlmQueueRepository : IRepository<LlmRequest>
     Task<IEnumerable<LlmRequest>> GetCapturesByUserAsync(Guid userId, int limit, int offset, CancellationToken cancellationToken = default);
 
     /// <summary>
-    /// Returns all requests with the given status, newest-first, unbounded. Callers that genuinely need
-    /// the full set rely on this (the health backlog gauge measures total depth), so do NOT add a default
-    /// cap here. Bounded background work-drains must use the type-aware <c>GetOldest*</c> methods below;
-    /// a bounded DISPLAY listing must use <see cref="GetByStatusForDisplayAsync"/>.
+    /// Returns all requests with the given status, newest-first, unbounded. The health backlog gauge
+    /// currently inspects every row through this (it should move to count primitives — tracked in #1251),
+    /// so do NOT add a default cap here. Bounded background work-drains must use the type-aware
+    /// <c>GetOldest*</c> methods below; a bounded DISPLAY listing must use
+    /// <see cref="GetByStatusForDisplayAsync"/>.
     /// </summary>
     Task<IEnumerable<LlmRequest>> GetByStatusAsync(RequestStatus status, CancellationToken cancellationToken = default);
 
--- a/backend/src/Taskdeck.Application/Interfaces/ILlmQueueRepository.cs
+++ b/backend/src/Taskdeck.Application/Interfaces/ILlmQueueRepository.cs
@@ -24,12 +24,20 @@ public interface ILlmQueueRepository : IRepository<LlmRequest>
     Task<IEnumerable<LlmRequest>> GetCapturesByUserAsync(Guid userId, int limit, int offset, CancellationToken cancellationToken = default);
 
     /// <summary>
-    /// Returns all requests with the given status, newest-first, unbounded. Callers that need the
-    /// full set rely on this (the health backlog gauge and the ops queue listing count/inspect every
-    /// row), so do NOT add a default cap here. Bounded background work-drains must use the type-aware
-    /// <c>GetOldest*</c> methods below, which bound the read at the database.
+    /// Returns all requests with the given status, newest-first, unbounded. Callers that genuinely need
+    /// the full set rely on this (the health backlog gauge measures total depth), so do NOT add a default
+    /// cap here. Bounded background work-drains must use the type-aware <c>GetOldest*</c> methods below;
+    /// a bounded DISPLAY listing must use <see cref="GetByStatusForDisplayAsync"/>.
     /// </summary>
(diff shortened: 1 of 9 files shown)


Chris0Jeky:
## Adversarial review (2 independent lenses: correctness + regression) — addressed

No CRITICAL/HIGH/MEDIUM. Both reviewers confirmed: the refactor preserves `GetByStatusAsync` byte-for-byte at `limit:null`, raw SQL is parameterized (no injection), FIFO ordering is preserved via the oldest-first primitive, and the test restructure loses no coverage (capture-skipping moved into the query, which #1236's repo test already covers). Four LOW findings, all handled:

- **LOW — SQLite display read didn't re-sort.** `FromSqlInterpolated + Include` wraps the raw SQL in a subquery whose outer `ORDER BY` isn't guaranteed to survive, so the documented newest-first wasn't guaranteed within the page. **Fixed** (`05182bed`'s parent): `GetByStatusForDisplayAsync` now re-sorts newest-first in memory (same workaround the `GetOldest*` primitives use).
- **LOW — health gauge materializes all rows to count.** Out of scope for #1237 (the two reads it bounds). **Tracked in #1251** + softened the `GetByStatusAsync` doc comment to say the gauge "currently inspects every row" rather than "genuinely needs the full set."
- **LOW — OpsCli `queue.pending` had no service-level test** (pre-existing gap). **Added** an `OpsCliServiceTests` case asserting it uses the bounded display read (cap 50) and never the unbounded `GetByStatusAsync`, locking the limit constant against drift.
- **LOW — `ClaimScanLimit=100` behavior note (transparency):** `ProcessNextRequestAsync` now scans only the oldest 100 non-capture candidates, so under sustained heavy claim contention (>100 concurrent in-flight claim losses — far beyond realistic single-instance concurrency) it can return `NotFound` while a claimable item exists past position 100. This is self-healing (the only caller is the manual pull endpoint, which simply re-invokes) and documented inline. No code change recommended.

**Verification:** `dotnet build -c Release` → 0 errors; targeted suite (LlmQueue repo/service + OpsCli) = 24 Application + 31 Api = 0 failures, incl. the new display re-sort path and the OpsCli cap test.

PR #1253, opened 2026-06-27
perf(health): readiness gauge uses count primitives instead of materializing Pending rows (#1251)

### LINKED ISSUE #1251: Health-gauge queue depth materializes all pending rows instead of using count primitives
**Severity:** LOW (surfaced by the #1250 / #1237 review). Out of scope for #1237 (which bounded the two *worker/CLI* reads).

## Problem

`HealthController` (`HealthController.cs:83-86`) calls the unbounded `ILlmQueueRepository.GetByStatusAsync(RequestStatus.Pending)` and then `.ToList()`/groups it only to compute three integers — `queueDepth` (non-capture count), `captureDepth`, and `totalDepth`. On a frequently-polled readiness/liveness probe with a large backlog this materializes every pending row just to count them.

The interface already exposes `CountPendingNonCaptureAsync` (returns `queueDepth` without materializing rows). This is arguably a larger blast radius than the two reads #1237 bounded, but it is a distinct call site.

## Proposed fix

Replace the gauge's row materialization with count primitives:
- `CountPendingNonCaptureAsync()` for `queueDepth`;
- add `CountPendingAsync()` (or `CountPendingCaptureAsync()`) to derive `captureDepth` / `totalDepth`.

Keep `GetByStatusAsync` unbounded for any genuine full-set caller, but stop the gauge from using it.

## References
- `backend/src/Taskdeck.Api/Controllers/HealthController.cs:83`
- `ILlmQueueRepository.CountPendingNonCaptureAsync` (added in #1236)
- Found in the PR #1250 review (#1237).

Chris0Jeky:
## #1251 — readiness gauge uses count primitives

`HealthController`'s queue check called the unbounded `GetByStatusAsync(RequestStatus.Pending)` and `.ToList()`/grouped it only to compute three integers (`depth`, `captureDepth`, `totalDepth`). On a frequently-polled readiness/liveness probe with a large backlog, that materialized every Pending row just to count them.

### Change
- `queueDepth` ← `CountPendingNonCaptureAsync()` (already existed; the automation backlog the threshold gates on)
- `captureDepth` ← new `CountPendingCaptureAsync()` (Pending capture-triage, in-query count)
- `totalDepth` ← `queueDepth + captureDepth`
- `GetByStatusAsync` stays unbounded for genuine full-set callers; only the gauge stops using it.

### Tests
- `LlmQueueRepositoryIntegrationTests.BacklogCounts_*` now also asserts `CountPendingCaptureAsync() == 2` (Pending captures only — not the 3 Processing captures, not the 4 Pending non-captures).
- `HealthApiTests` already seeds Pending capture rows and asserts the gauge depths → exercises the rewired controller end-to-end.
- The three hand-written `ILlmQueueRepository` fakes (worker tests) implement the new method.

Verified: full sln builds clean (0 errors); targeted suite (count integration + HealthApi gauge + worker fakes) 70/70.

Closes #1251. Completes the query-bounds quality line started in #1195/#1236/#1237.

### DIFF
--- a/backend/src/Taskdeck.Api/Controllers/HealthController.cs
+++ b/backend/src/Taskdeck.Api/Controllers/HealthController.cs
@@ -80,10 +80,12 @@ public class HealthController : ControllerBase
         {
             using var scope = _serviceProvider.CreateScope();
             var unitOfWork = scope.ServiceProvider.GetRequiredService<IUnitOfWork>();
-            var pending = await unitOfWork.LlmQueue.GetByStatusAsync(RequestStatus.Pending, ct);
-            var pendingList = pending.ToList();
-            var queueDepth = pendingList.Count(item => !CaptureRequestContract.IsCaptureRequestType(item.RequestType));
-            var captureDepth = pendingList.Count - queueDepth;
+            // Count primitives instead of materializing every Pending row on this frequently-polled
+            // readiness probe (#1251): queueDepth = Pending non-capture (the automation backlog the
+            // threshold gates on), captureDepth = Pending capture-triage, totalDepth = their sum.
+            var queueDepth = await unitOfWork.LlmQueue.CountPendingNonCaptureAsync(ct);
+            var captureDepth = await unitOfWork.LlmQueue.CountPendingCaptureAsync(ct);
+            var totalDepth = queueDepth + captureDepth;
             var queueThreshold = Math.Max(_workerSettings.MaxBatchSize * 20, 100);
             var queueHealthy = queueDepth <= queueThreshold;
 
@@ -91,7 +93,7 @@ public class HealthController : ControllerBase
             {
                 status = queueHealthy ? "Healthy" : "Degraded",
                 depth = queueDepth,
-                totalDepth = pendingList.Count,
+                totalDepth,
                 captureDepth,
                 threshold = queueThreshold
             };
--- a/backend/src/Taskdeck.Application/Interfaces/ILlmQueueRepository.cs
+++ b/backend/src/Taskdeck.Application/Interfaces/ILlmQueueRepository.cs
@@ -62,6 +62,9 @@ public interface ILlmQueueRepository : IRepository<LlmRequest>
 
     /// <summary>Counts Processing capture-triage requests for backlog telemetry, without materializing rows.</summary>
     Task<int> CountProcessingCaptureAsync(CancellationToken cancellationToken = default);
+
+    /// <summary>Counts Pending capture-triage requests for the readiness gauge, without materializing rows.</summary>
+    Task<int> CountPendingCaptureAsync(CancellationToken cancellationToken = default);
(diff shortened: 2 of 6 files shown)


Chris0Jeky:
### Self-review (correctness + drift)

**Correctness — classification equivalence (the key risk):** the old gauge classified capture via `CaptureRequestContract.IsCaptureRequestType` (`StartsWith("inbox.capture.", OrdinalIgnoreCase)`); the count queries use the repo's `CaptureRequestTypeLike` = `LIKE 'inbox.capture.%'`. These agree: SQLite `LIKE` is ASCII-case-insensitive (matches the Ordinal**IgnoreCase** prefix), `.` is a literal in both, and `%` ≡ "starts-with". `RequestType` is a required non-null column, so `LIKE` ∪ `NOT LIKE` is a **complete, non-overlapping partition** of the Pending set → `queueDepth (NOT LIKE) + captureDepth (LIKE) == total Pending`, exactly the old `pendingList.Count`. The `BacklogCounts_*` integration test (Pending captures=2 vs Pending non-captures=4 vs Processing captures=3) and the `HealthApiTests` gauge test (depth==0, captureDepth≥3) both confirm the partition.

**Drift / completeness:**
- The telemetry `AutomationQueueBacklog.Record(queueDepth, …)` is unchanged (same value, count-derived).
- `GetByStatusAsync(Pending)` is retained for genuine full-set callers; only the gauge stops materializing.
- No now-unused usings flagged (build clean, 0 new warnings).
- All three hand-written `ILlmQueueRepository` test fakes implement the new method consistently with their `CountProcessingCaptureAsync`.

**Verification:** full sln builds clean (0 errors); count integration + HealthApi gauge + worker-fake suites 70/70.

Chris0Jeky:
### Independent adversarial review — no defects found

A second, independent adversarial pass (skeptic prompted to refute the change) confirms the refactor is behavior-preserving. It went a level deeper than the self-review on the partition equivalence:

- **NULL `RequestType` edge (the only theoretical divergence):** in SQL three-valued logic, `NOT (NULL LIKE 'inbox.capture.%')` is NULL, so a NULL-`RequestType` Pending row would be counted in *neither* `queueDepth` nor `captureDepth` (unlike the old in-memory path, which counted it as non-capture). **This cannot occur** — `RequestType` is `IsRequired()` (`LlmRequestConfiguration.cs:20`) with domain default `string.Empty` (`LlmRequest.cs:15`). So `LIKE`/`NOT LIKE` is a complete, disjoint partition.
- **No internal inconsistency from two COUNTs:** `totalDepth` is *derived* (`queueDepth + captureDepth`), not a third query, so the sum-identity always holds even if a row changes status between the two counts — benign for a racy health snapshot.
- Case-folding (SQLite ASCII-case-insensitive `LIKE` ≡ `OrdinalIgnoreCase` on the all-ASCII prefix), literal `.`, telemetry `Record(queueDepth)`, response shape, and the `queueHealthy`/threshold logic are all unchanged.

Combined with the self-review above, this PR has had two independent adversarial reviews, both clean. CI green (20/20).

Issue #1189, opened 2026-06-06
Thread.Sleep inside lock in RedisCacheService.GetConnection starves thread pool


Chris0Jeky:
## Summary
`RedisCacheService.GetConnection()` acquires `lock (_connectionLock)` and then calls `Thread.Sleep(delay)` with delays of 500ms and 1000ms during the retry loop. This blocks an ASP.NET Core thread-pool thread for up to 1.5 seconds while holding the lock, starving concurrent requests during Redis failures.

## Location
`backend/src/Taskdeck.Infrastructure/Services/RedisCacheService.cs`, lines 224-282

## Impact
Under load with Redis outage, all concurrent callers block on the lock for the full sleep duration, cascading into thread-pool starvation.

## Suggested Fix
Replace the retry loop inside the lock with a fire-and-forget reconnect outside the lock, or record the failure and return `null` immediately, only attempting reconnect after `ReconnectMinInterval` has passed (which the outer throttle already provides — the inner retry loop is redundant).

## Source
Found during codebase health analysis (2026-06-06).

PR #1392, opened 2026-07-17
Make Redis dispose-vs-connect guard reach the seam deterministically (#1332)

### LINKED ISSUE #1332: Test reliability: Redis dispose concurrency guard misses connect seam under full-suite load
## Symptom

During required full-backend verification for #1298 on local Windows, the API suite failed with:

```
RedisCacheServiceTests.Dispose_IsNotSerialized_BehindAnInFlightConnect
Expected connectEntered.Wait(TimeSpan.FromSeconds(5)) to be True ... but found False.
```

Command:

```
dotnet test backend/Taskdeck.sln -c Release -m:1 --no-restore
```

Result: API tests 1,845 passed / 1 failed; every other backend project passed (apart from the existing INV-09 architecture skip). The #1298 diff does not touch Redis, threading, test scheduling, or cache behavior.

The exact failing test subsequently passed 5/5 sequential narrow runs:

```
dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release --no-restore --filter FullyQualifiedName~Taskdeck.Api.Tests.RedisCacheServiceTests.Dispose_IsNotSerialized_BehindAnInFlightConnect
```

## Investigation

The assertion fails before `Dispose()` is timed: a `Task.Run` worker did not reach `OnBeforeConnect` within the fixed five-second wait under full-suite load. Narrow runs reach the seam and pass the actual lock-serialization invariant. This distinguishes the observed failure from a deterministic RedisCacheService regression but does not yet prove the scheduler/thread-pool cause.

This is separate from the SQLite integration-factory defect tracked in #1282.

## Recommended fix

Make the regression guard's connecting actor start deterministically (for example, a named dedicated background thread plus explicit coordination) while preserving the real invariant: `Dispose()` must acquire `_connectionLock` promptly while another actor is parked immediately before the blocking connect. Do not merely increase the timeout or remove the concurrency assertion.

## Acceptance criteria

- [ ] The connecting actor reaches the seam deterministically under full API-suite load on Windows.
- [ ] The test still fails against the pre-#1189 lock-around-connect implementation.
- [ ] Repeated narrow runs and repeated required full-backend runs are green.
- [ ] No production Redis behavior is weakened to make the test pass.

Chris0Jeky:
## Summary

`RedisCacheServiceTests.Dispose_IsNotSerialized_BehindAnInFlightConnect` flaked under full-API-suite load on Windows with `Expected connectEntered.Wait(TimeSpan.FromSeconds(5)) to be True ... but found False`. The connecting actor was launched with `Task.Run(...)`, which queues a thread-pool work item. Under a saturated pool the injection throttle delayed running that item, so the worker never reached `OnBeforeConnect` before the wait window elapsed — a scheduling artifact, not a real regression.

This is a **test-only** change. No production `RedisCacheService` behavior is touched (the production file has an empty diff).

## Mechanism (deterministic, not a bigger timeout)

Replaced `Task.Run(...)` with a **dedicated, named background `Thread`** (`redis-connector`, `IsBackground = true`) that is created and scheduled by the OS immediately, independent of thread-pool saturation. Coordination is unchanged and explicit:

- `GetConnection` invokes `OnBeforeConnect` synchronously on the connecting thread before any `await`, so the existing `connectEntered` / `releaseConnect` handshake is a real rendezvous.
- The connecting thread is now `Join`ed (bounded, 5s) instead of awaited, and its result/exception are captured on fields so the post-dispose degrade assertions still hold.

The 5s wait was **not** increased; determinism comes from the dedicated OS thread, not from a wider margin. The `disposeElapsed < 500ms` concurrency assertion is preserved verbatim.

## Negative proof (still fails against pre-#1189 semantics)

Temporarily reverted the production `GetConnection` to hold `_connectionLock` across the blocking connect (the pre-#1189 lock-around-connect bug), rebuilt, and ran the test:

```
Expected disposeElapsed to be less than 500ms because Dispose must not be serialized
behind another thread's in-flight connect (#1189), but found 7s, 75ms and 86.9µs.
Failed!  - Failed: 1, Passed: 0
```

With the fixed production code the dedicated connecting thread releases `_connectionLock` before the connect, so `Dispose()` acquires it in microseconds. The revert was restored before committing (production diff is empty); the new mechanism did not weaken the guard.

## Verification

- `dotnet build backend/Taskdeck.sln -c Release` — clean (0 errors).
- Targeted repeat: 20/20 runs green of `--filter "FullyQualifiedName~RedisCacheServiceTests"` (8/8 tests each run).
- Negative proof: pre-#1189 revert → Dispose test fails (disposeElapsed ~7s), as required.
- Full project run: `Taskdeck.Api.Tests` — **2060 passed, 0 failed, 0 skipped**.

Closes #1332

### DIFF
--- a/backend/tests/Taskdeck.Api.Tests/RedisCacheServiceTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/RedisCacheServiceTests.cs
@@ -94,8 +94,8 @@ public sealed class RedisCacheServiceTests : IDisposable
         //   - Pre-#1189 code: lock held across connect -> probe times out -> lockWasFree == false.
         var probeAcquiredLock = false;
         var probeElapsed = TimeSpan.MaxValue;
-        var probeCompleted = new ManualResetEventSlim(false);
-        var seamFired = new ManualResetEventSlim(false);
+        using var probeCompleted = new ManualResetEventSlim(false);
+        using var seamFired = new ManualResetEventSlim(false);
 
         _cache.OnBeforeConnect = () =>
         {
@@ -156,8 +156,8 @@ public sealed class RedisCacheServiceTests : IDisposable
         //                    Dispose acquires it immediately.
         //   - Pre-#1189 code: the connecting thread holds _connectionLock across the multi-second
         //                    connect, so Dispose blocks behind it for the whole connect.
-        var releaseConnect = new ManualResetEventSlim(false);
-        var connectEntered = new ManualResetEventSlim(false);
+        using var releaseConnect = new ManualResetEventSlim(false);
+        using var connectEntered = new ManualResetEventSlim(false);
 
         _cache.OnBeforeConnect = () =>
         {
--- a/backend/tests/Taskdeck.Api.Tests/RedisCacheServiceTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/RedisCacheServiceTests.cs
@@ -168,7 +168,31 @@ public sealed class RedisCacheServiceTests : IDisposable
         };
 
         // Park a connecting thread inside the seam (simulated mid-connect).
-        var connectingCall = Task.Run(() => _cache.GetAsync<TestData>("connector").GetAwaiter().GetResult());
+        //
+        // Mechanism note (#1332): use a DEDICATED, named background Thread — NOT Task.Run. A
+        // Task.Run work item is queued to the thread pool, and under full-API-suite load the pool
+        // is saturated; its injection throttle then delays running the item, so the connecting
+        // worker could fail to reach OnBeforeConnect before the wait window elapsed (the observed
+        // flake: "connectEntered.Wait(5s) ... found False"). A dedicated Thread is created and
+        // scheduled by the OS immediately, independent of thread-pool saturation, so the connect
+        // seam is reached deterministically. This is a synchronization guarantee, not a bigger
+        // timeout — GetConnection invokes OnBeforeConnect synchronously on this thread before any
+        // await, so the handshake below is a real rendezvous.
+        TestData? connectorResult = null;
+        Exception? connectorError = null;
+        var connectingThread = new Thread(() =>
+        {
+            try
+            {
+                connectorResult = _cache.GetAsync<TestData>("connector").GetAwaiter().GetResult();
+            }
+            catch (Exception ex)
+            {
+                connectorError = ex;
+            }
+        })
+        { IsBackground = true, Name = "redis-connector" };
+        connectingThread.Start();
 
         connectEntered.Wait(TimeSpan.FromSeconds(5)).Should().BeTrue(
             "the connect seam must be reached so Dispose is timed against a real mid-connect window");
@@ -181,7 +205,9 @@ public sealed class RedisCacheServiceTests : IDisposable
 
         // Let the parked connecting thread proceed and finish (it must not throw post-dispose).
         releaseConnect.Set();
-        var connectorResult = connectingCall.GetAwaiter().GetResult();
+        connectingThread.Join(TimeSpan.FromSeconds(5)).Should().BeTrue(
+            "the connecting thread must complete once the simulated connect is released");
+        connectorError.Should().BeNull("the parked connector must degrade cleanly post-dispose, not throw");
         connectorResult.Should().BeNull("an unreachable Redis degrades the connector to a miss");
 
         // On the fixed path Dispose acquires the free lock in microseconds. On the old path it would


Chris0Jeky:
## Adversarial Code Review

Scope: single changed file `backend/tests/Taskdeck.Api.Tests/RedisCacheServiceTests.cs` (test-only). Production `RedisCacheService` diff is empty.

### CRITICAL
- None

### HIGH
- None

### MEDIUM
- None

### LOW
- None

### Analysis notes (verified correct, no action)
- **Memory visibility:** `connectorResult` / `connectorError` are written on the `redis-connector` thread and read on the test thread. `Thread.Join` establishes a happens-before edge, so the post-join reads are safe. On a `Join` timeout the `.Should().BeTrue()` assertion throws before any read, so stale values are never observed.
- **Determinism claim holds:** a dedicated `Thread` is OS-scheduled and bypasses the thread-pool injection throttle that was the actual root cause of the flake. `GetConnection` invokes `OnBeforeConnect` synchronously on that thread before any `await`, so the `connectEntered`/`releaseConnect` handshake is a genuine rendezvous — not a probabilistic sleep or a widened timeout (the 5s wait is unchanged).
- **No hang risk:** background thread + bounded `Join(5s)` + bounded `releaseConnect.Wait(5s)`.
- **Guard not weakened:** `disposeElapsed < 500ms` assertion preserved verbatim. Negative proof: reverting production to pre-#1189 lock-around-connect makes the test fail (disposeElapsed ~7s).

### Bot Comments Addressed
- None present at review time.

### Summary
0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW. Not merge-blocking. Test-only change; 20/20 targeted runs green, full Taskdeck.Api.Tests 2060/2060 pass, negative proof confirmed against pre-#1189 semantics.

Chris0Jeky:
## Adversarial Review — Round 2 (consolidated)

Two independent adversarial reviews were run against this PR with distinct lenses:

1. **Concurrency-correctness lens** — 0 substantive findings.
2. **Guard-strength / regression-sensitivity lens** — 0 substantive findings.

### Attack surface traced (all refuted against `RedisCacheService` source)

- **Seam synchronicity:** `GetConnection` invokes `OnBeforeConnect` synchronously on the connecting thread before any `await`, so the `connectEntered` / `releaseConnect` handshake is a genuine rendezvous — the dedicated thread cannot "miss" the seam.
- **Join happens-before:** `connectorResult` / `connectorError` are written on the `redis-connector` thread and read only after `Thread.Join` succeeds; `Join` establishes the happens-before edge, and the `Join(...).Should().BeTrue()` assertion throws before any read on the timeout path — stale values are never observed.
- **Revert sensitivity:** the test still fails hard against pre-#1189 lock-around-connect semantics (`disposeElapsed` ~7s vs. the unchanged 500ms bound) — the mechanism change did not weaken the regression guard.
- **Join(5s) flake surface:** post-`releaseConnect.Set()` the connecting thread only has to finish a failed connect against a refused loopback port (connectRetry=0, 3s cap) and the degrade path; the 5s bounded Join is not a new starvation-sensitive wait because the thread is already running (OS-scheduled, not pool-queued).
- **Cross-test bleed:** the seam (`OnBeforeConnect`) and cache instance are per-test-instance (xUnit creates a fresh fixture per test), so no state leaks between the two seam-using tests.

### Actionable items

- **INFO-1 (informational, folded into this PR):** the `ManualResetEventSlim` instances in the test file are never disposed — `releaseConnect` + `connectEntered` in `Dispose_IsNotSerialized_BehindAnInFlightConnect`, and the sibling events in `ConnectionLock_IsNotHeld_WhileBlockingConnectRuns`. Fix: `using var` declarations in BOTH tests for file consistency. No assertion, timing, or mechanism change.

### Summary

0 CRITICAL / 0 HIGH / 0 MEDIUM / 0 LOW substantive findings across both lenses. One informational disposal-hygiene item (INFO-1) to be fixed in this PR. Not merge-blocking.

Chris0Jeky:
## Adversarial Review — Fixes Applied (Round 2)

|---------|----------|------------|----------|
| INFO-1: `ManualResetEventSlim` instances never disposed (`releaseConnect` + `connectEntered` in `Dispose_IsNotSerialized_BehindAnInFlightConnect`; `probeCompleted` + `seamFired` in `ConnectionLock_IsNotHeld_WhileBlockingConnectRuns`) | INFO | `c9119919` | Build clean (0 errors); 5/5 consecutive targeted runs green (`--filter FullyQualifiedName~RedisCacheServiceTests`, 8/8 tests each run) |

Fix detail: all four events converted to `using var` declarations — disposal-only, no assertion, timing, or mechanism change. Disposal at method exit cannot race a live waiter: in the Dispose test both events are last touched before the bounded `Join` succeeds, and in the ConnectionLock test the probe thread sets `probeCompleted` before the seam returns.

All findings addressed. CI status: PENDING (fresh push, checks re-running).

Issue #1383, opened 2026-07-17
Flaky: AuditRetentionRepositoryIntegrationTests.DeleteOldEntriesAsync deleted=0 despite 70-day margin (windows API Integration)


Chris0Jeky:
## Symptom

`AuditRetentionRepositoryIntegrationTests.DeleteOldEntriesAsync_DeletesEntriesOlderThanCutoff` failed once on `main@ded2b020` (`API Integration (windows-latest)`, run 29548745315 attempt 1) with:

```
Expected deleted to be greater than or equal to 1, but found 0 (difference of -1).
```

1 failed / 2007 passed. The rerun of the identical commit (attempt 2) passed — classified flaky, not a regression. The triggering merge (#1377) touched only `SerializedMigratorTests` and cannot interact with audit retention.

## Why this is NOT a time-boundary tie

The test backdates the old entry 100 days via raw SQL and cuts off at -30 days — a 70-day margin. `deleted == 0` means either the raw-SQL `UPDATE` silently affected 0 rows, or `DeleteOldEntriesAsync` did not see/match the updated row.

## Candidate mechanisms (unproven)

1. **Unchecked UPDATE**: `ExecuteSqlRawAsync("UPDATE AuditLogs SET Timestamp = {0} WHERE Id = {1}", ...)` — the affected-row count is never asserted. If the `Id` parameter binding ever mismatches the stored format, the test premise breaks silently.
2. **String-format comparison**: the test hand-formats the timestamp (`yyyy-MM-dd HH:mm:ss.fffffff+00:00`) to match the repo's string-based `<` comparison; any environment-dependent formatting divergence between the seeded string and the cutoff string would flip the comparison.
3. **Test-host substrate change**: #1373 (merged 2026-07-17) gave the shared `TestWebApplicationFactory` production SQLite parity (per-connection WAL + busy_timeout + command timeout). A WAL cross-connection visibility subtlety between the raw `UPDATE` connection and the repository's delete connection is theoretically possible, though WAL should guarantee committed-write visibility to new snapshots.

## Suggested hardening (cheap)

- Assert `ExecuteSqlRawAsync` returned exactly 1 affected row.
- Optionally read the timestamp back through the repository connection before calling `DeleteOldEntriesAsync`, so a visibility/format failure is diagnosed at the right step.

## Evidence

- Failed: https://github.com/Chris0Jeky/Taskdeck/actions/runs/29548745315 (attempt 1)
[...]

PR #1373, opened 2026-07-17
Mirror production SQLite settings in the API test factory

Chris0Jeky:
Closes #1282

## Defect

The API integration test factory (`TestWebApplicationFactory`) replaced the production `TaskdeckDbContext` registration with a bare `UseSqlite($"Data Source={dbPath}")`. That drop lost three pieces of production parity:

- **Busy timeout** — the `SqlitePragmaConnectionInterceptor` (WAL journal mode + `PRAGMA busy_timeout`) was never attached, so the test host ran with `busy_timeout=0`. Under hosted concurrency the write lock returned `SQLITE_BUSY` immediately instead of waiting, surfacing as transient HTTP 500s.
- **Command timeout** — the resolved `DatabaseSettings.CommandTimeoutSeconds` was not applied to the SQLite command timeout, so long-running commands under contention did not get production's grace window.
- **Settings binding** — the factory did not bind the resolved `DatabaseSettings` at all, so any future drift in these values would silently diverge from production.

## Fix

Two layers, per the adjudicated adversarial reviews on this PR:

1. **Shared registration helper (structural drift prevention).** SQLite option-building is extracted into a single helper, `UseTaskdeckSqlite(connectionString, databaseSettings)` (`backend/src/Taskdeck.Infrastructure/Persistence/TaskdeckSqliteDbContextOptionsExtensions.cs`), which applies `CommandTimeout(CommandTimeoutSeconds)` and attaches `SqlitePragmaConnectionInterceptor(BusyTimeoutMilliseconds)`. Production `DependencyInjection.AddInfrastructure` and `TestWebApplicationFactory` BOTH call it — only the connection string differs (the factory keeps its isolated per-factory temp database). Semantics are byte-identical to the previous production registration; hand-copied drift (how #1282 happened) is now structurally impossible.
2. **Contract test (wiring guard).** `DatabaseRegistration_ShouldPreserveProductionSqliteSettings` asserts every load-bearing setting deterministically, so a factory registration that loses the interceptor or timeouts fails the build rather than resurfacing as flaky 500s.

The factory `Dispose` also now calls `SqliteConnection.ClearAllPools()` before deleting the temp database files, so pooled connections cannot hold `-wal`/`-shm` sidecar handles and silently leak temp files through the best-effort delete (review finding; mirrors `SqlitePragmaInterceptorTests.Dispose`).

## Red-before-green contract evidence

Against the **pre-fix** factory, the contract test fails on all three load-bearing assertions independently — each discriminates on its own:

[...]

PR #1451, opened 2026-07-23
Move AuditLogRepositoryIntegrationTests onto the workerless factory (#1438)

### LINKED ISSUE #1438: Flaky: AuditLogRepositoryIntegrationTests races the live AuditRetentionWorker (migrate to workerless factory, #1335 family)
## Symptom

In a full backend suite run (`dotnet test backend/Taskdeck.sln -c Release -m:1`) on 2026-07-18, `Taskdeck.Api.Tests.AuditLogRepositoryIntegrationTests.GetByUserAsync_ShouldOrderByTimestampDesc` failed once:

```
Expected firstIdx to be greater than or equal to 0 because first item should be in results, but found -1 (difference of -1).
  at AuditLogRepositoryIntegrationTests.cs:282
```

Determinism check: **6/6 isolated runs pass** (4× on the branch where the suite ran, `issue-1313/atomic-quota-reservation` @ 31235fca; 2× on `main` @ 4fdae133). The branch's diff touches only the LLM-quota seam (no audit-log code), and the identical suite was 7408/0/5 on the parent head — classic suite-interaction flake, not a regression.

## Root cause (same family as #1383, fixed for other classes by #1391)

`AuditLogRepositoryIntegrationTests` still uses the plain `TestWebApplicationFactory` (live hosted workers) and seeds `AuditLog` rows explicitly backdated to `2024-01-01` — ~2.5 years past the `AuditRetentionWorker` 90-day cutoff. When the worker's retention pass interleaves between the seed `SaveChangesAsync` and the `GetByUserAsync` read, the backdated rows are deleted mid-test and `FindIndex` returns -1. This is precisely the hosted-worker pre-emption family (#1335 boundary): #1391 migrated the `AuditRetention*` classes to `HostedWorkerDisabledTestWebApplicationFactory` (#1394) for this exact race, but this sibling class was not migrated.

## Fix

Adopt `HostedWorkerDisabledTestWebApplicationFactory` for `AuditLogRepositoryIntegrationTests` (mechanical, mirrors the #1394 adoptions). Any test in the class that genuinely needs live workers (none apparent) would need splitting, per the `ProcessNextClaimRaceTests` precedent.

## Evidence trail

- Full-suite failure: 7413 passed / 1 failed / 5 skipped on 31235fca (2026-07-18).
- 6/6 isolated passes (branch + main), 2026-07-18.
- Test source: `backend/tests/Taskdeck.Api.Tests/AuditLogRepositoryIntegrationTests.cs` (class fixture at top; failing test at ~line 256).

Chris0Jeky:
## Root cause

`AuditLogRepositoryIntegrationTests` used the plain `TestWebApplicationFactory`, which boots the app with all real hosted services — including `AuditRetentionWorker`. That worker runs a retention cleanup pass at host start with the default 90-day cutoff. Several tests in this class (notably `GetByUserAsync_ShouldOrderByTimestampDesc`) seed `AuditLog` rows backdated to `2024-01-01` — roughly 2.5 years past the cutoff. When the worker's retention pass interleaves between the test's seed `SaveChangesAsync` and its subsequent read, the backdated rows get deleted mid-test, and lookups by row `Id` return `-1` from `FindIndex`.

This is the same hosted-worker pre-emption family as #1383: a full-suite run on 2026-07-18 hit exactly this failure (7413 passed / 1 failed / 5 skipped on `31235fca`), while 6/6 isolated reruns passed cleanly — the signature of a suite-interaction race, not a real regression. Full details in #1438.

## Fix (mirrors the #1391 template)

#1391 already solved this exact race for the sibling `AuditRetention*` classes by migrating them onto `HostedWorkerDisabledTestWebApplicationFactory` (introduced in #1394), which boots the same app but strips app-assembly hosted services so no worker can pre-empt test-seeded state. That factory is also already in use by `LlmQueueRepositoryIntegrationTests`, `OutboundWebhookDeliveryRepositoryTests`, and `AutomationProposalRepositoryIntegrationTests`.

This PR mechanically applies the same migration to `AuditLogRepositoryIntegrationTests`:
- `IClassFixture<TestWebApplicationFactory>` -> `IClassFixture<HostedWorkerDisabledTestWebApplicationFactory>`
- constructor parameter and backing field type updated to match
- added a class-level doc comment (mirroring `AuditRetentionRepositoryIntegrationTests`'s) explaining why the workerless factory is required here

No test in this class exercises the retention worker itself (that behavior is already covered by `AuditRetentionRepositoryIntegrationTests`), so there was no conflicting requirement to keep any test on a live-worker host. No test assertions were changed.

## Verification

All commands run from the worktree, targeted only (no full-suite run, to avoid contending with another agent's full-suite slot):

- `dotnet build backend/Taskdeck.sln -c Release` -> **0 errors** (pre-existing warnings only, unrelated to this change).
- `AuditLogRepositoryIntegrationTests` run 6x consecutively:
  - Run 1: `Passed! - Failed: 0, Passed: 10, Skipped: 0, Total: 10`
  - Run 2: `Passed! - Failed: 0, Passed: 10, Skipped: 0, Total: 10`
  - Run 3: `Passed! - Failed: 0, Passed: 10, Skipped: 0, Total: 10`
  - Run 4: `Passed! - Failed: 0, Passed: 10, Skipped: 0, Total: 10`
  - Run 5: `Passed! - Failed: 0, Passed: 10, Skipped: 0, Total: 10`
  - Run 6: `Passed! - Failed: 0, Passed: 10, Skipped: 0, Total: 10`
- Cross-impact check, `AuditRetentionRepositoryIntegrationTests` run once: `Passed! - Failed: 0, Passed: 6, Skipped: 0, Total: 6` (confirms the sibling class the fix is modeled on is unaffected).

Not verified: full backend suite (`dotnet test backend/Taskdeck.sln -c Release -m:1`) — deliberately out of scope for this worker per the task's instructions, since another agent owns the box-wide full-suite slot concurrently. CI will run the full suite on this PR.

Closes #1438

### DIFF
--- a/backend/tests/Taskdeck.Api.Tests/AuditLogRepositoryIntegrationTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/AuditLogRepositoryIntegrationTests.cs
@@ -13,12 +13,21 @@ namespace Taskdeck.Api.Tests;
 /// Integration tests for AuditLogRepository against real SQLite.
 /// Covers cross-user isolation, unknown level handling, board-scoped queries,
 /// time-range filtering, and entity-type case-insensitivity.
+///
+/// Runs on <see cref="HostedWorkerDisabledTestWebApplicationFactory"/> (#1438, same family as
+/// #1391/#1383): several tests here (e.g. <c>GetByUserAsync_ShouldOrderByTimestampDesc</c>) seed
+/// <c>AuditLog</c> rows backdated well past the production <c>AuditRetentionWorker</c>'s 90-day
+/// cutoff. On the worker-enabled base factory, the worker's startup retention pass can delete
+/// those backdated rows between seed and assertion, intermittently failing lookups with a -1
+/// index. With the worker removed from this host, every seed/read here is deterministic. No test
+/// in this class exercises the retention worker itself, so there is no conflicting need to keep
+/// it live.
 /// </summary>
-public class AuditLogRepositoryIntegrationTests : IClassFixture<TestWebApplicationFactory>
+public class AuditLogRepositoryIntegrationTests : IClassFixture<HostedWorkerDisabledTestWebApplicationFactory>
 {
-    private readonly TestWebApplicationFactory _factory;
+    private readonly HostedWorkerDisabledTestWebApplicationFactory _factory;
 
-    public AuditLogRepositoryIntegrationTests(TestWebApplicationFactory factory)
+    public AuditLogRepositoryIntegrationTests(HostedWorkerDisabledTestWebApplicationFactory factory)
     {
         _factory = factory;
     }
--- a/backend/tests/Taskdeck.Api.Tests/AuditLogRepositoryIntegrationTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/AuditLogRepositoryIntegrationTests.cs
@@ -15,13 +15,14 @@ namespace Taskdeck.Api.Tests;
 /// time-range filtering, and entity-type case-insensitivity.
 ///
 /// Runs on <see cref="HostedWorkerDisabledTestWebApplicationFactory"/> (#1438, same family as
-/// #1391/#1383): several tests here (e.g. <c>GetByUserAsync_ShouldOrderByTimestampDesc</c>) seed
+/// #1391/#1383): one test — <c>GetByUserAsync_ShouldOrderByTimestampDesc</c> — seeds
 /// <c>AuditLog</c> rows backdated well past the production <c>AuditRetentionWorker</c>'s 90-day
 /// cutoff. On the worker-enabled base factory, the worker's startup retention pass can delete
 /// those backdated rows between seed and assertion, intermittently failing lookups with a -1
-/// index. With the worker removed from this host, every seed/read here is deterministic. No test
-/// in this class exercises the retention worker itself, so there is no conflicting need to keep
-/// it live.
+/// index. With the worker removed from this host, every seed/read here is deterministic; the
+/// class-level swap (rather than isolating just that one test) matches the granularity of the
+/// other #1391-family adoptions. No test in this class exercises the retention worker itself,
+/// so there is no conflicting need to keep it live.
 /// </summary>
 public class AuditLogRepositoryIntegrationTests : IClassFixture<HostedWorkerDisabledTestWebApplicationFactory>
 {


Chris0Jeky:
## Adversarial review — two independent read-only lenses (factory-swap correctness + coverage/regression blindspots)

**Findings (both fixed/tracked):**

- **LOW (both lenses converged):** the class doc-comment said "several tests here … seed AuditLog rows backdated well past the … 90-day cutoff" — exactly ONE test does (`GetByUserAsync_ShouldOrderByTimestampDesc`; the other nine seed current-time rows the cutoff can never touch). **Fixed in `61bd588d`** — the comment now names the single test and notes the class-level swap matches the #1391-family granularity. Class re-run post-edit: 10/10.
- **LOW (coverage lens, out-of-scope real gap → tracked as #1452):** after this swap, no test anywhere exercises `AuditRetentionWorker`'s real startup/scheduling path (`ExecuteAsync`) — unit tests call `CleanupOldEntriesAsync` against a fake repo, the repo integration tests run on the workerless host, and the guard test only asserts registration. Pre-existing gap, not a regression of this PR (a flake was never a detector), tracked rather than expanded into this PR per review policy.

**Survived refutation (evidence in the review passes):** the swap is character-for-character the proven peer wiring (fixture interface, field, ctor — zero residual plain-factory references), compared against all four #1391-family adopters; no test in the class depends on a live worker (all ten seed via DbContext and read via `IAuditLogRepository`; retention behavior lives in the separate workerless `AuditRetentionRepositoryIntegrationTests`); the workerless factory strips only app-assembly hosted services after `base.ConfigureWebHost`, so the #1373 `UseTaskdeckSqlite` WAL/busy-timeout parity is intact; this class was the LAST plain-factory consumer seeding backdated audit rows (repo-wide sweep — the remaining plain-factory audit touchers seed only current-time rows); the old flake shadowed no real product bug (deleting 2024 rows was correct behavior; the ordering assertions were already hardened against the vacuous −1-index pass); fixture semantics unchanged (one sealed factory instance per class, per-instance GUID-named temp DB, no collection coupling, xUnit parallelization unchanged).

**Determinism evidence** (from the PR body): 6× consecutive class runs green (10/10 each) + sibling `AuditRetentionRepositoryIntegrationTests` 6/6 unaffected, plus the post-fix 7th run.

Issue #1470, opened 2026-07-25
DATA_MODEL.md omits ApprovedRevisionId and DeferredUntil, and its snapshot figure is stale


Chris0Jeky:
`docs/architecture/DATA_MODEL.md` describes `AutomationProposal` without the two columns the proposal-revision subsystem added, and the snapshot figure its mirrors advertise for it is stale.

> **Corrected 2026-07-25:** the original body said the document "still describes itself as an '855 lines, 37 entities' snapshot". That is wrong — `DATA_MODEL.md` contains no such self-description (grep for `855`: 0 matches). The figure lives only in the three mirrors listed below. Corrected so nobody greps the wrong file. Caught by the PR #1471 review.

## What is wrong

- The `AutomationProposal` table description (`docs/architecture/DATA_MODEL.md:244-263`) omits **`ApprovedRevisionId`** (added by #1439, `backend/src/Taskdeck.Domain/Entities/AutomationProposal.cs:47`) and **`DeferredUntil`** (`:31`). Verified: `grep -cE "ApprovedRevisionId|DeferredUntil" docs/architecture/DATA_MODEL.md` → **0**.
- Three documents describe the file as an "855 lines, 37 entities, Mermaid ERD" artefact. It is **897 lines**. The mirrors are:
  - `docs/STATUS.md:333`
  - `docs/AUDIT.md:258`
  - `docs/IMPLEMENTATION_MASTERPLAN.md:1623`

  All three sit in dated historical delivery ledgers recording what PR #917 shipped on 2026-04-22, so they are not *false* as as-delivered records — but they are the only place the figure exists, and they are what a reader consults for the document's shape.

## Why it matters

`ApprovedRevisionId` is the column the whole approve-time pinning contract turns on (#1428/#1439/#1444/#1462) — the one an agent or contributor reading the data model to understand what Apply executes most needs to see. Its absence makes the canonical data-model doc actively misleading about the proposal lifecycle rather than merely incomplete.

## Scope

This is drift across the whole document, not one missing row, so it needs a recertification pass rather than a two-line patch:

1. Reconcile every entity block against the current EF model / migrations.
2. Re-derive the line/entity count, and decide where (if anywhere) it should live. A hand-maintained count in three historical ledgers is exactly the sort of claim that rots silently — consider dropping it rather than refreshing it, or moving it into the document itself where it can be checked against the file it describes.
3. If the figure is kept, update all three mirrors together.

## Provenance

[...]

Issue #1468, opened 2026-07-25
Frontend test files are excluded from typecheck, so type errors in specs ship silently


Chris0Jeky:
## What

`frontend/taskdeck-web/tsconfig.app.json` sets:

```json
"include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.vue"],
"exclude": ["src/tests/**", "src/stories/**"]
```

So `npm run typecheck` (`vue-tsc -b`) never type-checks anything under `src/tests/**`, and `vitest` does not type-check by default either (that needs `--typecheck`). Between them, **no gate type-checks the frontend test suite** — CI included.

## How it surfaced

While closing `#1462` (expose `approvedRevisionId` on the proposal type) I wrote a spec asserting the field and wanted to confirm the pairing actually guarded the declaration. Removing `approvedRevisionId` from the `Proposal` interface and re-running `npm run typecheck` **passed** — because the only file referencing it lives under `src/tests/`. TS types are erased at runtime, so the vitest run passed too.

That is a narrow instance of a general gap: a spec can reference a property that does not exist, pass a wrong-typed argument, or mis-type a mock, and nothing reports it. Type-level regressions are invisible in exactly the files whose job is catching regressions.

## Why it is not a one-line fix

Removing the exclude will almost certainly surface a backlog of pre-existing type errors across the spec suite (mocks typed loosely, partial fixtures cast with `as`, `vi.mocked` shapes). That is the work, and it needs a decision on how strict to be:

1. **Separate project** — add a `tsconfig.vitest.json` covering `src/tests/**` with relaxed options (e.g. no `noUnusedLocals`), referenced from the root tsconfig so `vue-tsc -b` builds it. Conventional for Vue/Vite projects; lets test strictness differ from app strictness.
2. **Fold into the app project** — drop the exclude, fix the fallout, one strictness bar everywhere.
3. **`vitest --typecheck`** — checks test files as part of the test run rather than the typecheck script. Also unlocks `expectTypeOf` for genuine type-level assertions, which is what would let a test pin a type declaration (the thing `#1462` could not do).

Option 1 plus option 3's `expectTypeOf` is probably the useful combination: the first stops silent type rot in specs, the second makes "this field is declared" testable at all.

## Acceptance criteria

- [ ] A recorded decision between the options above.
- [ ] `src/tests/**` is type-checked by a gate that runs in CI.
[...]

Issue #1465, opened 2026-07-25
Dismissing a rejected proposal drops its decision-time frozen revision back to the originals


Chris0Jeky:
## The behaviour

`GetEffectiveRevisionAsync`/`SelectEffectiveRevisionRef` in `AutomationProposalService` resolve a proposal's effective operation set from, in order: a pinned `ApprovedRevisionId`; the unconditional latest revision while `PendingReview`; the latest revision saved at or before `DecidedAt` while `Rejected` (the `#1439` decision-time freeze); otherwise the proposal's original operations.

A **rejected** proposal never pins a revision — `Reject` does not write `ApprovedRevisionId` (only `Approve` does). Its revised content is therefore reachable *only* through the `Rejected` status branch. So the moment such a proposal is **dismissed**, it leaves that branch with no pin to fall back on, and its effective set silently flips from the decision-time-frozen revision back to the **original** operations.

Concretely: reviewer edits a proposal (revision saved), rejects it (the frozen revision is what the review surfaces show), later files it away. The card now shows the operations of the original submission — a set the reviewer had already replaced before deciding.

## Why it is worth a decision

- It is **pre-existing** (`#1439` introduced the freeze; nothing here is a `#1444` regression), but `#1444` makes it *newly visible*: the review-queue list now returns effective operations, so the flip is observable on the queue card itself rather than only on a detail read.
- The freeze exists precisely so that "what was rejected" is stable and auditable. Dismissal is a filing action, not a re-decision, so losing the frozen content on dismissal arguably defeats the freeze's purpose for exactly the proposals a user tidied away.
- The symmetric case does **not** have this problem: a proposal dismissed after Approve/Apply/Fail keeps its pin, so it keeps showing what was executed. Only the rejected route degrades.

Current behaviour is now pinned by a regression test so it cannot change silently either way:
`AutomationProposalServiceTests.GetProposalsAsync_ShouldShowOriginals_ForProposalDismissedAfterRejection` (added in `#1444`), which documents the flip without endorsing it. Its sibling `GetProposalsAsync_ShouldReturnPinnedRevision_ForProposalDismissedAfterApply` pins the pinned route.

## Options

1. **Accept and document** — dismissal is terminal filing; the revision history is still queryable via `GET /{id}/revisions`, so nothing is lost, only the summary view changes. Cheapest; the test above already documents it.
[...]

Issue #1462, opened 2026-07-25
Frontend proposal types do not expose approvedRevisionId (#1439 residual)


Chris0Jeky:
## Context

`#1439` added approved-revision pinning: approve persists `AutomationProposal.ApprovedRevisionId` and Apply executes exactly that revision. The backend exposes it — `backend/src/Taskdeck.Application/DTOs/AutomationProposalDtos.cs:51` carries `public Guid? ApprovedRevisionId { get; init; }`, and `AutomationProposalService.cs:409` documents it as being on the DTO **"so clients can detect a pinned revision"**.

No frontend type declares it. `grep -rn "approvedRevisionId" frontend/taskdeck-web/src/` returns nothing, and neither proposal type carries the field:

- `frontend/taskdeck-web/src/types/automation.ts:36` — `interface Proposal`
- `frontend/taskdeck-web/src/types/queue.ts:32` — `interface AutomationProposal`

This was recorded as a known residual in `#1439`'s wave entry in `docs/STATUS.md` but never got its own tracking issue — it is distinct from `#1444` (list endpoint returns original operations) and `#1453` (reject-time pin is display-only), both of which are backend-side.

## Why it is worth closing

Additive and harmless today — the field is simply invisible to the client. But the backend contract was written specifically so a client *could* detect a pinned revision, and that capability currently has no consumer and no type-level existence. Any future review-surface work that wants to say "this proposal is pinned to a revision the approver saw" has to rediscover the field first, and a TypeScript type that silently omits a wire field is the kind of drift the review-first surface can least afford.

## Scope

Type/contract exposure only. **No behavior change, and explicitly no UI claim** — do not add a badge or copy asserting anything about pinning without a separate design decision; `#1298` (fake undo timeline) is the standing precedent that the review surface must not advertise semantics it cannot back.

## Acceptance criteria

- [ ] `approvedRevisionId?: string | null` added to the proposal type(s) that mirror the backend `ProposalDto` — confirm which of `types/automation.ts` / `types/queue.ts` actually map that payload before editing both.
- [ ] A unit test pins the field surviving deserialization/mapping, so a future mapper refactor cannot silently drop it again.
- [ ] `npm run typecheck` clean.
- [ ] No user-visible copy or affordance added in this slice.

## References

- `#1439` (approved-revision pinning), recorded residual in `docs/STATUS.md`
- Siblings, both backend-side: `#1444`, `#1453`

PR #1471, opened 2026-07-25
docs: re-hydrate canonical docs after #1469 (#1462) merged

Chris0Jeky:
## Summary

Post-merge re-hydration for **PR #1469** (`#1462`, merge commit `eb8c950f`). Docs-only.

Both canonical docs already recorded the `approvedRevisionId` exposure as delivered — that landed inside #1469 itself. This records what the **review round** changed about it, and closes two gaps the sweep turned up.

- **The pin is declared `required`, not optional.** That was a review outcome (an accepted Codex P2), and it is the load-bearing half of the contract: because the key is always on the wire — `MapToDto` always assigns it and the API serializes with default options — a field-by-field rebuild that drops the pin now **fails to compile** instead of silently shipping. The docs described the old optional shape.
- **`docs/STATUS.md` was missing `#1465`** from its residual list while `IMPLEMENTATION_MASTERPLAN.md` carried it. The two lists now agree.
- **`#1468` gains its measured scale** — 383 pre-existing errors with the `src/tests/**` exclusion lifted — instead of an unquantified "specs are never type-checked". That number came out of the #1469 review and belongs on the issue's record.
- **`#1470` added to both lists** — `docs/architecture/DATA_MODEL.md` omits `ApprovedRevisionId` and `DeferredUntil`, and its "855 lines, 37 entities" self-description is stale. Surfaced by the #1469 review, confirmed pre-existing, seeded rather than folded into that diff.

## Verification

- `node scripts/check-docs-governance.mjs` — passed
- `node scripts/check-golden-principles.mjs` — passed
- Diff is two prose lines across the two canonical docs; no code, config, workflow, or schema touched.

## Documentation

- [x] `docs/STATUS.md` updated — this PR *is* the shipped-reality sync
- [x] `docs/IMPLEMENTATION_MASTERPLAN.md` updated — residual list reconciled

## Tracking

No `Closes` keyword: this is a post-merge sync, and `#1462` is already closed by #1469. `#1470` is referenced, deliberately **not** closed — it tracks the DATA_MODEL recertification pass, which this PR does not perform.


Chris0Jeky:
## Adversarial review — 2 lenses, 7 findings, 5 refuted, 2 confirmed and fixed

Two read-only lenses (claim-truth; cross-doc consistency), each finding handed to a separate verifier prompted to refute it. On a docs-only PR the deliverable *is* the prose, so the bar was: a statement that is false, unverifiable, or self-contradictory.

Head at review `a81b612a`; fixes in `5f490ddf`. **Both confirmed findings are against claims I introduced one commit earlier.**

### CONFIRMED — MEDIUM: the 383 figure is wrong; the real count is 415

Both lenses landed on it and both verifiers reproduced it independently, then I re-ran it myself rather than take their word.

**383 was never the method the sentence described.** It came from a throwaway `tsconfig` that added `vitest/globals` to `types` and dropped `noUnusedLocals` / `noUnusedParameters` / `erasableSyntaxOnly`. The described method — lift only `src/tests/**` from `tsconfig.app.json`, run `npm run typecheck` (`vue-tsc -b --force`) — yields:

| | errors |
|---|---|
| baseline, untouched | **0** |
| `src/tests/**` lifted | **415** (64 files, **all** under `src/tests/`) |

`tsconfig.app.json` restored and verified byte-identical to HEAD (`git diff --exit-code`).

That the number was faithfully copied from #1469's review comment is **not** a defense. Sizing #1468 is the sentence's only job, and an unreproducible measurement is exactly what that job would trip over. Fixed in both docs, and the method is now stated precisely enough to re-run. A verifier also found the count is sensitive to `types` (415 as-is / 399 `+node` / 379 `+node,vitest/globals`) — recorded on **#1468**, because it means a "just flip the exclusion" patch does not exist; the spec type environment has to be decided first.

Corrections posted to **#1469** (the merged PR whose comment carries 383) and **#1468**. The before==after claim that finding rested on is unaffected and still reproduces — this PR added zero latent spec errors.

### CONFIRMED — LOW: `DATA_MODEL.md` does not describe itself

My STATUS line called the "855 lines, 37 entities" figure that file's *self-description*. It is not: `grep 855 docs/architecture/DATA_MODEL.md` → **0 matches**, and the file is **897 lines**. The figure lives only in mirrors — `STATUS.md:333`, `AUDIT.md:258`, `IMPLEMENTATION_MASTERPLAN.md:1623` — all inside dated historical ledgers recording what PR #917 shipped in April.

Left as written it would have sent whoever picks up #1470 grepping the wrong document. Both doc lines now name the real locations. **Issue #1470's body carried the same mis-attribution and has been corrected in place** with a dated note, plus a sharper scope step: a hand-maintained line count living in three historical ledgers may be better dropped than refreshed.

The omission half — the actual point of #1470 — was verified and is unchanged: `grep -cE "ApprovedRevisionId|DeferredUntil" docs/architecture/DATA_MODEL.md` → **0**.

### Refuted (5) — recorded, not dropped

- **"STATUS.md:11 contradicts STATUS.md:333"** — refuted: `:333` is a dated as-delivered ledger entry ("delivered at 855 lines in April 2026"), not a current-state assertion, and it is untouched by this diff. Both can be true at once.
- **"The masterplan repeats the mis-attribution"** — refuted twice over: its line makes no location claim at all (no possessive, no "self-description"), and it compresses #1470's own title.
- **"The masterplan contains no 855, unlike STATUS"** — refuted, and the finding's own load-bearing check was wrong: `IMPLEMENTATION_MASTERPLAN.md:1623` **does** carry it. Noted because the whole argument rested on a contrast that does not exist.
- **"An uncommitted working-tree edit removes the `src/tests/**` exclusion"** — refuted: a **stale stat cache**, not content. All three blobs (worktree / index / HEAD) hash to `1aae9932`; `git diff` was empty throughout. An artefact of #1469's lift-and-revert touching the file's mtime. Worth recording as a review-noise pattern: ` M` in `git status` after a restore-from-backup can be a mtime ghost, and the hash is the arbiter.
- **"The PR body says 383 belongs on #1468's record but #1468 was never updated"** — refuted as a defect (the PR body is not the diff, and neither doc line claims the issue was updated). Acted on anyway: #1468 now carries the measurement.

### Verification on `5f490ddf`

- `node scripts/check-docs-governance.mjs` — passed
- `node scripts/check-golden-principles.mjs` — passed
- 415/0 re-measured directly, `tsconfig.app.json` confirmed byte-identical to HEAD afterwards
- `grep` counts for the `DATA_MODEL.md` claims re-run and quoted above
- Diff remains two prose lines across the two canonical docs; no code, config, workflow or schema touched

Chris0Jeky:
@codex review

Issue #1598, opened 2026-08-07
[MCP] ModelContextProtocol 2.0.0 flips HttpServerTransportOptions.Stateless to true and drops Mcp-Session-Id


Chris0Jeky:
## Summary

Dependabot PR #1597 bumps `ModelContextProtocol` and `ModelContextProtocol.AspNetCore` from **1.4.1 to 2.0.0** (a stable major, not a prerelease). It fails the **required** gate — `ci-required.yml` run 30808552160 at head `ac50b519`, with `API Integration` red on **both** ubuntu-latest and windows-latest. This is deterministic on both platforms, not a flake.

It must not be merged as a dependency bump. Landing it needs a deliberate decision and a code change.

## Root cause

Exactly one test fails; 2183 pass:

- `Taskdeck.Api.Tests.McpHttpTransportApiKeyTests.McpEndpoint_UserScopedResource_IsolatedByApiKeyOwner`
- `backend/tests/Taskdeck.Api.Tests/McpHttpTransportApiKeyTests.cs:700`
- The failing assertion is `initializeResponse.Headers.TryGetValues("Mcp-Session-Id", out var sessionValues).Should().BeTrue()`

That is a direct hit on 2.0.0's first breaking change: `HttpServerTransportOptions.Stateless` now defaults to **`true`**. Stateless servers do not create transport sessions and therefore never emit `Mcp-Session-Id`.

Both MCP hosts call `AddMcpServer().WithHttpTransport()` with no options — `backend/src/Taskdeck.Api/Program.cs:130` and `Program.cs:374` — so both silently flip to stateless on the bump.

## What is *not* broken

The declarative tool/resource surface is unaffected. `backend/src/Taskdeck.Api/Mcp/*` uses only `[McpServerResourceType]`, `[McpServerResource]`, `[McpServerToolType]`, and `[McpServerTool]` (BoardResources, CaptureResources, ProposalResources, ProposalTools, ReadTools, WriteTools), and the failing run shows **0 `error CS`** and **0 MCP9005/MCP9006 analyzer warnings**. Nothing in `Mcp/*` moved at compile level.

The break is purely the transport default.

## Why merging as-is would be wrong

It would ship an unannounced MCP **wire-contract** change to existing clients — no session IDs, no standalone SSE GET/DELETE, no unsolicited server-to-client requests — and it would leave the user-scoped API-key isolation path unexercised, since the test that proves that isolation is the one that fails.

## Options

[...]

PR #1601, opened 2026-08-07
Pin MCP transport Stateless=false and take ModelContextProtocol 2.0.0

Chris0Jeky:
Refs #1598. Supersedes dependabot PR #1597, which cannot merge as-is.

## Why #1597 is red

`ModelContextProtocol` 2.0.0 flips `HttpServerTransportOptions.Stateless` from `false` to `true`. A stateless server never creates a transport session, so it stops emitting `Mcp-Session-Id` and drops standalone SSE GET/DELETE and unsolicited server-to-client requests.

Both HTTP hosts called `WithHttpTransport()` with no options, so both would have inherited that flip silently. On #1597 that showed up as a **required**-gate failure — `ci-required` run 30808552160 at `ac50b519`, `API Integration` red on ubuntu-latest *and* windows-latest, deterministic on both. Exactly one test failed out of 2184:

`McpHttpTransportApiKeyTests.McpEndpoint_UserScopedResource_IsolatedByApiKeyOwner` at `McpHttpTransportApiKeyTests.cs:700`, asserting `Mcp-Session-Id` is present on the initialize response.

Merging #1597 would have shipped an unannounced MCP wire-contract change to existing clients, and would have left the user-scoped API-key **isolation** path unexercised — the failing test is precisely the one proving that isolation.

## This change

Takes the 2.0.0 bump and pins the contract:

- `backend/Directory.Packages.props` — `ModelContextProtocol` and `ModelContextProtocol.AspNetCore` 1.4.1 to 2.0.0
- `Program.cs:130` (standalone `--mcp --transport http` host) and `Program.cs:374` (co-hosted) — `WithHttpTransport(options => options.Stateless = false)`

Set **explicitly** rather than relying on the library default, so a future default flip cannot move the contract again without an edit here. `Program.cs:239` uses `WithStdioServerTransport` and is unaffected.

This is the **option 1** disposition recorded on #1598: preserve today's contract, no behavior change, therefore no ADR and no `STATUS.md` update. Deliberately adopting stateless remains option 2 and would require an ADR plus replacement isolation coverage.

The declarative tool/resource surface needed no changes — `Mcp/*` uses only the attribute API, and the build shows 0 `error CS` and 0 MCP9005/MCP9006 analyzer warnings.

## Verification

| Check | Result |
|---|---|
| `dotnet build backend/src/Taskdeck.Api/Taskdeck.Api.csproj -c Release` | succeeded, no new warnings |
| Full `Taskdeck.Api.Tests` (the suite behind the red `API Integration` job) | **2184 passed / 0 failed / 4 skipped** |
| `McpHttpTransportApiKeyTests` + `StandaloneMcpHostFilteringTests` | **55/55** |

2184 passing is exactly #1597's 2183 plus the one test that was failing, so the fix resolves it and nothing else moved.

No new test was added because the coverage already exists and is **proven discriminating by CI itself** — `McpEndpoint_UserScopedResource_IsolatedByApiKeyOwner` failed on #1597 without this pin and passes with it. That is stronger evidence than a test I wrote and asserted about myself.

## Found while doing this — filed separately, not fixed here

While checking whether the standalone `--mcp --transport http` host was also covered, I found it has **no session coverage at all** and, more importantly, that it **500s on every authenticated MCP request** — on current `main`, at 1.4.1, independent of this change. Root cause is missing CORS middleware. Filed with the stack trace and a repro. It is pre-existing and out of scope here; this PR does not make it better or worse.

**Not verified:** hosted CI at this head has not run yet. The standalone host's `Stateless` pin at `Program.cs:130` is therefore correct-by-inspection but not exercised end-to-end, because that host cannot currently serve a request at all for the unrelated reason above.


Chris0Jeky:
## Independent review: two lenses, both SHIP — one MEDIUM fixed in `3d650a5c`

Ran a fresh-context adversarial review (contract/API-surface lens + evidence/test-adequacy lens). Both returned **SHIP**. Highlights worth recording:

**The `Stateless` claim is now verified directly, not inferred.** The contract lens built throwaway console projects referencing `ModelContextProtocol.AspNetCore` 1.4.1 and 2.0.0 and reflected over a freshly constructed `HttpServerTransportOptions`: **1.4.1 `Stateless=False`, 2.0.0 `Stateless=True`** — and all nine other properties (`ConfigureSessionOptions`, `RunSessionHandler`, `EnableLegacySse`, `EventStreamStore`, `SessionMigrationHandler`, `PerSessionExecutionContext`, `IdleTimeout`, `MaxIdleSessionCount`, `TimeProvider`) are byte-identical across versions. So `Stateless` is the **only** transport default that moved, and pinning it restores the complete 1.4.1 HTTP surface rather than patching one symptom. `McpServerOptions` defaults are likewise unchanged.

**Removed 2.0.0 APIs are unused here.** `McpServerOptions.TaskStore`, `SendTaskStatusNotifications`, and `McpServerToolAttribute.TaskSupport` are gone in 2.0.0; Taskdeck references none, proven by a clean `--no-incremental` build with 0 errors and 0 warnings.

**No third host was missed.** A repo-wide grep for `AddMcpServer|WithHttpTransport|WithStdioServerTransport|HttpServerTransportOptions` returns exactly three registrations: the two HTTP hosts (both pinned) and the stdio host (correctly untouched — `HttpServerTransportOptions` does not apply). `MapMcp` appears once, in the shared `McpEndpointMapping`, so the two HTTP hosts cannot diverge. No lock file, and no MCP version references in `deploy/`, Dockerfiles, README, AGENTS.md, or `scripts/mcp/*`.

**The one MEDIUM, now fixed.** Both lenses independently flagged that `docs/STATUS.md:471` still asserted "current packages pin `ModelContextProtocol` 1.4.1" in the present tense. My commit message had reasoned "no behavior change, so no STATUS update" — correct about the wire contract, wrong about that factual sentence. Fixed in `3d650a5c`, which also records *why* the explicit `Stateless` pin exists so a future reader does not delete it as redundant. Both governance checkers pass. STATUS:281/282 also name v1.2.0, but those are historical spike records and correctly left alone.

**Informational, not blocking:** the 2026-07-28 protocol revision arrives with 2.0.0, and every MCP test pins `2025-11-25`. With `Stateless=false` a client that *explicitly* pins 2026-07-28 is refused with `-32022 UnsupportedProtocolVersion` so a dual-path client downgrades — which is the documented, intended semantics of the option chosen here. The SDK's default client probes and auto-falls-back, so no stall. Also noted: the MCP SDK's Roslyn analyzer never reaches the compilation (analyzer assets do not flow transitively through `ModelContextProtocol.Core`), so the clean 0-warning build is genuine but is not evidence that MCP-specific analyzers approved anything. Both pre-existing and unchanged by this PR.

The `3d650a5c` push is documentation-only and changed no logic, so no further independent pass is owed under global law 2 — but the aging window resets and CI re-runs at the new head.

Issue #1512, opened 2026-07-27
[CI] Diagnose post-#1373 Windows concurrent-card HTTP 500


Chris0Jeky:
## Problem

Required CI on PR #1505, exact source head `6ee17a4849427d679722a9d1e034d4622536f4fe`, failed only on `API Integration (windows-latest)` in run `30239134685`, job `89892554337`.

`CardUpdateConflictTests.ConcurrentCardCreation_SameColumn_AllCreatedNoDuplicates` issued five concurrent create-card requests; two returned HTTP 500 instead of 201. Windows finished 2,135 passed / 4 skipped / 1 failed of 2,140. Ubuntu passed 2,136 / 4 / 0 on the same head.

The failing test, `TestWebApplicationFactory`, `CardService`, and `UnitOfWork` are byte-identical to prior green PR head `bf4549a7` and locally full-green `484b5537`; later commits changed only docs/skills. This is not caused by #1505's LLM/provider diff.

## Why this is a successor

#1282 tracked the earlier Windows-only concurrent HTTP-500 family and PR #1373 restored production SQLite timeout/WAL registration parity. This post-#1373 recurrence proves that resolution was necessary but not sufficient, or that a second cause shares the same symptom. The current TRX cannot distinguish them: the factory clears logging providers and the test asserts only status codes, so `SQLITE_BUSY` is plausible but unproven.

## Acceptance criteria

- [ ] Preserve the first causal server-side error safely: failed concurrent responses include bounded non-secret diagnostic context and correlation identity; the test/harness captures the corresponding middleware exception classification without logging request bodies, credentials, or user content.
- [ ] Reproduce or falsify the current card-creation failure with 20 Windows runs of the exact test and five Windows runs of the API project under CI-equivalent configuration.
- [ ] Root-cause the 500 from evidence. Do not assume SQLite contention, add HTTP retries, quarantine the test, or accept 500 as success.
- [ ] Fix the proven cause at the narrowest production/test-harness seam and add a regression that fails before the fix.
- [ ] Re-run the historical #1282 auth/capture/queue/webhook concurrency families plus the current card-creation test on Windows.
- [ ] Run the full serialized backend suite and exact-head hosted Required CI; both Windows and Ubuntu API lanes must be green.
- [ ] If the defect is environmental/nondeterministic rather than product logic, retain a deterministic diagnostic assertion and document the measured trigger instead of deleting evidence.

## Immediate gate

[...]

Issue #1357, opened 2026-07-13
Flaky: SerializedMigratorTests.Concurrent_migrators_apply_schema_exactly_once (windows API Integration)


Chris0Jeky:
**Classification:** low-frequency environmental flake — investigated, not dismissed (never-merge-failing-CI / never-dismiss-as-flaky discipline).

**Symptom:** `SerializedMigratorTests.Concurrent_migrators_apply_schema_exactly_once` intermittently fails on `API Integration (windows-latest)` with `Microsoft.Data.Sqlite.SqliteException: SQLite Error 1: 'table "ProposalProvenances" already exists'` — i.e. two concurrent migrators both applied an *old* migration, so the advisory lock did not serialize them that run.

**Evidence it is flaky (not a regression):** failed once on `main@04e7969d` (the 2026-07-13 revival/GEN wave head), then **passed on `main@1f65fc24`** (docs-only diff — identical `SerializedMigrator` + migration code). Not in the failure ledger before; not the known #1282/#1335 hosted-worker race.

**Root-cause read:** the test runs 2 migrators against a fresh cold DB behind a `Barrier`, 4 iterations. The wave added 2 migrations (`AddRegistrationGating`, `AddArtefactExtractions`), lengthening the cold-DB chain and widening the window where both migrators can take the lock-free fail-open path (uncreatable/timed-out lock under CI load) and collide mid-chain; the loser's `NoMigrationsPending` re-check sees migrations still pending (winner not finished) and rethrows. `SerializedMigrator` itself (WAL + busy_timeout + advisory file lock + pending re-check) is sound; this is test-harness concurrency fragility on shared CI.

**Suggested hardening (pick one):**
- Make the assertion state-based: allow the documented fail-open collision but assert the *final* schema is applied exactly once (`__EFMigrationsHistory` complete, no dupes), rather than `NotThrowAsync`.
- Or give the test a longer lock timeout and/or fewer iterations to reduce the CI-load window.
- Or investigate why the sidecar lock file becomes uncreatable/contended under Windows CI load and make `AcquireLock` retry that case.

**Severity:** LOW (test-only; product mechanism is sound). Reruns green.

Issue #1282, opened 2026-07-03
Flaky: ConcurrencyRaceConditionStressTests.BoardCreation_ConcurrentMultiUser_NoCrossContamination — transient 500 during concurrent auth setup


Chris0Jeky:
## Symptom

On the required **API Integration (windows-latest)** lane, `BoardCreation_ConcurrentMultiUser_NoCrossContamination` intermittently fails with:

```
Expected the enum to be HttpStatusCode.OK {value: 200}, but found HttpStatusCode.InternalServerError {value: 500}.
  at ...ConcurrencyRaceConditionStressTests.cs:line 808
```

Line 808 is `ApiTestHarness.AuthenticateAsync(...)`, run for 5 users concurrently behind a `SemaphoreSlim` barrier. So the 500 comes from the **concurrent auth/registration setup**, not board creation. ubuntu-latest passed the same commit; 1844/1845 passed on windows. Observed on PR #1280 (a frontend-only change — the failure is unrelated to the PR diff).

## Likely cause

Transient SQLite contention (SQLITE_BUSY-class) when 5 auth requests hit the DB simultaneously on the slower/more-contended windows runner. WAL + busy_timeout landed in #1130, but the stress harness fans out with no tolerance for a transient 5xx during setup. Related surface: #1206 (change-tracker staleness), #1194.

## Options

- Harden `AuthenticateAsync` (or the auth endpoint) against transient contention (bounded retry on 5xx during test setup), or stagger/serialize the concurrent auth setup so the *board-creation* race — the actual thing under test — is what's stressed, not user registration.
- Or classify + quarantine per ARCHIVE-07 (#1275) CI right-sizing; this blocks exit-criterion (h) "zero always-red CI lanes" when it flakes a required lane.

## Acceptance

- [ ] The test is deterministic on windows-latest across repeated runs (no transient-500 setup flake).

Chris0Jeky:
Additional occurrence during the 2026-07-13 overnight run:

- PR #1328 (docs-only, head `c1bce5d`) failed required `API Integration (windows-latest)`.
- Failure: `QueueAccumulationResilienceTests.RapidCaptureSubmission_DoesNotCorruptQueue` received one HTTP 500 from 10 concurrent capture POSTs; 1,845/1,846 tests passed. Ubuntu passed the same merge commit.
- Narrow local Windows reproduction on current `main` passed once:
[...]

PR #1610, opened 2026-08-08
Disable SQLite pooling in tests instead of clearing the pool globally

### LINKED ISSUE #1609: Flaky: SqliteConnection.ClearAllPools() in 13 test files is process-global and races any concurrent SQLite Open()
## Symptom

`API Integration (windows-latest)` failed on PR #1608 with a test that has nothing to do with that
PR — which changes **zero** backend files.

```
Taskdeck.Api.Tests.ProposalRevisionRepositoryBatchIntegrationTests.GetRefsByProposalIdsAsync_DoesNotSelectTheRevisionPayload [FAIL]
  System.ObjectDisposedException : Cannot access a disposed object.
  Object name: 'SQLitePCL.sqlite3'.
     at SQLitePCL.SQLite3Provider_e_sqlite3.NativeMethods.sqlite3_create_function_v2(...)
     at Microsoft.Data.Sqlite.SqliteConnection.Open()
     at Microsoft.EntityFrameworkCore.Migrations.HistoryRepository.ExistsAsync(...)
     at ...ProposalRevisionRepositoryBatchIntegrationTests.GetRefsByProposalIdsAsync_DoesNotSelectTheRevisionPayload():251
```

Run [31227769095](https://github.com/Chris0Jeky/Taskdeck/actions/runs/31227769095), head `1700d404`.
1 failed / 2183 passed / 4 skipped.

## Root cause — `ClearAllPools()` is process-global, and the isolation is only apparent

`ProposalRevisionRepositoryBatchIntegrationTests.Cleanup()` (`:397-399`) ends every test with:

```csharp
SqliteConnection.ClearAllPools();
```

Each test creates its own database file — `taskdeck-revision-batch-{Guid:N}.db` — so the tests *look*
mutually isolated. They are not, because **`ClearAllPools()` is a static, process-wide operation**. It
disposes the pooled connections for *every* SQLite database in the process, not just the one the
calling test owns.

xUnit runs each test class as its own collection in parallel by default, and this repo declares no
`CollectionBehavior` / `maxParallelThreads` override anywhere under `backend/tests`. So the race is:

1. Test **A** finishes and calls `ClearAllPools()`, disposing every pooled `sqlite3` handle.
2. Test **B**, running concurrently in another collection, is inside
   `MigrateAsync` → `HistoryRepository.ExistsAsync` → `SqliteConnection.Open()`, and has already
   taken a handle from the pool.
3. `Open()` calls `sqlite3_create_function_v2` on that handle, which A has just disposed →
   `ObjectDisposedException: 'SQLitePCL.sqlite3'`.

The stack is diagnostic of exactly this: the failure is on the **native handle**, inside `Open()`, on
the **first** connection the migration opens. Nothing in the test's own logic can produce it.

## Why this is a defect and not a rerun-and-move-on

**13 test files call `ClearAllPools()`**, including `TestWebApplicationFactory.cs:106` — the shared
factory a large fraction of the API tests build on. Every one of them is a process-wide grenade for
whatever else is mid-`Open()`:

```
ApiKeyMiddlewareLastUsedPersistenceTests.cs:136      NotificationRepositoryIntegrationTests.cs:420
ApiKeyMiddlewarePerKeyRateLimitTests.cs:392          ProposalRevisionRepositoryBatchIntegrationTests.cs:399
ArtefactExtractionRepositoryBatchIntegrationTests.cs:266   SerializedMigratorTests.cs:625
ConnectorEncryptionKeyFailFastTests.cs:131, :189     SourceArtefactRepositoryIntegrationTests.cs:238
LlmQueueRepositoryIntegrationTests.cs:1251           SqlitePragmaInterceptorTests.cs:123
StandaloneMcpHostFilteringTests.cs:392, :612         TestWebApplicationFactory.cs:106
TranscriptRepositoryIntegrationTests.cs:104
```

So the observable failure is not specific to `ProposalRevisionRepositoryBatchIntegrationTests` — that
class is simply the one that lost the race this time. Any SQLite-backed test in the assembly can be
the victim, and any of the 13 can be the shooter. That is consistent with this repo's history of
one-off, non-reproducing Windows integration failures that were each investigated in isolation
(`#1282`, `#1357`, `#1512`).

## Evidence that it is pre-existing and timing-dependent

Byte-identical backend code, same job, four runs:

| Head | `API Integration (windows-latest)` |
|---|---|
| `28b9fe8f` (main) | pass |
| `61214efe` (PR #1608) | pass |
| `49ca6674` (PR #1608) | pass |
| `1700d404` (PR #1608) | **fail** |

PR #1608 changes only `docs/**` and `frontend/**` — `git diff --name-only main...HEAD | grep -c backend`
returns **0**. It cannot have caused this.

## Proposed fix

**Stop pooling instead of clearing the pool.** These tests use a throwaway file per test and gain
nothing from connection pooling; the `ClearAllPools()` calls exist only so the file can be deleted on
Windows, which holds a lock while a pooled connection survives.

```csharp
.UseSqlite($"Data Source={dbPath};Pooling=False")
```

With pooling off, nothing is retained after the `DbContext` is disposed, the file deletes cleanly, and
`Cleanup()` no longer needs a process-global call at all. That removes the shared mutable state rather
than trying to time it.

Do it as one sweep across all 13 sites, not file by file — a partial fix leaves the grenade in play,
and a victim can be any SQLite test in the assembly.

**Rejected alternative:** serialising the affected classes into one xUnit collection. It would work, but
it trades a correctness bug for a wall-clock cost across a large share of the suite, and it leaves the
process-global call in place for anything added later.

## Acceptance criteria

- [ ] No `SqliteConnection.ClearAllPools()` remains under `backend/tests/`, or every remaining call is
      justified in a comment explaining why it cannot race.
- [ ] Every affected test harness disables pooling in its connection string instead.
- [ ] Temp database files are still deleted on Windows — verify no `taskdeck-*.db` leaks in
      `Path.GetTempPath()` after a full `Taskdeck.Api.Tests` run.
- [ ] `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1` passes,
      and the previously-failing test passes in a run where the other 12 call sites also execute.
- [ ] **Mutation-check the isolation claim:** with the fix in place, deliberately reintroduce one
      `ClearAllPools()` and confirm the race can still be provoked (a stress loop over the affected
      classes), so the fix is shown to be what removed it rather than timing luck.

## Note on how this was handled on #1608

The failed job was re-run after the cause was diagnosed and shown to be independent of the diff — not
before. #1608 will only be merged on a genuinely green head. Filing this so the next occurrence is
recognised as a known, root-caused defect with a proposed fix, rather than investigated from scratch
for a fourth time.

Chris0Jeky:
Closes #1609.

## The bug

`SqliteConnection.ClearAllPools()` is **process-global**. It disposes the pooled connections of every
SQLite database in the process, not just the caller's.

13 test files called it — including `TestWebApplicationFactory`, which a large share of the API suite
builds on. Each test creates its own `{Guid}.db`, so the tests *look* mutually isolated. They were
not. xUnit runs test classes as parallel collections by default and this assembly declares no
parallelism override, so that call was **shared mutable state reachable from every test class at
once**, and a victim died inside `SqliteConnection.Open()` with
`ObjectDisposedException: 'SQLitePCL.sqlite3'` on a migration's first connection.

(An earlier revision of this body set out a confident three-step interleaving here. The review
falsified part of it — see *What is not proven* below — so it is gone. The shared state and
the failure are facts; the exact window is not.)

That is the CI red on #1608 — a PR that changed **zero** backend files. Any of the 13 could be the
shooter; any SQLite-backed test in the assembly could be the victim, which is why it has read as an
unrelated one-off each time (#1282, #1357, #1512).

## The fix

Route every **file-backed** connection string through a new `TestSqlite.ConnectionString()`, which
appends `Pooling=False`, and delete all `ClearAllPools()` calls — **15 call sites across 13 files** (`ConnectorEncryptionKeyFailFastTests` and `StandaloneMcpHostFilteringTests` have two each; the commit message for `015eafb9` miscounts these as 13 calls).

With pooling off, no handle outlives its connection: the files unlock as soon as the connection is
disposed, cleanup can delete them, and **there is no pool to clear**. The shared mutable state is
removed rather than timed.

The helper exists so the decision has one home and a written rationale, instead of `Pooling=False`
appearing in a dozen interpolated strings where the next person would reasonably delete it as noise.

In-memory connection strings (`Data Source=:memory:`) are deliberately untouched — they back no file,
so they never needed the pool cleared.

## The second bug, which I did not expect

**The old approach was not merely racy — it was not working.**

`ClearAllPools()` was there so Windows would release the `.db`/`-wal`/`-shm` lock and the delete loop
could succeed. It frequently did not, and the delete loop's `catch` swallows the `IOException`
silently. Measured on this box before the change:

> **3,033 leaked `taskdeck*.db` files, 841 MB, dating back to 2026-08-01.**
> Biggest contributor `taskdeck-api-tests-*` (2,049 files — `TestWebApplicationFactory`).
> Disk was at **97%**.

So the call being removed was paying a real correctness cost for a benefit it was not reliably
delivering.

**Corrected after review:** this change does **not** make the assembly leak-free, and an earlier
revision of this body plus the commit message for `015eafb9` said it did ("this run leaked zero").
An independent full-suite run on this branch leaks **zero `-wal`/`-shm`** but still accumulates
**~199 zero-byte `.db.migrate.lock` files and one real `.db`**. The `.migrate.lock` half is
pre-existing and unrelated to pooling: `TestWebApplicationFactory.GetDatabaseCleanupTargets`
enumerates `.db`/`-wal`/`-shm`/`-journal` and never `.migrate.lock`, where `SerializedMigratorTests`
and the CLI harness both do. Filed as a follow-up. History is not rewritten; the commit message
stands uncorrected and this note is the record.

## Verification

**Full suite, this branch:** `dotnet test backend/tests/Taskdeck.Api.Tests -c Release -m:1` →
**2,184 passed / 0 failed / 4 skipped** (7 m 19 s). Same total as `main`, with
`ProposalRevisionRepositoryBatchIntegrationTests.GetRefsByProposalIdsAsync_DoesNotSelectTheRevisionPayload`
— the test that failed on #1608 — now passing.

**Mutation-checked, one variable, same 10 tests.** `taskdeck-revision-batch-*` is owned by exactly one
test class, so it isolates the measurement from everything else:

| | files present | leaked by the run |
|---|---|---|
| baseline (quiesced) | 3 | — |
| **with** `Pooling=False` | 3 | **0** |
| **without** `Pooling=False` (mutation) | 13 | **+10 — exactly one per test** |
| fix restored | 13 | **0** |

Ten tests, ten leaked files without the fix, none with it. The mutation was reverted and the tree
re-proved clean.

### What is *not* proven — narrowed further after review

I originally wrote that "the mechanism" was established from the stack trace plus the documented
process-global semantics of `ClearAllPools()`. **The review falsified part of that**, by probing
Microsoft.Data.Sqlite 8.0.29 directly: the pool-clearing call does **not** dispose a connection that
is currently checked out, and ~233,000 racing cycles produced no failure. A synthetic probe failing
to reproduce is weak evidence of absence — but it is enough that I should not have called the
step-by-step interleaving established, and I had baked it into a permanent code comment. Narrowed in
`32568178`.

What *is* established: the call was **process-global shared mutable state that no test owned**,
reachable from every parallel collection at once; the failure was real, was in a PR that changed no
backend code, and matches three prior one-off investigations (#1282, #1357, #1512); and removing the
shared state removes the class of problem. The *leak* half remains proven by the controlled
experiment above. The precise interleaving is **inferred, not demonstrated** — please do not repeat
it downstream as fact.

Also not run: the other backend test projects. None of them references `TestSqlite` or contained a
`ClearAllPools()` call — `grep -rn "ClearAllPools" backend/ --include=*.cs` now returns **nothing
at all** (the helper's docs say *the pool-clearing call*, never the identifier). Note that
`Taskdeck.Cli.Tests` and `Taskdeck.Application.Tests` still build pooled file-backed connection
strings; they run in separate test-host processes so there is no cross-assembly race, but
"stops new leaks" is true of `Taskdeck.Api.Tests` only.

## Note for the maintainer

The **841 MB of pre-existing leaked files in `%TEMP%` are still there** — this change stops the
`-wal`/`-shm` leaks in `Taskdeck.Api.Tests` (see the corrected scope above) and cleans up none of
the existing files. I deliberately did not bulk-delete them: that is 3,000+ files in your
temp directory and deleting from a listing is exactly the pattern the safety rules forbid. With the
disk at 97% it is probably worth reclaiming, but by your hand, not mine. My own mutation experiment
added 10 more, which I also left in place rather than reach for a wildcard delete.

### DIFF
--- a/backend/tests/Taskdeck.Api.Tests/ApiKeyMiddlewareLastUsedPersistenceTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/ApiKeyMiddlewareLastUsedPersistenceTests.cs
@@ -98,7 +98,7 @@ public sealed class ApiKeyMiddlewareLastUsedPersistenceTests : IDisposable
     private Task<TaskdeckDbContext> CreateContextAsync()
     {
         var options = new DbContextOptionsBuilder<TaskdeckDbContext>()
-            .UseSqlite($"Data Source={_dbPath}")
+            .UseSqlite(TestSqlite.ConnectionString(_dbPath))
             .Options;
         return Task.FromResult(new TaskdeckDbContext(options));
     }
@@ -133,7 +133,6 @@ public sealed class ApiKeyMiddlewareLastUsedPersistenceTests : IDisposable
 
     public void Dispose()
     {
-        SqliteConnection.ClearAllPools();
         foreach (var suffix in new[] { "", "-wal", "-shm", "-journal" })
         {
             var path = _dbPath + suffix;
--- a/backend/tests/Taskdeck.Api.Tests/ApiKeyMiddlewarePerKeyRateLimitTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/ApiKeyMiddlewarePerKeyRateLimitTests.cs
@@ -338,7 +338,7 @@ public sealed class ApiKeyMiddlewarePerKeyRateLimitTests : IDisposable
     private async Task<TaskdeckDbContext> CreateSeededContextAsync()
     {
         var options = new DbContextOptionsBuilder<TaskdeckDbContext>()
-            .UseSqlite($"Data Source={_dbPath}")
+            .UseSqlite(TestSqlite.ConnectionString(_dbPath))
             .AddInterceptors(_interceptor)
             .Options;
         var db = new TaskdeckDbContext(options);
@@ -389,7 +389,6 @@ public sealed class ApiKeyMiddlewarePerKeyRateLimitTests : IDisposable
 
     public void Dispose()
     {
-        SqliteConnection.ClearAllPools();
         foreach (var suffix in new[] { "", "-wal", "-shm", "-journal" })
         {
             var path = _dbPath + suffix;
--- a/backend/tests/Taskdeck.Api.Tests/ArtefactExtractionRepositoryBatchIntegrationTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/ArtefactExtractionRepositoryBatchIntegrationTests.cs
@@ -221,7 +221,7 @@ public sealed class ArtefactExtractionRepositoryBatchIntegrationTests
         var dbPath = Path.Combine(Path.GetTempPath(), $"taskdeck-extraction-batch-{Guid.NewGuid():N}.db");
         var interceptor = new CapturingCommandInterceptor();
         var options = new DbContextOptionsBuilder<TaskdeckDbContext>()
-            .UseSqlite($"Data Source={dbPath}")
+            .UseSqlite(TestSqlite.ConnectionString(dbPath))
             .AddInterceptors(interceptor)
             .Options;
         return (options, interceptor, dbPath);
@@ -263,7 +263,6 @@ public sealed class ArtefactExtractionRepositoryBatchIntegrationTests
 
     private static void Cleanup(string dbPath)
     {
-        SqliteConnection.ClearAllPools();
         foreach (var suffix in new[] { "", "-wal", "-shm", "-journal" })
         {
             var path = dbPath + suffix;
--- a/backend/tests/Taskdeck.Api.Tests/ConnectorEncryptionKeyFailFastTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/ConnectorEncryptionKeyFailFastTests.cs
@@ -92,7 +92,7 @@ public class ConnectorEncryptionKeyFailFastTests
         var config = new ConfigurationBuilder()
             .AddInMemoryCollection(new Dictionary<string, string?>
             {
-                ["ConnectionStrings:DefaultConnection"] = $"Data Source={dbPath}",
+                ["ConnectionStrings:DefaultConnection"] = TestSqlite.ConnectionString(dbPath),
                 ["Connectors:EncryptionKey"] = validKey,
                 ["Database:BusyTimeoutMilliseconds"] = "5000",
             })
@@ -128,7 +128,6 @@ public class ConnectorEncryptionKeyFailFastTests
         }
         finally
         {
-            Microsoft.Data.Sqlite.SqliteConnection.ClearAllPools();
             foreach (var suffix in new[] { "", "-wal", "-shm" })
             {
                 var path = dbPath + suffix;
@@ -156,7 +155,7 @@ public class ConnectorEncryptionKeyFailFastTests
         var config = new ConfigurationBuilder()
             .AddInMemoryCollection(new Dictionary<string, string?>
             {
-                ["ConnectionStrings:DefaultConnection"] = $"Data Source={dbPath}",
+                ["ConnectionStrings:DefaultConnection"] = TestSqlite.ConnectionString(dbPath),
                 ["Connectors:EncryptionKey"] = validKey,
             })
             .Build();
@@ -186,7 +185,6 @@ public class ConnectorEncryptionKeyFailFastTests
         }
         finally
         {
-            Microsoft.Data.Sqlite.SqliteConnection.ClearAllPools();
             foreach (var suffix in new[] { "", "-wal", "-shm" })
             {
                 var path = dbPath + suffix;
--- a/backend/tests/Taskdeck.Api.Tests/LlmQueueRepositoryIntegrationTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/LlmQueueRepositoryIntegrationTests.cs
@@ -1236,7 +1236,7 @@ public class LlmQueueRepositoryIntegrationTests : IClassFixture<HostedWorkerDisa
         try
         {
             var builder = new DbContextOptionsBuilder<TaskdeckDbContext>()
-                .UseSqlite($"Data Source={dbPath}");
+                .UseSqlite(TestSqlite.ConnectionString(dbPath));
             if (interceptor != null)
             {
                 builder.AddInterceptors(interceptor);
@@ -1248,7 +1248,6 @@ public class LlmQueueRepositoryIntegrationTests : IClassFixture<HostedWorkerDisa
         }
         finally
         {
-            SqliteConnection.ClearAllPools();
             foreach (var suffix in new[] { "", "-wal", "-shm", "-journal" })
             {
                 var path = dbPath + suffix;
--- a/backend/tests/Taskdeck.Api.Tests/McpBoardResourcesTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/McpBoardResourcesTests.cs
@@ -35,7 +35,7 @@ public class McpBoardResourcesTests : IDisposable
             new ConfigurationBuilder()
                 .AddInMemoryCollection(new Dictionary<string, string?>
                 {
-                    ["ConnectionStrings:DefaultConnection"] = $"Data Source={_dbPath}",
+                    ["ConnectionStrings:DefaultConnection"] = TestSqlite.ConnectionString(_dbPath),
                     ["Connectors:EncryptionKey"] = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
                 })
                 .Build());
--- a/backend/tests/Taskdeck.Api.Tests/McpResourcesTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/McpResourcesTests.cs
@@ -33,7 +33,7 @@ public class McpResourcesTests : IDisposable
             new ConfigurationBuilder()
                 .AddInMemoryCollection(new Dictionary<string, string?>
                 {
-                    ["ConnectionStrings:DefaultConnection"] = $"Data Source={_dbPath}",
+                    ["ConnectionStrings:DefaultConnection"] = TestSqlite.ConnectionString(_dbPath),
                     ["Connectors:EncryptionKey"] = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
                 })
                 .Build());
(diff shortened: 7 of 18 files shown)


Chris0Jeky:
## Parked pending its review verdict — run wrapped up by the maintainer

Recording state so this is not merged on incomplete evidence.

**Do not merge yet.** A fresh-context adversarial review was in flight when the run ended and **its verdict was never collected**. It was briefed to attack the things most likely to be wrong here: whether `Pooling=False` broke any test that depended on pooling (particularly WAL sidecar behaviour in `SqlitePragmaInterceptorTests`, `SerializedMigratorTests`, `MigrationBootstrapTests` — with pooling off, closing the last connection checkpoints and removes `-wal`/`-shm`), sweep completeness, comments corrupted by the mechanical script, and overclaiming. Whoever picks this up should collect that verdict or run a fresh review; the evidence below is mine alone and has not been independently challenged.

**Evidence as it stands:**

| Check | Result |
|---|---|
| `dotnet test backend/tests/Taskdeck.Api.Tests -c Release -m:1` | **2,184 passed / 0 failed / 4 skipped** (7m19s) — same total as main, with the test that failed on #1608 now passing |
| `grep -rn "ClearAllPools" backend/ --include=*.cs` | empty apart from this helper's own docs |
| leak, mutation-checked | 10 tests → **+10 leaked files without `Pooling=False`, 0 with it**; mutation reverted, tree re-proved clean |
| CI at head `015eafb9` | 24 pass / 11 skip / **1 pending** (`API Integration (windows-latest)` — fittingly, the job whose flake this fixes) |

**Still not proven, and stated in the PR body too:** the race itself was never reproduced. The *leak* half is experimental; the *race* half rests on the stack trace plus the documented process-global semantics of `ClearAllPools()`, and on the shared state now being gone. Please do not upgrade that claim on my behalf.

**Not done here:** the ~841 MB of pre-existing leaked databases in `%TEMP%`. This PR stops new ones; it deletes none. That is the maintainer's call, not something I will do from a file listing.

Chris0Jeky:
## Review round: **SHIP**, no merge blockers — 2 MEDIUM claim defects corrected in `32568178`, 1 follow-up filed (#1611)

A fresh-context adversarial review returned. It independently reproduced the full suite (**2,184 passed / 0 failed / 4 skipped**, matching my numbers exactly), confirmed `grep -rn "ClearAllPools" backend/` is empty, confirmed no parallelism override exists anywhere in `backend/tests`, and did a semantics review that found **no defect** — including the WAL question I most wanted attacked.

### The finding that matters: my stated mechanism did not reproduce

I wrote a confident three-step interleaving into the PR body **and into a permanent XML doc comment** on `TestSqlite`. The reviewer probed Microsoft.Data.Sqlite 8.0.29 directly:

- an **in-use (checked-out) connection is not disposed** by the pool-clearing call — it ran a query fine afterwards;
- 6 opener threads racing a 7th looping the call for 40 s: **232,880 cycles, zero exceptions.**

A synthetic probe failing to reproduce is weak evidence of absence, and it does not route through EF's `HistoryRepository`. But it is more than enough that I should not have called the interleaving *established* — and it would have outlived this PR as a code comment future agents inherit as fact.

**This is the second time tonight I asserted a right conclusion with an unverified mechanism** (the first was on #1608, where I wrote that node types break `PaperHomeView.vue` "because `setTimeout` returns `NodeJS.Timeout`" — that file has no `setTimeout`). Same shape: the conclusion holds, so nothing flags the reasoning. Narrowed in `32568178` to what is actually established — the call was process-global shared mutable state no test owned, the failure was real and unrelated to the diff that surfaced it, and removing the state removes the class of problem.

### Second correction: "this run leaked zero" is false

An independent full-suite run leaks **zero `-wal`/`-shm`** — the fix does what it claims there — but still leaves **~199 zero-byte `.db.migrate.lock` files and one real `.db`**. Cause: `GetDatabaseCleanupTargets` never enumerates `.migrate.lock`, where `SerializedMigratorTests` and the CLI harness both do. Pre-existing, byte-identical on `main`, unrelated to pooling.

My own 841 MB figure counted `taskdeck*.db`, and my isolated experiment used a test class that does not touch the factory — so that number is a **floor, not a total**. Filed as **#1611**; PR body and the helper's docs corrected. The commit message on `015eafb9` still says "this run leaked zero" and "13 calls" (it is 15 call sites across 13 files) — history is not rewritten; this thread and the corrected body are the record.

### Dispositions

| Finding | Disposition |
|---|---|
| MEDIUM — leak claim overstated | **Fixed** in body + docs; **#1611** filed |
| MEDIUM — mechanism asserted as established | **Fixed** in `32568178`; narrowed everywhere it appeared |
| INFO — 15 call sites, not 13 | **Fixed** in body; commit message left uncorrected on the record |
| INFO — grep returns nothing *at all*, not "nothing outside the docs" | **Fixed** |
| INFO — `Cli.Tests`/`Application.Tests` still pooled | **Fixed** — scope now stated as `Api.Tests` only. Separate processes, no cross-assembly race, and neither ever called the API |
| LOW — 9 dead `using Microsoft.Data.Sqlite;` | **Declined.** Zero build warnings, no `TreatWarningsAsErrors`; churn on 9 files for no signal |
| LOW — interpolation vs `SqliteConnectionStringBuilder` | **Declined.** Unreachable (Windows forbids `;`/`=` in usernames; `TMPDIR` is `/tmp`) and pre-existing |
| INFO — orphaned `&lt;summary&gt;` in `SerializedMigratorTests:376` | **Declined**, pre-existing and out of scope |
| INFO — E2E lane documents `Pooling=True` | **Noted, not changed.** Different process, different purpose; a doc note belongs with #1611 |

### Semantics review — clean

Worth recording, since it is the risk I was most worried about: **no test anywhere asserts a `-wal`/`-shm` file exists** (every hit in the repo is a cleanup loop). `journal_mode` is persistent in the file header, so `SerializedMigratorTests.PreSetWalJournalMode` survives a real close; `WaitForMidChainHistory`'s poll connection is held open across the loop, so pooling was never what kept it alive; and `SqlitePragmaInterceptorTests`' lock connection now provably opens its own handle instead of possibly drawing a pooled one sharing the string, which makes the contention it tests **more** faithful, not less.

### Still open

CI at the new head, and whether the suite got slower (single run, no A/B baseline). Two review rounds is the ceiling; this is round 1 of 2, and the fix changed only comment text, so no further round is owed unless CI says otherwise.

PR #1612, opened 2026-08-09
Cover queue cancellation propagation

### LINKED ISSUE #1606: [Test] Three of the four cancellation guards from #1603 have no test
## Summary

PR #1603 (#1523) added a filtered cancellation rethrow at **four** sites. Only **one** of them has a test.

Raised as a MEDIUM by an independent review lens on #1603, accepted as accurate, and tracked here rather than fixed there — the two-round review ceiling had been reached and the uncovered sites are the same four-line idiom the file already uses at `LlmQueueToProposalWorker.cs:79-82`.

## The gap

| Site | Covered? |
|---|---|
| `AutomationPlannerService.ParseInstructionAsync` | **yes** — `ParseInstruction_ShouldPropagateCancellation_WhenCallerTokenCanceled` (mutation-checked) plus the `..._WhenCancellationIsNotCallerRequested` guard rail |
| `AutomationPlannerService.ParseBatchInstructionAsync` (~`:588-592`) | no |
| `LlmQueueToProposalWorker` proposal lane (~`:366-370`) | no |
| `LlmQueueToProposalWorker` capture lane (~`:531-535`) | no |

The review confirmed by grep that **none of the 54 worker tests would fail if both worker catch blocks were deleted.** The only existing cancellation test, `ExecuteAsync_CancellationRequested_StopsGracefully` (`LlmQueueToProposalWorkerTests.cs:370`), exercises the *outer* loop's pre-existing guard and never enters `ProcessSingleItemAsync` or `ProcessCaptureTriageItemAsync` at all.

This matters more than usual because the worker lanes are exactly what the Codex P1 on #1603 was about.

## Known trap for the worker test

You **cannot** pass an already-cancelled token to `InvokeProcessBatchAsync`. `Task.Run(delegate, ct)` at `LlmQueueToProposalWorker.cs:134`/`:161` never invokes the body when the token is already cancelled, so `ProcessSingleItemAsync` is unreachable and the test would pass vacuously.

Cancel from *inside* the planner double instead — the existing `FakeAutomationPlannerService.ResultFactory` (`LlmQueueToProposalWorkerTests.cs:1112`) is a plain `Func`, so it can throw:

```csharp
var cts = new CancellationTokenSource();
var planner = new FakeAutomationPlannerService
{
    ResultFactory = _ => { cts.Cancel(); throw new OperationCanceledException(cts.Token); }
};
await Assert.ThrowsAnyAsync<OperationCanceledException>(
    () => InvokeProcessBatchAsync(worker, cts.Token));
item.Status.Should().Be(RequestStatus.Processing);
```

The `Status == Processing` assertion is the discriminating one: **without** the worker guard the catch runs `HandleFailureWithRetryAsync` → `MarkAsFailed` → `SaveChangesAsync` (the test's `FakeUnitOfWork` does not throw, unlike production EF) → backoff `[0]` skips the delay → `ResetForRetry()`, leaving the item **Pending**, not Processing.

Use `Assert.ThrowsAnyAsync`, not `ThrowsAsync` — xUnit's generic overload is exact-type and a cancelled await can surface `TaskCanceledException`.

## Acceptance criteria

- [ ] `ParseBatchInstructionAsync` has a cancellation-propagation test mirroring the single-instruction one
- [ ] Each worker lane has a test that fails if its `catch (OperationCanceledException) when (ct.IsCancellationRequested) { throw; }` block is deleted
- [ ] Each new test is **mutation-checked** — delete the guard, watch the test go red, restore. A test that passes either way is worse than no test, and the fake `UnitOfWork` makes that trap live here
- [ ] The internal-cancellation guard rail is preserved: an OCE whose origin is *not* the caller token must still map to `UnexpectedError`

## References

- Review finding: https://github.com/Chris0Jeky/Taskdeck/pull/1603#issuecomment-5222299513
- Parent: #1523 (PR #1603)
- Related seam: #1605

Chris0Jeky:
## Summary
- add batch-parser caller-cancellation propagation coverage
- cover proposal and capture worker-lane cancellation guards with Processing-state assertions

## Tests
- `dotnet test backend/tests/Taskdeck.Application.Tests/Taskdeck.Application.Tests.csproj -c Release -m:1 --filter FullyQualifiedName~AutomationPlannerServiceTests` (53 passed)
- `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1 --filter FullyQualifiedName~LlmQueueToProposalWorkerTests` (41 passed)
- new worker tests targeted together (2 passed)
- new batch test targeted (1 passed)

## Mutation evidence
- removing the proposal-lane filtered cancellation guard made its targeted test fail: no `OperationCanceledException` was thrown
- removing the capture-lane filtered cancellation guard made its targeted test fail: no `OperationCanceledException` was thrown
- both production guards were restored; production file is absent from the final diff

## Docs and risk
- docs unchanged; this is focused test coverage only
- no production behavior changes

Closes #1606

### DIFF
--- a/backend/tests/Taskdeck.Api.Tests/LlmQueueToProposalWorkerTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/LlmQueueToProposalWorkerTests.cs
@@ -233,6 +233,32 @@ public class LlmQueueToProposalWorkerTests
         item.Status.Should().Be(RequestStatus.Failed);
     }
 
+    [Fact]
+    public async Task ProcessBatch_ProposalLaneCallerCancellation_PropagatesAndLeavesItemProcessing()
+    {
+        var item = CreatePendingItem();
+        var queueRepo = new FakeLlmQueueRepository([item]);
+        using var cts = new CancellationTokenSource();
+        var planner = new FakeAutomationPlannerService
+        {
+            ResultFactory = _ =>
+            {
+                cts.Cancel();
+                throw new OperationCanceledException(cts.Token);
+            }
+        };
+        using var sp = BuildServiceProvider(queueRepo, planner);
+        var worker = CreateWorker(sp.GetRequiredService<IServiceScopeFactory>(),
+            DefaultSettings(retryBackoff: [0]));
+
+        await Assert.ThrowsAnyAsync<OperationCanceledException>(
+            () => InvokeProcessBatchAsync(worker, cts.Token));
+
+        // Discriminating assertion: without the worker cancellation guard, the generic
+        // failure path resets this item to Pending for retry.
+        item.Status.Should().Be(RequestStatus.Processing);
+    }
+
     #endregion
 
     #region Unhandled exception in planner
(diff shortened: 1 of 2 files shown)


Chris0Jeky:
## Independent review triage (round 1)

- **MEDIUM — capture retry-count assertion:** declined as non-blocking test strengthening. The deliberate guard-removal mutation already makes the targeted test fail because the expected propagated `OperationCanceledException` is absent; the PR also retains the issue-required `Processing` assertion. Adding `RetryCount == 0` would be useful redundancy, but there is no direct changed-behavior defect in this focused test-only slice.
- **MEDIUM — batch non-caller OCE guard rail:** declined as a scope extension. This PR does not alter `ParseBatchInstructionAsync` behavior, and the existing single-instruction guard-rail test remains green. A dedicated batch counterpart is reasonable future coverage but does not meet the repository's review follow-up severity bar.

No CRITICAL/HIGH finding was confirmed. The bounded review remains in round 1; no fix round is owed. Exact-head CI is still running.

PR #1613, opened 2026-08-09
[#1611] Clean up migration lock files

### LINKED ISSUE #1611: TestWebApplicationFactory never cleans .db.migrate.lock, so a full Api.Tests run leaks ~199 zero-byte files
## What

`TestWebApplicationFactory.GetDatabaseCleanupTargets` enumerates `.db`, `-wal`, `-shm` and `-journal` — and never `.db.migrate.lock`. `SerializedMigrator` creates that lock file beside the database, so every factory instantiation leaves one behind.

Measured by an adversarial review on PR #1610, snapshotting `%TEMP%` immediately before and after a full `Taskdeck.Api.Tests` run on that branch:

- **199 new zero-byte `taskdeck-api-tests-*.db.migrate.lock` files**
- **1 real `taskdeck-api-tests-*.db` (4,096 bytes)** that also survived cleanup
- **zero `-wal`/`-shm`** — that part of #1609's fix works as intended

## Why it is separate from #1609

Pre-existing and unrelated to connection pooling. `git show main:` confirms `GetDatabaseCleanupTargets` is byte-identical on `main`; #1610 does not touch it. Two other harnesses already get this right — `SerializedMigratorTests.CleanupDbFiles` includes `.migrate.lock` in its suffix list, and the CLI harness cleans it too — so the factory is the outlier, not the convention.

## Why it went unnoticed, including by me

My own leak measurement on #1609 counted files matching `taskdeck*.db`. A `.db.migrate.lock` file does match an unanchored pattern but not an anchored one, and the isolated experiment I ran covered a single test class that does not use the factory at all. So the 841 MB figure reported on #1610 is a **floor, not a total** — the `.migrate.lock` population was never in scope of that count. Worth stating plainly: the disk-pressure diagnosis there is incomplete rather than wrong.

## Acceptance criteria

- [ ] `GetDatabaseCleanupTargets` enumerates `.db.migrate.lock` alongside the existing four suffixes.
- [ ] A full `Taskdeck.Api.Tests` run leaves **zero** new `taskdeck-api-tests-*` files of any suffix in `%TEMP%`. Measure with a quiesced before/after count, not a `find -newermt` predicate — that form silently matches nothing on the Windows dev box.
- [ ] Root-cause the **one surviving `.db`** before assuming it is the same problem; it may be a distinct disposal-ordering issue in the factory, and 1-in-199 is the kind of number that gets waved away.
- [ ] Consider whether the delete loop's `catch` should stop swallowing `IOException` silently. That silence is why ~3,000 files accumulated since 2026-08-01 with nobody noticing.

## References

#1609 / PR #1610 (where this was found) · `backend/tests/Taskdeck.Api.Tests/TestWebApplicationFactory.cs` · `backend/tests/Taskdeck.Api.Tests/SerializedMigratorTests.cs` (the correct pattern)

Chris0Jeky:
## Summary
- Include `{dbPath}.migrate.lock` in API test database cleanup targets.
- Extend the target-list regression test to assert the migration-lock path.

## Verification
- `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1 --filter FullyQualifiedName~GetDatabaseCleanupTargets` — 1 passed, 0 failed, 0 skipped.
- `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1` — 2,184 passed, 4 skipped, 0 failed.
- Quiesced count-only `%TEMP%\taskdeck-api-tests-*` inventory: before `2,417`, after `2,417`; zero new matching files proven. One surviving `.db` is out of scope.
- `git diff --check` — passed.

## Scope
Only `backend/tests/Taskdeck.Api.Tests/TestWebApplicationFactory.cs` and `backend/tests/Taskdeck.Api.Tests/TestWebApplicationFactoryTests.cs` changed. No docs updated; this is a narrow test cleanup repair.

Closes #1611


Chris0Jeky:
## Independent review (post-base refresh)

Fresh exact-head review passed: base `99400d6e3b100818d19139b8fea678d72afe4154`, head `51ea220e6086d29034ec2578d9056db250e5ad2f`. The effective diff remains the intended two test-factory files; it adds the exact `.migrate.lock` cleanup target and an exact five-item regression assertion. No CRITICAL/HIGH finding or non-blocking observation was identified.

Coordinator re-ran the focused contract test at this refreshed head (1/1) and `git diff --check` passed. Hosted exact-head CI remains pending.

Chris0Jeky:
### Fresh-base review record

After merging `a7ce02c9` into this branch, the effective `origin/main...HEAD` diff remains exactly the two intended factory cleanup files. A fresh independent review found no Critical/High issue: the `.db.migrate.lock` target matches `SerializedMigrator`'s absolute-path convention, cleanup retains its explicit five-path scope and disposal ordering, and the regression test asserts the exact list.

Local proof at `f7ab2063`: `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1 --filter FullyQualifiedName~TestWebApplicationFactoryTests` passed 3/3; `git diff --check origin/main...HEAD` passed. The API test project emitted seven pre-existing nullable warnings outside this diff.

PR #1630, opened 2026-08-12
Fix standalone MCP HTTP host 500 on every authenticated request (#1602)

### LINKED ISSUE #1602: [MCP] Standalone --mcp --transport http host 500s on every authenticated request (missing CORS middleware)
## Summary

The standalone MCP HTTP host (`taskdeck --mcp --transport http`) returns **HTTP 500 on every authenticated MCP request**. It maps the MCP endpoint, which carries CORS metadata, but never registers CORS middleware, so ASP.NET Core's `EndpointMiddleware` throws before the endpoint runs.

This is on current `main` and is **not** caused by the ModelContextProtocol 2.0.0 work — I reproduced it at both 1.4.1 and 2.0.0.

## Evidence

Booting the real entry point in-process with `--mcp --transport http`, seeding a valid API key, and POSTing a well-formed `initialize`:

```
fail: Taskdeck.Api.Mcp.McpTelemetryMiddleware[3]
      MCP HTTP request failed: Method=POST DurationMs=128.5 ErrorType=InvalidOperationException
      UserId=d35bec39-319a-4f64-88d5-c9b0ff590982
      System.InvalidOperationException: Endpoint MCP Streamable HTTP | HTTP: POST /mcp/ contains
      CORS metadata, but a middleware was not found that supports CORS.
         at Microsoft.AspNetCore.Routing.EndpointMiddleware.ThrowMissingCorsMiddlewareException(Endpoint)
         at Taskdeck.Api.Middleware.ApiKeyMiddleware.InvokeAsync(...) ApiKeyMiddleware.cs:218
         at Taskdeck.Api.Middleware.McpAuthenticationRateLimitingMiddleware.InvokeAsync(...) :67
         at Taskdeck.Api.Mcp.McpTelemetryMiddleware.InvokeAsync(...) :58
         at Taskdeck.Api.Middleware.CorrelationIdMiddleware.InvokeAsync(...) :53
```

Note `UserId` is populated — **authentication succeeds**. The failure is at endpoint execution, so this is not an auth or key problem. The response body is empty, so a client sees a bare 500 with no diagnostic.

The co-hosted host is unaffected: it calls `UseCors` as part of the full web pipeline. Only the standalone branch in `Program.cs` (the `--mcp --transport http` path, roughly lines 62 to 160) builds a minimal pipeline that omits it.

## Why no existing test catches it

`StandaloneMcpHostFilteringTests` has five tests and none of them can fail on this:

- `..._SeparatorOnlyAllowedHosts_FailsClosedToLoopbackFiltering` asserts 400 (hostile Host) and 401 (missing key) — both short-circuit **before** the endpoint.
- `..._ForwardedHeaders_*` and `..._OutOfRangeConcurrency_*` likewise never reach endpoint execution.
- `..._PerKeyBudget_RejectsOverQuotaValidKey` posts `{}` with a valid key and asserts only `NotBe(Unauthorized)` and `NotBe(TooManyRequests)`. **A 500 satisfies both.**

So the suite proves the host boots and that its middleware rejects the right things, but never that it can successfully serve a single MCP request.

## Impact

The standalone MCP HTTP server is non-functional for real MCP traffic. This matters for the MCP packaging work (#1154) and multi-user-safe identity (#1309), and it is a surface the revival plan positions as the "developer second act".

Severity depends on whether anyone runs `--mcp --transport http` today; the stdio transport (the Claude Code / Cursor path) is unaffected.

## Suggested fix — needs a decision, not just a line

The mechanical fix is to register CORS in the standalone pipeline (`AddCors` plus `UseCors`). But the right *policy* is a real decision for a loopback-bound MCP server:

1. **Register CORS with no permissive default policy.** The middleware satisfies the endpoint requirement and adds no cross-origin headers, so browsers still block cross-origin calls. Smallest, most conservative.
2. **Register an explicit, restrictive named policy** and apply it to the MCP endpoint.
3. **Strip the CORS metadata** from the mapped endpoint instead of adding middleware, if the MCP library allows it — arguably the most honest for a server with no browser clients.

Option 1 is probably right, but this is a security-adjacent surface on a host that binds loopback by default and authenticates with API keys, so it deserves an explicit choice rather than the first thing that makes the error go away.

## Acceptance criteria

- [ ] Decision recorded among the options above
- [ ] `taskdeck --mcp --transport http` serves a successful `initialize` with a valid API key
- [ ] A test in `StandaloneMcpHostFilteringTests` asserts a **successful** MCP request end to end — status 200 and a non-empty `Mcp-Session-Id` — so this class of failure cannot recur silently
- [ ] `..._PerKeyBudget_RejectsOverQuotaValidKey` tightened so its first request asserts success rather than merely `NotBe(401)`/`NotBe(429)`
- [ ] Confirm no cross-origin browser access is unintentionally enabled

## Repro

Add a test to `StandaloneMcpHostFilteringTests` modelled on `..._PerKeyBudget_RejectsOverQuotaValidKey`: same host boot and `SeedApiKeyAsync`, budgets raised to 1000, then POST `/mcp` with `Authorization: Bearer <key>`, `Accept: application/json` and `text/event-stream`, and a JSON-RPC `initialize` body (`protocolVersion` `2025-11-25`). Assert 200 and `Mcp-Session-Id`. It currently fails with the 500 above at both 1.4.1 and 2.0.0.

Found while validating the ModelContextProtocol 2.0.0 bump (#1598) on 2026-08-07. Deliberately not fixed there to keep that PR focused.

Chris0Jeky:
﻿## Problem

`taskdeck --mcp --transport http` returned **HTTP 500 with an empty body on every authenticated MCP request**. Authentication succeeded (the failing log line carries a resolved `UserId`); the request died at endpoint execution.

Where the CORS metadata comes from — the first thing #1602 asked to establish — is **our own code, not the MCP SDK**:

`backend/src/Taskdeck.Api/Extensions/McpEndpointMapping.cs:22`
```csharp
mcpEndpoint.WithMetadata(new DisableCorsAttribute());
```

`DisableCorsAttribute` implements `IDisableCorsAttribute : ICorsMetadata`, and that is exactly what `EndpointMiddleware` checks: an endpoint carrying `ICorsMetadata` will not be executed unless the CORS middleware ran first and set its marker in `HttpContext.Items`. The standalone branch of `Program.cs` never registered CORS, so `ThrowMissingCorsMiddlewareException` fired before the endpoint. The co-hosted pipeline calls `UseCors("AllowFrontend")` (`PipelineConfiguration.cs:64`) and was unaffected.

## Fix

Option 1 from the issue. Two lines in the standalone `--mcp --transport http` branch only:

- `mcpHttpBuilder.Services.AddCors();` — registers the CORS services with an **empty policy map**. Deliberately *not* `AddTaskdeckCors`, which would bring the credentialed `AllowFrontend` policy.
- `mcpHttpApp.UseCors();` — parameterless, so it resolves the default policy name, finds nothing, and passes through. Placed to mirror the co-hosted order (forwarded headers → CORS → correlation ID).

No default policy exists to resolve, so `CorsMiddleware` never emits an `Access-Control-*` header; the endpoint's own `DisableCorsAttribute` suppresses CORS handling for it regardless. The middleware exists purely to satisfy the endpoint-metadata contract.

### Why not option 3 (strip the metadata in standalone mode)

I investigated it as invited, and rejected it as *not* strictly cleaner:

1. `MapTaskdeckMcpEndpoint` exists specifically to map the endpoint **identically** for both hosts. Parameterizing it to diverge weakens the one thing it guarantees.
2. `DisableCorsAttribute` is a defence-in-depth latch: if the standalone pipeline ever gains a CORS policy, the metadata keeps the MCP endpoint out of it. Dropping the attribute in standalone removes that latch permanently, and silently.

The co-hosted pipeline's CORS behaviour is untouched.

## Named assumption (ratified)

> Assumption: the standalone MCP host registers CORS middleware with a deny-by-default policy (no allowed origins) purely to satisfy the endpoint's CORS metadata; browser-origin MCP clients remain unsupported and any relaxation is a future explicit config decision. Reversible by replacing the policy. Reason: restores endpoint execution with zero loosening of posture.

Implemented exactly as stated — no adaptation needed.

## Verification

All runs foreground, `-c Release -m:1`, on this box.

| Command | Result |
| --- | --- |
| `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1 --filter "FullyQualifiedName~StandaloneMcpHost"` | **25 passed / 0 failed**, 5s |
| `... --filter "FullyQualifiedName~Mcp\|FullyQualifiedName~Cors"` | **173 passed / 0 failed**, 1m21s |
| `dotnet test backend/tests/Taskdeck.Api.Tests/Taskdeck.Api.Tests.csproj -c Release -m:1` (whole project) | **2206 passed / 0 failed / 4 skipped**, 7m02s |

The 4 skips are pre-existing `LlmQuota*ConcurrencyTests` skips, unrelated to this change.

### Mutation check — both directions

The test code was **identical** in both runs; only the `Program.cs` fix differed.

- **Without the fix** (tests added, `Program.cs` untouched): `Failed: 2, Passed: 23, Total: 25`. Both failures were exactly the documented defect —
  `Expected response.StatusCode to be HttpStatusCode.OK {value: 200} ... body was: , but found HttpStatusCode.InternalServerError {value: 500}` — note the **empty body**, matching the issue's report.
- **With the fix**: `Failed: 0, Passed: 25, Total: 25`.

So the new assertions fail for the right reason and pass only because of the change.

### Test changes (all in `StandaloneMcpHostFilteringTests.cs`)

- **New** `StandaloneMcpHttpHost_ValidKey_ServesSuccessfulInitialize` — boots the real entry point, seeds a valid API key, POSTs a well-formed JSON-RPC `initialize` (`protocolVersion` `2025-11-25`, `Accept: application/json` + `text/event-stream`) and asserts **200**, a non-empty `Mcp-Session-Id`, and an initialize *result* in the body — not merely not-401.
- **Cross-origin denial (issue AC 5)**: the same request is re-sent with `Origin: https://evil.example` — the branch where `CorsMiddleware` actually evaluates a policy — asserting the response carries **no** `Access-Control-Allow-Origin` and **no** `Access-Control-Allow-Credentials`. A permissive policy would fail this. Cross-origin browser access is therefore proven closed by test, not just by comment.
- **Tightened** `StandaloneMcpHttpHost_PerKeyBudget_RejectsOverQuotaValidKey` (issue AC 4): its first request is now a real `initialize` asserted to be **200** instead of `NotBe(401)`/`NotBe(429)` — the pair a 500 satisfied.
- The two duplicated host-boot blocks are folded into one `RunSeededApiKeyHostAsync(plaintextKey, perKeyPermitLimit, assertions)` helper, so the file has two boot harnesses rather than three.

## Residual risk

- **Preflight `OPTIONS` is not covered.** *(Corrected after review — the original text here inverted the order.)* `UseCors()` runs **before** `ApiKeyMiddleware` in the standalone pipeline (Program.cs:184 vs :208, mirroring the co-hosted host where `UseCors` at PipelineConfiguration.cs:64 is likewise pre-auth). On the `DisableCorsAttribute` endpoint, `CorsMiddleware` short-circuits a browser preflight (OPTIONS + `Access-Control-Request-Method`) with an empty **204 and no `Access-Control-Allow-Origin`**, so the browser's preflight fails and no cross-origin request proceeds — the same posture as the co-hosted host. Consequences: an unauthenticated preflight is answered 204 (not 401) without touching `ApiKeyMiddleware`, the pre-auth failure limiter, telemetry, or correlation — negligible on a loopback host returning an empty response — and preflight behaviour is asserted nowhere (tracked follow-up).
- The `Mcp-Session-Id` assertion depends on `Stateless = false`, pinned just above the fix in `Program.cs`. If that pin is ever flipped, this test fails loudly — which is the desired coupling, but worth knowing it is deliberate.
- Verified in-process via the real assembly entry point, as the existing tests in this class do. **Not** verified against a separately launched `taskdeck.exe --mcp --transport http` process or a real MCP client (Claude Code / Cursor over HTTP).
- The stdio transport is untouched and was not re-verified.
- No docs were updated: `docs/STATUS.md`, `docs/IMPLEMENTATION_MASTERPLAN.md`, `docs/TESTING_GUIDE.md` and `OUTSTANDING_TASKS.md` are outside this task's ownership. A STATUS line noting the standalone HTTP host is now functional may be worth a follow-up.

## Worktree teardown

Branch is pushed; the worktree is spent and ready for the coordinator to `git worktree remove` (plain, never `--force`). `git status --porcelain --ignored` showed no work product to rescue — the only non-build ignored file is `.claude/settings.local.json`, a per-worktree permissions cache. Nothing was copied out.

Closes #1602

### DIFF
--- a/backend/src/Taskdeck.Api/Program.cs
+++ b/backend/src/Taskdeck.Api/Program.cs
@@ -125,6 +125,17 @@ if (args.Contains("--mcp"))
         // MCP telemetry (operation logger, etc.).
         mcpHttpBuilder.Services.AddMcpTelemetry();
 
+        // CORS services with NO policies registered at all -- deliberately not AddTaskdeckCors (#1602).
+        // MapTaskdeckMcpEndpoint stamps the endpoint with DisableCorsAttribute, which is ICorsMetadata,
+        // and ASP.NET Core's EndpointMiddleware refuses to execute an endpoint carrying CORS metadata
+        // unless the CORS middleware ran first -- so without this pair (AddCors here, UseCors below) the
+        // standalone host threw InvalidOperationException and returned a bare 500 on EVERY authenticated
+        // request. Registering the services with an empty policy map satisfies that contract and adds no
+        // cross-origin capability: there is no default policy to resolve, so CorsMiddleware never emits
+        // an Access-Control-* header, and the endpoint's DisableCorsAttribute suppresses CORS handling
+        // for it regardless. Browser-origin MCP clients stay unsupported here by construction.
+        mcpHttpBuilder.Services.AddCors();
+
         // MCP server: HTTP transport + all resources and tools.
         // Stateless is pinned explicitly rather than left to the library default:
         // ModelContextProtocol 2.0.0 flipped that default from false to true, which drops
@@ -163,6 +174,14 @@ if (args.Contains("--mcp"))
         {
             mcpHttpApp.UseForwardedHeaders(mcpForwardedHeadersOptions);
         }
+
+        // CORS middleware with no named policy: required so EndpointMiddleware will execute the MCP
+        // endpoint, which carries DisableCorsAttribute (ICorsMetadata) -- see AddCors above (#1602).
+        // Deny-by-default: no policy is registered, so this emits no Access-Control-* headers and
+        // grants no cross-origin access; it exists purely to satisfy the endpoint's metadata contract.
+        // Positioned to mirror the co-hosted pipeline (forwarded headers -> CORS -> correlation ID),
+        // and after the auto-inserted UseRouting so an endpoint is selected when it runs.
+        mcpHttpApp.UseCors();
 
         // Correlation ID propagation: honours client X-Request-Id header.
         mcpHttpApp.UseMiddleware<Taskdeck.Api.Middleware.CorrelationIdMiddleware>();
--- a/backend/tests/Taskdeck.Api.Tests/StandaloneMcpHostFilteringTests.cs
+++ b/backend/tests/Taskdeck.Api.Tests/StandaloneMcpHostFilteringTests.cs
@@ -2,6 +2,7 @@ using System.Net;
 using System.Net.Http.Headers;
 using System.Net.Sockets;
 using System.Security.Cryptography;
+using System.Text;
 using FluentAssertions;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Data.Sqlite;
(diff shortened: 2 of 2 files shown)


Chris0Jeky:
## Independent adversarial review — verdict SHIP; claim-accuracy fixes applied at dd65140c

A fresh-context review (security-posture, middleware-order, test-discrimination, base-drift, and claims-accuracy lenses) ran against head `d6282fd8`. The fix itself was found mechanically correct, minimal, and complete (both MCP hosts covered; posture not loosened). Dispositions:

**Fixed:**
- **F2 (MEDIUM, permanent comment)** — the cross-origin test comment claimed the empty policy map is what the assertion guards; actually `CorsMiddleware` short-circuits on the endpoint's `IDisableCorsAttribute` before resolving ANY policy, so a permissive policy would pass the assertion too (double-refuted: framework source + this repo's own `McpHttpTransportApiKeyTests:304-319` allowed-origin test). Comment reworded in `dd65140c` to name the real latch. Comment-only diff; hosted CI at the new head is the compile proof.
- **F1 (MEDIUM, PR body)** — the residual-risk paragraph inverted the middleware order ("CORS after ApiKeyMiddleware, preflight 401'd"); actually `UseCors()` is pre-auth (Program.cs:184 vs :208) and a preflight gets an empty pre-auth 204, same as the co-hosted host. Body corrected, marked as a post-review correction; `Closes #1602` verified intact after the edit.

**Tracked as #1637 (not fix-cascaded):** F3 preflight OPTIONS coverage, F4 invariant-culture env formatting (prior history #1400), F5 boot-harness count nit.

**Declined with reason:** F6 (no ADR) — the deny-by-default decision, the rejected alternative (stripping the endpoint metadata in standalone), and the reversibility path are recorded verbatim in this PR body, the code comments, and issue #1602; a 2-statement middleware registration that changes no policy surface doesn't meet the ADR bar (hard-to-reverse / project-wide constraint).

**Refuted candidates (recorded):** config-injected permissive policy (CorsOptions never bound from IConfiguration; `AddTaskdeckCors`'s only call site is the co-hosted branch); standalone inheriting `AllowFrontend` (separate ServiceCollection); UseCors-before-routing 500 (WebApplication auto-injects UseRouting; proven by #1602's own stack trace); missing-Origin marker skip (marker set before the Origin check); MCP 2.1.0 base drift (release notes are client-side; API Integration green on both OSes at the merge).

Merge gate remaining: green CI at `dd65140c` + aging.

Issue #1607, opened 2026-08-07
Burn down the frontend spec typecheck quarantine (40 files, ~391 errors — was 64/415)


Chris0Jeky:
## What

`#1468` put the frontend spec tree under `vue-tsc` behind a **quarantining** project
(`frontend/taskdeck-web/tsconfig.vitest.json`). 222 of 286 spec files are gated from now on; the 64
that carried pre-existing type errors are listed in that file's `exclude` array. This issue tracks
burning that list down to empty.

**The list may only shrink.** Delete an entry once its file compiles; never add one to turn a red
build green. A new spec file is checked by default, because it is not in the list — that property is
the whole point of the quarantine and must not be eroded.

## Current baseline (measured 2026-08-23, on `main` `a43df082c`)

**40 files remain quarantined, holding ~391 errors** — down from the 64 / 415 this issue was filed
against. `frontend/taskdeck-web/tsconfig.vitest.json` carries exactly 40 entries in `exclude`, all
spec files; gated spec files went 220 → 244.

**Shrink-only confirmed.** Comparing the 2026-08-07 per-file table below against the live `exclude`
array: 24 entries cleared, **zero added**. The 24 that cleared are precisely the 24 one-error files,
exactly as the suggested order predicted:

`components/AppShell.paperVariant` · `components/AppShell` · `components/BoardCanvas.coverage` ·
`components/BoardSettingsModal` · `components/inbox/inboxUtils` · `components/paper/useInkBleed` ·
`composables/useEscapeToClose` · `composables/useShortcutContext` · `composables/useVirtualList` ·
`store/auditStore` · `store/board/boardStoreHelpers` · `store/captureStore.integration` ·
`store/savedViewStore` · `views/ActivityView` · `views/AutomationQueueView` · `views/BoardView` ·
`views/MetricsView.coverage` · `views/NotificationInboxView` · `views/ReviewView.coverage` ·
`views/ReviewView` · `views/paper/PaperBoardView` · `views/paper/inbox/PaperTriageTable` ·
`views/paper/review/PaperReviewView` · `views/paper/review/ReviewProvenance`

**Scope correction the body never absorbed** (raised in the 2026-08-07 comment): emptying this
`exclude` array is **not** the same as "the frontend specs are type-checked". There is a *second*
unchecked spec tree — 18 frontend-root `tests/` specs that no tsconfig includes at all. That tree
needs its own project, and it is out of scope for this issue's acceptance criteria as written.

[...]

PR #1634, opened 2026-08-12
Burn down the frontend spec typecheck quarantine: 64 -> 40 files

Chris0Jeky:
## Problem

`tsconfig.vitest.json` (ADR-0049, #1608) type-checks the frontend spec tree but quarantines
64 files in its `exclude` array that carried 415 pre-existing type errors. #1607 tracks burning
that list down. This is the first chunk.

## Chunk selection

The **24 files that carry exactly one error each** — the issue's own suggested starting order
("the 24 one-error files first (cheap, proves the workflow)"). They are mechanical rather than
structural, and they span 7 distinct error classes, so this batch also establishes a fix pattern
for each class that later chunks can reuse. The 7 heavy hitters (227 errors, 55%) are deliberately
left for dedicated PRs.

Measured before starting: removing exactly these 24 from `exclude` produced exactly 24 errors,
one per file — the issue's per-file table reproduces precisely.

### Fixes by class, one commit each

| Class | Files | Fix |
|---|---|---|
| TS6133 unused binding | 3 | drop the unused local (keeping the call it was made for); `_`-prefix the unused *parameter* |
| TS2339 `.value` on a union | 4 | bind `options.count` to a local so the `typeof` narrowing survives into the closure |
| TS2339 `.exists()` | 3 | `get`/`getComponent` are typed `Omit<Wrapper, "exists">`; switched to `find`/`findComponent` so the explicit assertion survives |
| fixture literals (TS2322/2739/2353/2345) | 8 | make the fixture a valid domain object; type two payload builders instead of `Partial<Record<string, unknown>>` |
| misc one-offs | 6 | missing `afterEach` import; type-only export aliased; mock signature; definite-assignment; `VirtualItem`; `node:fs` -> Vite `?raw` |

Two judgement calls worth a reviewer's eye:

- **`PaperTriageTable`** asserts on `status: 'Triage Failed'`, deliberately outside the
  `CaptureStatus` union — it exercises `statusTone()`'s failed-before-triage ordering for an
  unrecognised server status. I kept the value behind a documented `as unknown as CaptureStatusValue`
  rather than substituting a valid status, which would have silently deleted what the test covers.
- **`AppShell.paperVariant`** imported `readFileSync` from `node:fs` (TS2591). Adding `"node"` to
  `types` is forbidden — it breaks `PaperHomeView.vue`. Replaced with Vite's `?raw` loader, typed by
  `vite/client`, which is already in this project's `types`. Same string, same assertions, and it now
  resolves relative to the spec instead of the process CWD.

No production source under `src/` was touched. No test was weakened: no assertion was deleted and no
`as any` blanket was added.

## Verification

From `frontend/taskdeck-web`, after `npm ci`:

- `npm run typecheck` — **exits 0**. Also re-run as `npx vue-tsc -b --force` so incremental
  build-info could not mask a stale result: also clean.
- `npx vitest --run --maxWorkers=2 <the 24 paths>` — **24 files, 409 tests, all passing**
  (162 + 247 across two batches).
- **Behaviour-identical, measured not assumed.** I restored the base (`78be65c6`) spec content into
  the worktree and re-ran the same 24 files: **24 files, 409 tests, all passing — identical counts**,
  then restored the fixed versions. So the fixes changed types, not behaviour.
- Baseline control: `npm run typecheck` was green *before* any edit, and removing the 24 files from
  `exclude` produced exactly 24 errors.

### Constraints (all four on #1607 hold, provably)

`git diff` on `tsconfig.vitest.json` touches **no** `compilerOptions` key, **no** `types` entry and
**no** `include` entry — only the comment block and the `exclude` list. So: no `"node"`, no
`"vitest/globals"`, `noUnusedLocals`/`noUnusedParameters`/`erasableSyntaxOnly` all still on, and
`src/**/*.d.ts` still in `include`.

## Counts

| | Before | After |
|---|---|---|
| Quarantined files | 64 | **40** |
| Errors in quarantine | 415 | **391** |
| Gated spec files (of the 302 Vitest runs) | 220 | **244** |
| Not covered | 82 (64 + 18) | **58 (40 + 18)** |

**Residual quarantine: 40 files / 391 errors.** The burn-down continues — the 7 heavy hitters
(ReviewKeymap 48, useStarterPackCatalog 42, useStarterPackImport 36, useReviewActions 33,
PaperBranding 25, scenarioSchema 24, boardStore 19) are the natural next PRs, one each.

The config's own header comment is updated to these numbers; it would otherwise have been left
stating a false count. Canonical docs (`STATUS.md`, `TESTING_GUIDE.md`, ADR-0049) are deliberately
untouched here — the coordinator reconciles those, and the acceptance criterion for the testing-guide
note is "when the list empties", which it has not.

Refs #1607

### DIFF
--- a/frontend/taskdeck-web/src/tests/components/AppShell.paperVariant.spec.ts
+++ b/frontend/taskdeck-web/src/tests/components/AppShell.paperVariant.spec.ts
@@ -1,8 +1,11 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { mount } from '@vue/test-utils'
 import { reactive } from 'vue'
-import { readFileSync } from 'node:fs'
 import AppShell from '../../components/shell/AppShell.vue'
+// Read through Vite's `?raw` loader rather than node:fs — this project deliberately excludes
+// node types (adding them breaks production source), and `?raw` also resolves relative to this
+// file instead of the process CWD.
+import appShellSource from '../../components/shell/AppShell.vue?raw'
 
 const mockRouter = {
   push: vi.fn(),
@@ -230,10 +233,7 @@ describe('AppShell — paper variant routing', () => {
   })
 
   it('keeps Paper phone content padded above the fixed bottom bar', () => {
-    const source = readFileSync(
-      'src/components/shell/AppShell.vue',
-      'utf8',
-    )
+    const source = appShellSource
 
     expect(source).toContain('.td-shell--paper-phone .td-content')
     expect(source).toContain('56px + var(--paper-safe-bottom')
--- a/frontend/taskdeck-web/src/tests/components/AppShell.spec.ts
+++ b/frontend/taskdeck-web/src/tests/components/AppShell.spec.ts
@@ -28,7 +28,7 @@ const mockFeatureFlags = {
     newAutomation: true,
     newArchive: true,
   },
-  isEnabled: vi.fn(() => true),
+  isEnabled: vi.fn((_flag: keyof FeatureFlags) => true),
 }
 
 const mockWorkspace = reactive({
@@ -127,7 +127,7 @@ describe('AppShell workspace navigation and command palette', () => {
     mockWorkspace.homeLoading = false
     mockWorkspace.preferenceLoading = false
     mockWorkspace.preferencesHydrated = false
-    mockFeatureFlags.isEnabled = vi.fn(() => true)
+    mockFeatureFlags.isEnabled = vi.fn((_flag: keyof FeatureFlags) => true)
   })
 
   afterEach(() => {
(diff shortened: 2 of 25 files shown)


Chris0Jeky:
## Independent fresh-context review — verdict SHIP, zero findings

A test-integrity-focused review ran against head `7c7df2e3`:

- **No test's discriminating power was reduced**: zero assertions deleted, zero tests removed, zero suppressions added (mechanical `it`/`describe`/`expect` count comparison base-vs-head across all 24 files: 0 deltas; diff scan for `any`/`@ts-ignore`/`@ts-expect-error`/`eslint-disable`/`.skip`/`.only`: 0 hits).
- **Both flagged judgment calls verified sound.** The `as unknown as CaptureStatusValue` in PaperTriageTable keeps the byte-identical out-of-contract runtime value that discriminates `statusTone()`'s `failed`-before-`triag` ordering — substituting a valid status would have silently deleted that coverage. The `node:fs` → `?raw` swap was verified at the plugin level (`plugin-vue` bails on `query.raw`, returning untransformed SFC text including the asserted `<style>` rule) and removes a latent CWD dependency.
- **Constraint check programmatic**: non-exclude tsconfig config byte-identical; `exclude` 64→40 with 0 additions; the 24 removed entries are set-equal to the 24 edited files — nothing un-quarantined without being fixed.
- Twelve test-weakening candidates examined and refuted individually (`get`→`find` is strictly more correct; the `expectedVersion: '7'` fixture now matches the backend's `string?` DTO; the added fixture fields are never rendered/asserted; etc.).

**Non-blocking note:** Workflow Lint (CI Extended, the declared optional lane) failed on a curl 503 downloading the actionlint toolchain — infra flake, no repo file read; re-run requested by the coordinator.

Merge gate remaining: API Integration (windows-latest) completing green.

Issue #1625, opened 2026-08-12
[Security] .codex/config.toml prepends gitignored .runtime-codex/bin to PATH (writable-ignored-first tool shimming)


Chris0Jeky:
Found during the 2026-08-12 re-triage of PR #1557 (pre-existing on main, NOT introduced by that PR).

## Problem

`.codex/config.toml:37` prepends `C:\Users\jekyt\source\Taskdeck\.runtime-codex\bin` to PATH ahead of `C:\Program Files\Git\cmd` and `C:\Program Files\GitHub CLI`, and `.gitignore:63` (`/.runtime-codex/`) makes that directory invisible to `git status`. Any process that runs during a Codex session can drop an executable named `gh`/`git`/`jq`/etc. into that directory and have it selected by every later bare-name invocation, while the worktree still reports clean. This is the enabling condition for finding 1 of the PR #1557 park (evidence-collector tool shimming), but it applies to every script and skill run under that config, not just the collector.

## Suggested direction

Either stop prepending a writable, gitignored, repo-local directory to PATH (append instead, or point at a read-only location outside the worktree), or have security-sensitive scripts resolve their tools to absolute trusted paths (the planned #1557 successor does this for the evidence collector — that neutralizes the collector consumer only, not the config hazard itself).

Refs #1547, #1501 (externally pinned launcher — same trust-boundary family).

Chris0Jeky:
PR #1646 was parked after two bounded review rounds. Successor issue #1651 owns the trusted-only PATH repair with explicit resolution and forgery tests; keep this parent open until that successor is proven and merged.

Issue #1544, opened 2026-07-29
[Harness] Remove Taskdeck-local review-policy copies and route to canon


Chris0Jeky:
## Problem

Taskdeck still has active agent manuals and skills that restate estate-wide review/merge policy instead of routing to its canonical source. The drift is recurring:

- PR #1531 attempted a one-home consolidation but merged with contradictory local review rules, requiring #1532 / PR #1534.
- During PR #1534's current-main conflict resolution, both overnight manuals retained “There is no aging requirement” while the canonical policy requires a post-push eligibility floor. Fresh exact-head review caught the bypass before landing.
- The same surfaces have separately drifted on manual `@codex` triggers, reviewer counts, tier rows, fix-loop behavior, and merge-action ownership.

This repeated friction now costs more than the underlying edits. This issue is Taskdeck's concrete slice of agent-harness #101 and a bounded child of #1291 / #1269. It does not create another policy home or another gate.

## Goal

Delete Taskdeck-local copies of estate-wide review count, severity, convergence, post-push eligibility, reviewer invocation, and merge-authority policy. Keep only repository-specific lenses, mechanics, and proving checks, with concise pointers to the canonical global pipeline.

## Scope

Active surfaces include, at minimum:

- `docs/agentic/OVERNIGHT_LOOP.{claude,codex,fable}.md`
- `.claude/skills/{pre-merge-gate,issue-to-pr,taskdeck-pr-review-loop,taskdeck-issue-batch-orchestrator,taskdeck-worktree-issue-worker}/SKILL.md`
- live `.codex/skills/` mirrors while #1291 has not retired them
- hook/config routing strings and documentation that claims the one-home boundary

Historical delivery records are evidence, not executable instruction, and should not be rewritten merely to remove old wording.

## Acceptance criteria

- [ ] Inventory every active Taskdeck instruction that states a review count, severity threshold, fix/re-review loop, bot invocation rule, post-push wait/aging rule, tier row, or merge-authorization rule.
- [ ] Delete those local policy statements and replace only the necessary entry/exit points with canonical pointers; retain Taskdeck-specific risk lenses, exact-head/base evidence mechanics, local proving checks, dependency-safe ordering, and explicit stop boundaries.
- [ ] Use one-time before/after `rg` evidence to show the obsolete copies are gone. Do **not** add a permanent scanner, meta-gate, or doc-consistency gate whose subject is other gates.
[...]

Issue #1543, opened 2026-07-29
[Gate] Derive the pre-merge secrets verdict from real scan evidence


Chris0Jeky:
## Problem

PR #1534's `.claude/skills/pre-merge-gate/SKILL.md` report template unconditionally renders `Secrets scan: CLEAN`, but the skill never binds that verdict to a local scan or a named hosted check. A readiness report can therefore attest to a security check that was not read.

Tracked from the current review thread on PR #1534. This is a confirmed P2 / MEDIUM reporting defect, not a demonstrated bypass of hosted Gitleaks, so it is a bounded follow-up rather than another fix in #1534's HIGH-only repair batch.

## Scope

Make the report derive its secrets verdict from explicit evidence. Prefer the existing exact-head hosted Gitleaks check; if that evidence is missing, pending, or ambiguous, report `NOT VERIFIED` or block rather than printing `CLEAN`.

## Acceptance criteria

- [ ] The pre-merge gate names the exact local command or hosted check that supplies secrets-scan evidence.
- [ ] `CLEAN` is emitted only when that exact-head evidence is terminal-success.
- [ ] Missing, pending, stale-head, or failed scan evidence cannot produce `CLEAN`.
- [ ] A targeted contract assertion rejects an unconditional `Secrets scan: CLEAN` template.
- [ ] Hook smoke and all documentation/governance checks pass.

## Relationships

Follow-up to #1532 and PR #1534. This does not replace the repository's hosted Gitleaks workflow or the branch-protection tracker #1173.

Chris0Jeky:
[Codex lane claim]

- lane: testing-continuation
- base: `main` `88bb3b4498e975eda17f64dfbd4a975373aa5890`
- owned paths: `.claude/skills/pre-merge-gate/SKILL.md`, a focused pre-merge-gate contract validator/test under `scripts/`, and the existing docs-governance invocation seam
- scope: derive `CLEAN` only from the exact `pr_head_oid` latest `Gitleaks Scan` check run being `completed/success`; missing, ambiguous, pending, cancelled, skipped, stale-head, or failed evidence reports `NOT VERIFIED`
- non-overlap: #1547 retains skill argument transport, review-thread GraphQL completeness, and final OID re-read/atomicity
- method: red contract first, then minimal skill/guard implementation; draft PR until exact-head CI and Codex review complete

Issue #1532, opened 2026-07-27
[Bug] Repair merged review-doctrine gate regressions from PR #1531


Chris0Jeky:
## Summary

PR #1531 merged at `0db4538838874160d7de9f3c3b2ec1cc061b1352` before its exact-head review findings were repaired. The merged review-doctrine sweep therefore leaves two merge-gate defects and six validated doctrine/routing regressions on `main`.

Source evidence: PR #1531, exact-head adversarial review [4787433305](https://github.com/Chris0Jeky/Taskdeck/pull/1531#pullrequestreview-4787433305), and its seven unresolved inline threads.

## Reproduction Steps

1. Inspect the merged PR #1531 head and review 4787433305.
2. Follow `.claude/skills/issue-to-pr` or the Codex batch orchestrator as written.
3. Observe that the former can invoke a merge-capable workflow despite promising an unmerged handoff, while the latter makes Taskdeck's required T3 independent pass optional.
4. Inspect the remaining live wrappers/manual/hook and observe superseded local review-count, fix-loop, aging, risk-routing, and disposition copies.

## Expected Behavior

Taskdeck entrypoints route to the estate-wide `review-and-ship` workflow without changing their own external-state contract. Taskdeck's declared T3 independent review remains mandatory. Local files retain only Taskdeck-specific lenses and operational mechanics, with no competing round/severity/aging doctrine.

### HIGH

- `.claude/skills/issue-to-pr/SKILL.md` invokes the full merge-capable pipeline while promising to leave an open, unmerged PR.
- `.codex/skills/taskdeck-issue-batch-orchestrator/SKILL.md` makes the first independent T3 review discretionary.

### MEDIUM

- `docs/agentic/OVERNIGHT_LOOP.fable.md` still hard-codes two reviewers and a 30–60 minute aging window.
- The Claude PR-review wrapper omits the sensitive-surface flag retained by the still-live Codex mirror.
- `.claude/skills/pre-merge-gate/SKILL.md` still creates unconditional fix/self-review loops.
- `scripts/agent_hooks/post_tool_use.py` restates round and finding disposition; its smoke assertion does not pin a policy-neutral message.
- Worktree-worker, batch, and OpenAI skill entrypoints retain standalone fix/re-review instructions outside the canonical workflow.
- `docs/STATUS.md` and `docs/IMPLEMENTATION_MASTERPLAN.md` do not record the active transition or `#1269` narrowing.

## Scope / Impact

[...]

PR #1638, opened 2026-08-12
Harden pre-merge evidence sessions (successor)

### LINKED ISSUE #1547: [Gate] Make pre-merge evidence atomic across argument, feedback, and OID changes
## Problem

PR #1546 correctly makes `.claude/skills/pre-merge-gate` an evidence collector rather than a local merge authority, but its final-head Codex review found three bounded completeness defects in that evidence packet:

1. Claude Code supports the literal `$ARGUMENTS` placeholder, not Bash `${ARGUMENTS:-}` substitution. An explicit PR argument can therefore be dropped before the exact-head lookup, mixing current-branch identity with commands that later target the requested PR.
2. The listed REST/comment commands do not expose GraphQL review-thread `isResolved`, so the packet cannot substantiate its promised thread-resolution state.
3. Head/base OIDs are proved only before potentially long checks. A push or base advance during those checks can combine old local evidence with current PR checks unless the OIDs are re-read before reporting.

These were classified as non-CRITICAL final-round follow-ups under the bounded review pipeline; PR #1546 is not widened for another fix cascade.

## Scope

Harden the existing evidence collector without recreating review, severity, or merge policy locally.

## Acceptance criteria

- [ ] Use a documented Claude Code skill-argument form (`$ARGUMENTS`, indexed, or named arguments) that handles both an explicit PR number and the current-branch default.
- [ ] Prove with focused canaries that an explicit PR selects that PR, an omitted argument selects only the current branch's PR, and a wrong checkout/PR pairing fails before checks.
- [ ] Query review threads through cursor-complete GraphQL and return each thread's resolution state and content alongside top-level comments and review-summary bodies.
- [ ] Immediately before reporting, re-query live `headRefOid` and `baseRefOid` and compare them with the initial/local identities; any change makes the packet incomplete and invalidates affected evidence.
- [ ] Keep the skill evidence-only: it returns content and exact identities to `review-and-ship` and adds no Taskdeck-local severity, convergence, reviewer-count, or merge-authority rule.
- [ ] Run affected skill validation, Bash syntax/canaries, hook smoke, docs/governance gates, diff check, DCO, exact-head CI, and one bounded independent review.

## Relationships

Follow-up to #1544 and PR #1546. Related to #1543's separate requirement that the secrets verdict be derived from real scan evidence.

Chris0Jeky:
## Disposition

Successor to **#1557**, which is parked at the two-round review ceiling with five verified HIGH
false-evidence findings. This branch is cut from #1557's head `5c6e4bd1` (its full history is
preserved as ancestors, not rewritten) and fixes all five together, each with a runtime canary.
Do not merge #1557; the coordinator closes it against this PR.

Re-verified read-only at `5c6e4bd1` on 2026-08-12 (park comment on #1557): none of the five was
fixed there, and the two nearest canaries were `rg` presence checks that passed while the defects
were live.

## The five findings

| # | Finding (thread) | Fix at this head | Canary evidence |
| --- | --- | --- | --- |
| 1 | Evidence tools resolved from PATH while `.codex/config.toml` prepends the gitignored, writable `.runtime-codex/bin` (collector `:18-26`) | `resolve_trusted_executable` (`:56-112`) resolves `gh`/`git`/`jq`/`cmp`/`openssl`/`sha256sum` — and every `TASKDECK_*_EXECUTABLE` override — to an absolute program path and refuses any resolution inside the collector checkout; `assert_trusted_executables_outside` (`:88`, called at `:184-185`) repeats the rejection for the measured worktree **and the primary checkout sharing its Git directory**, because the configured PATH entry lives in the primary checkout and evidence is usually collected from a linked worktree. `resolve_state_path` now runs before the first GitHub query, so a forged tool never executes. | canaries 23, 24: a forged `gh` planted on PATH and through the override inside the checkout is rejected with no `gh` invocation logged; a forged tool inside the measured checkout is rejected before any GitHub query. Mutation control: with the rejection disabled the forged `gh` **is** invoked. |
| 2 | The persistent state is authenticated, then copied; the copied bytes are never authenticated (`:490-494`) | After `cp`, the copy is read once and put through `assert_authenticated_state_document` (`:655`) before any field is read; every later read — comparison, `.opening` extraction, packet emission — consumes that authenticated buffer instead of reopening a mutable file. `validate_session_token` now takes the document text and runs its three probes against one in-memory buffer instead of reopening the file three times. | canary 25: a jq shim rewrites the persistent state in the window between authentication and `cp`; `finish` fails with `opening evidence state was rewritten after start`, and the canary asserts the rewrite actually landed. Mutation control: with the copy's binding check relaxed, the same run emits `COMPLETE` carrying the forged field. |
| 3 | Replacement refs are rejected only at the boundaries; a transient ref can redirect the definition reads between them (`:446/:467/:468`) | `export GIT_NO_REPLACE_OBJECTS=1` at `:14`, documented, for every Git invocation the collector makes. | canary 27 (real Git checkout): a `refs/replace/` ref is installed through the mocked GitHub call that immediately precedes the definition reads and removed at the next one, so the boundary checks never see it. The canary first records that the ref really did redirect `head:.gitleaks.toml` to the base blob, then asserts the packet is `NOT VERIFIED`/`INCOMPLETE` and that the recorded `localHeadBlob` is the authentic blob. Mutation control: with the export deleted, the same fixture emits `CLEAN`/`COMPLETE`. |
| 4 | Definition binding reads mutable `HEAD:$path` (`:468`) | `rev-parse "$opening_head:$path"` (`:621`); `bind_enforcing_gitleaks_definitions` takes the authenticated opening head. | canary 28 (real Git checkout): `refs/heads/main` is moved to the opening base for the duration of the definition reads and restored before the closing assertion; the packet stays `NOT VERIFIED` and binds the authentic head blob. Mutation control: reading `HEAD:$path` emits `CLEAN`/`COMPLETE`. The mocked `git` now also refuses `rev-parse HEAD:<path>` outright, so the four existing no-op-definition canaries enforce the same property. |
| 5 | The documented wrappers embed `VALIDATED_SESSION_TOKEN` in the outer shell's command text (`SKILL.md:62`, `:120`) | The token is captured once through `tee` into a private file outside every checkout and redirected into `abort`/`finish` (`SKILL.md:45`, `:77`, `:136`); the literal no longer appears in any command text. The path is not a secret. The skill states plainly what the channel does **not** protect: a same-user process that can read the file, and `start` still printing the token to the tool transcript. | canary 30: runs the documented redirect form against the collector and asserts a `COMPLETE` packet, then asserts the skill contains no token-literal placeholder and no `printf … \| collector` form, and that the limits paragraph is present. |

## Canary suite conversion

`scripts/github/test-collect-pre-merge-evidence.sh:706-707` held two static presence checks
(`rg -Fq 'state_snapshot='`, `rg -Fq 'refs/replace/'`) that passed while findings 2 and 3 were
live. Both are removed; their properties are now proven at runtime by canaries 25, 27, 28 and 29.
Findings 3 and 4 are exercised against a **real Git checkout** the harness builds (bare origin,
base commit pushed, PR-head commit weakening `.gitleaks.toml`), with the attack installed and
removed through the mocked GitHub calls that bracket the definition reads — no timing race, the
sequence is deterministic. Four canaries first assert that a deliberately defective copy of the
collector fails them, so none of them can pass by accident.

## Verification

- `bash scripts/github/test-collect-pre-merge-evidence.sh` — **30/30 canaries passed**, run twice
  end to end (22 pre-existing, all still green; 8 new), ~5 min per run on Windows/Git Bash.
- `bash -n` on the collector and the canary harness: clean.
- `node scripts/check-docs-governance.mjs`, `node scripts/check-github-ops-governance.mjs`: passed.
- `git diff --check`: clean.
- `scripts/ci/check-dco-signoffs.sh 20f6ae40 d3ab936b`: passed, checked=20.
- Merged current `origin/main` (`20f6ae40`) with `--signoff`; the two merges were conflict-free and
  touched no file in this change's seam.

## NOT verified

- Hosted CI at this head (this PR is its first run). No CI job executes the canary suite — the
  local run above is the whole runtime evidence for this seam, exactly as `docs/TESTING_GUIDE.md`
  prescribes.
- No evidence session was run against a live GitHub PR; `gh` is mocked in every canary except the
  real-Git fixtures, which still mock `gh`.
- Canaries 23-25 and 30 exercise the mocked collector paths only; the real-Git fixtures cover
  findings 3 and 4.

## Residual risk

- **Configuration surface (tracked, #1625).** This change hardens the collector as a consumer. The
  writable gitignored PATH entry in `.codex/config.toml` still affects every other tool the harness
  runs; nothing here narrows that.
- **Same-user filesystem access.** The token file, the state file, and the collector's environment
  are all readable by a process running as the operator. The token-file channel removes command-text
  and history exposure only, and the skill says so rather than implying more.
- **Containment test.** Executable rejection uses an ancestor walk with the same-file test, so
  Windows path spellings, symlinks, and case differences cannot spell around it. A bind mount or a
  hard link of a checkout-local tool into a directory outside every rejected root would not be
  caught; that requires filesystem-level setup rather than dropping a file on PATH.
- **Definition binding still trusts the local object store.** `GIT_NO_REPLACE_OBJECTS=1` removes the
  replacement-ref redirect and the head OID removes the ref redirect, but an attacker who can write
  raw objects into `.git` is outside this threat model.
- Two MEDIUM residuals recorded on #1557 (COMPLETE-before-state-removal ordering; ignored
  configuration-input hermeticity) remain open and untouched. The third — the overwritten snapshot
  cleanup trap — is fixed here in passing, since the same lines were being rewritten.

Closes #1547

Supersedes #1557. Related: #1555, #1532, #1543, #1544, #1625.


Chris0Jeky:
## ⚠️ Remediation pushed but UNVERIFIED — do not merge yet

Head is now `10f19dd7` ("Close two reproduced bypasses of the evidence-tool trust check", +714/−114 across the collector, its test suite, and SKILL.md). This was pushed for **preservation** after the authoring session was cut short by a usage limit — it is **not** confirmed working.

### Why this PR exists

An adversarial review of the previous head (`d3ab936b`) **reproduced two working HIGH bypasses**, each producing a fully forged evidence packet (`collectorState=COMPLETE`, `secrets.verdict=CLEAN`) while all 30 canaries passed:

1. **Forged `awk`** — `awk` (the back half of the only hashing primitive, `sha256_text`) and other externals (`cp`/`mv`/`mktemp`/`tr`/`basename`) were never hardened by `resolve_trusted_executable`, so a shim in `.runtime-codex/bin` decides whether rewritten state authenticates.
2. **`git` runs before the trust check that should reject it** — `resolve_state_path` calls `git rev-parse` three times, then checks the roots that *untrusted git just returned*; a forged `git` in the primary checkout of a linked-worktree layout reports decoy roots and is never rejected.

### What `10f19dd7` claims to do (UNVERIFIED)

Bash-only root discovery + PATH sanitization before any external runs; extend trusted resolution to all externals and take `awk` out of the hash chain; new runtime canaries for both bypasses; corrected the false "no external tool runs before the rejection" comment.

### Required before this can be reviewed for merge

- Run the full canary suite `bash scripts/github/test-collect-pre-merge-evidence.sh` to green, twice (determinism).
- Run both new mutation controls (disable each defense → its canary must fail).
- A fresh adversarial review that re-attempts both bypasses against `10f19dd7` (this is security-critical evidence-gate code — the previous head passed 30/30 canaries while fully bypassable, so canary-green is necessary but not sufficient).
- CI note: the Required CI red seen earlier on this branch's `Backend Unit (windows-latest)` "Run CLI contract tests" is the pre-existing Windows CLI subprocess-timeout flake #1530 (environmental, `last=migration-begin`, passed on rerun at identical content on main) — not from this PR's bash/SKILL.md-only diff.

Supersedes parked #1557. `Closes #1547`. Broader `.codex/config.toml` PATH hazard is #1625.

Chris0Jeky:
Parked at exact head `10f19dd73f836fe0ec2de974254d8b1e230db08e`; keep this PR draft.

Current merge blockers are direct false-evidence paths, not polish: GitHub queries can be redirected by inherited `GH_REPO`; repository-local clean filters can hide modified bytes; ancestry uses mutable `HEAD`; bootstrap `readlink` can execute before its trust check; and the documented proving block can mask an earlier failed command. The POSIX review also reports the prescribed canary run reaches 32 passes and then fails at canary 33.

This session independently proved both scripts parse with `bash -n`, but could not execute the full canary/mutation suite in Windows Git Bash because no native `jq` resolved. That gap is recorded as NOT verified, not treated as success. No live evidence session was attempted because the current head is already ineligible.

Resume only with one bounded remediation of the current P1 set, a green full canary + mutation run in a native-jq environment, a real GitHub-backed evidence session pinned to the exact head/base, and fresh security review. Refs #1639 for the missing hosted canary lane.

Chris0Jeky:
Fresh-context review reconfirmed that this successor is not mergeable and should remain parked in draft. Confirmed HIGH paths remain: mutable-repository redirection through `GH_REPO`, clean/smudge filters hiding byte differences, ancestry checks against mutable `HEAD`, execution through an unvalidated symlink bootstrap, and a documented proving block that can mask an earlier command failure. The POSIX canary is still red after 32 passes, and 13 review threads remain unresolved. No merge or closure: #1547 remains the repair/successor lane.

Chris0Jeky:
FINAL PARK / CLOSE UNMERGED — exact draft head remains `10f19dd73f836fe0ec2de974254d8b1e230db08e`; branch history is preserved.

A history-preserving repair was attempted locally from current main `867e87cc74d0d8ca6796f7989a998b05e25ede4d`, but the bounded gate terminated before commit/push because no full green canary proof exists. Three repair runs stopped at fixture controls (8/38, then 25/38 twice), and fresh review found direct false-evidence paths in the same trust boundary:

- `text=auto` / `-text eol=crlf` raw-cleanliness semantics could certify Git-dirty bytes;
- `.git/info/grafts` could forge immutable-OID ancestry;
- remaining blocker: an unstaged POSIX `100644`↔`100755` mode change is invisible when bytes and index match.

The first two have uncommitted partial local repairs that were never fully canaried; the mode blocker remains unfixed. The original 13 review threads remain unresolved and no required full suite is green. Per the review/check attempt ceiling, this draft is being closed without merge. #1547 retains ownership of any future clean-room successor.

